PR #5, opened 2015-12-01
Fix README example

cesarandreu:


README.md
```diff
@@ -31,7 +31,7 @@ import { VirtualScroll } from 'react-virtualized';
 
 // List data as an array of strings
```
bvaughn: Since we're fixing things...this comment should not say "array of strings"

PR #10, opened 2015-12-04
Remove react-dom as a peerDependency

cesarandreu:
* Removed the `findDOMNode` call in VirtualScroll. HTML element refs already give you the DOM node in [v0.14.x](https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#dom-node-refs).

* Removed react-dom as a peerDependency, since it's not used by react-virtualized

* Added react and react-dom as devDependencies, otherwise they don't get installed with npm@3

This breaks support with react 0.13, but I'm not sure it worked before?

At the very least, npm would yell at you for installing an incompatible react version, so the allowed versions in peerDependencies would probably need to be updated.

Furthermore, there's no react 0.13 `react-dom` package nor does it support pure function components, which are used in FlexTable.

If you decide to keep support for react 0.13, I'd probably suggest using following check in VirtualScroll:

```js
if (/^(0\.13\.\d+)$/.test(React.version)) {
  this._scrollingContainer = this.refs.scrollingContainer.getDOMNode()
} else {
  this._scrollingContainer = this.refs.scrollingContainer
}
```

Since react 0.14 still has `getDOMNode`, you can't do a key check on the refs.

Also, the pure function components would need to be updated as `React.Component` or `React.createClass` classes in order to support react 0.13.

I don't know what usage % is like, so I don't know if it's worth supporting react 0.13 or not. Since it's slightly annoying to setup a harness that can work with both, I probably wouldn't bother with it. But then again, the required changes aren't super big.

In any case, removing `react-dom` and adding `react` / `react-dom` as devDependencies is a good idea :D.


bvaughn:
> This breaks support with react 0.13, but I'm not sure it worked before?

Hmm.. I had tested what was there before with both React 0.13 and 0.14 and it worked for me. (I had a local test harness running off of 0.13.)

The check did feel kind of hackish though. I don't mind "officially" stop trying to support 0.13.

PR #11, opened 2015-12-04
Simplify and tidy up scripts~

cesarandreu:
Hopefully I didn't break anything.

Needless to say, I haven't tried the `deploy` script... But I'm feeling lucky, I think it should work. Running `http-server ./build` and then opening `localhost:8080` shows the demo as one would expect.

Otherwise the other stuff should work. This depends on `react` and `react-dom` being devDependencies, otherwise it won't work.


bvaughn:
Ah, the changes to `deploy` and `start` look nice. Let me test it out a bit.

bvaughn:
The `build` index was broken without the postbuild script running but the fix was a super small change in the webpack config for that target so... everything else looks good. Haven't tried deploy yet either but I share your optimism. :)

PR #13, opened 2015-12-04
Kill classnames dependency~

cesarandreu:
I think this is a dependency that's pretty easy to kill


bvaughn:
Yeah, agreed. I'd been considering doing this.

cesarandreu:
I hadn't looked at it, but you're right, it's pretty tiny~ I'll close this PR.

PR #15, opened 2015-12-04
Set shouldPureComponentUpdate on the prototype

cesarandreu:
Dunno if it has any real effect on performance, but this lets you avoid having to reassign it per instance.


bvaughn:
Doubtful, but seems like a reasonable change.

PR #30, opened 2015-12-16
rowClassName now supports dynamic classnames via function argument

guykr-stratoscale:


bvaughn:
Thank you for the contribution. Looks good! I'll update the docs and push out a release shortly.

bvaughn:
Just pushed out version `2.3.0` with this change.

PR #43, opened 2016-01-04
Fix conditional in VirtualScroll

cesarandreu:
```js
var x = 1
x instanceof Number
// false
```

You could do a stricter type check if you really wanted to handle Number objects...

```js
typeof value == 'number' || (!!value && typeof value === 'object' && Object.prototype.toString.call(value) === '[object Number]')
```

I don't think you should bother, though, there's no reason to be passing in Number objects here.


bvaughn:
D'oh! Thanks for fixing this silly mistake. :)

PR #51, opened 2016-01-11
Add InfiniteLoader HOC

### LINKED ISSUE #38: Support loading more items when doing infinite scrolling
The scenario is basically: You have X number of items and when you reach the end of the list you either: (a) show a "Load more items" button / row / element when they reach the end of the list, and show a spinner when it's pressed, or (b) you automatically show a spinner when they reach the end of the list. Once new items show up, you continue scrolling.

I'm not sure of what the ideal API would be, but thinking about the states that need to be tracked, (disregarding the property names) I think the most flexible lower level API would need to provide something like:

* `hasMoreItems` {boolean} flag to decide if it should do anything when the end of the list is reached
* `autoLoadItems` {boolean} (ignored if `hasMoreItems` is false) flag to decide what to do when the end of the list is reached: (true) automatically trigger an action, or (false) render a "Load more items" button / row / element
* `onInfiniteLoad` {() => void} (ignored if `hasMoreItems` is false) Called when you click on "Load more items" with false `autoLoadMoreItems`, or if you reach the end of the list with true `autoLoadMoreItems`
* `isInfiniteLoading` {boolean} (ignored if `hasMoreItems` is false) flag to decide if it should show a spinner or the "Load more items" button / row / element 
* `getLoadMoreItemElement` {({ isInfiniteLoading: boolean, onInfiniteLoad: function }) => ?ReactElement} (ignored if `hasMoreItems` is false, return value ignored when `autoLoadItems` is true) get the element to render when loading (e.g. spinner) or when showing "Load more items" button / row / element
* `loadMoreItemsElementHeight` {number} (ignored if `hasMoreItems` is false) height for the return value of `getLoadMoreItemElement`

Thoughts?

It could probably be taken further by supporting bi-directional infinite scrolling (e.g. you get dropped in the middle of a list and you can follow the cursors up or down the list), which means you could throw away items after a certain count in order to limit memory usage. Although, I don't know how useful or practical that would be.

bvaughn:
Resolves #38 

Couple of open questions before this PR is merged and the feature is released:
* Is it important to be able to compose `AutoSizer` and `InfiniteLoader` components? (Is it even possible without reverting the `ChildComponent` change in [2.8.0](https://github.com/bvaughn/react-virtualized/blob/master/CHANGELOG.md#280)?)
* ~~Is it important for `InfiniteLoader` not to override the `onRowsRendered` of the inner `FlexTable` or `VirtualScroll`?~~


olslash:
gave it a quick look over, ran the demo, and i think this will certainly work for my use case.

Quick question, not 100% sure if this was addressed, but what would you suggest doing to prevent lots of redundant API calls? In the example, while rows are loading, scrolling past them is calling loadMoreRows on every render -- but i would think it's safe to assume that once you've called loadMoreRows for a range, you shouldn't need to call it for that range again.

Would that be the responsibility of the user? Does doing it this way open up behaviors that I'm not considering, or is it just going to be overhead for most projects?

olslash:
y'know i'm looking at it more closely and it may already be doing the efficient behavior -- if i'm wrong let me know :)

bvaughn:
It will track and only call load for ranges that have not been loaded
previously. (It should never try to load the same row twice.) How
aggressively you prefetch can be controlled via the threshold attribute.

Depending on your situation you could prefetch in chunks- (so if the loader
told you to load rows 50..60 there's no reason you couldn't load 50..100
instead).

source/InfiniteLoader/InfiniteLoader.js
```diff
@@ -0,0 +1,139 @@
...
+     */
+    loadMoreRows: PropTypes.func.isRequired,
+    /**
+     * Number of rows in list; can be arbitrary high number if actual number is unknown.
+     */
+    rowsCount: PropTypes.number,
+    /**
+     * Threshold at which to pre-fetch data.
+     * A threshold X means that data will start loading when a user scrolls within X rows.
+     * This value defaults to 10.
```
olslash: should be 15, right?

bvaughn: Yeah good call. I was experimenting with different default values. I'll
tighten up the docs before merging. :)

PR #57, opened 2016-01-13
Add onHeaderClick callback to FlexColumn

olslash:
This isn't a complete PR, but I wanted to start a discussion about the feature. 

For my project I need some external side effects to occur when clicking the header of a particular column, which are not related to sorting. This adds a prop `onHeaderClick` to FlexColumn that fires regardless of whether sorting is disabled, and in addition to any `sort` callback.

Questions:
* Does the API make sense? Would this be better placed on the FlexTable level (like sort) and get the column data as args?
* If the prop stays at the Column level, what args would be useful in the callback?

### DIFF
--- a/source/FlexTable/FlexColumn.js
+++ b/source/FlexTable/FlexColumn.js
@@ -45,8 +45,7 @@ export default class Column extends Component {
     cellDataGetter: defaultCellDataGetter,
     cellRenderer: defaultCellRenderer,
     flexGrow: 0,
-    flexShrink: 1,
-    onHeaderClick: () => null
+    flexShrink: 1
   }
 
   static propTypes = {
--- a/source/FlexTable/FlexColumn.js
+++ b/source/FlexTable/FlexColumn.js
@@ -68,11 +68,6 @@ export default class Column extends Component {
     dataKey: PropTypes.any.isRequired,
     /** If sort is enabled for the table at large, disable it for this column */
     disableSort: PropTypes.bool,
-    /**
-    * Optional callback when this column's header is clicked.
-    * (): void
-    */
-    onHeaderClick: PropTypes.func,
     /** Flex grow style; defaults to 0 */
     flexGrow: PropTypes.number,
     /** Flex shrink style; defaults to 1 */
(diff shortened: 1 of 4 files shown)


bvaughn:
Hey @olslash,

Thanks for sharing your use-case and this PR with me. I think this feature makes sense. :)

> * Does the API make sense? Would this be better placed on the FlexTable level (like sort) and get the column data as args?
> * If the prop stays at the Column level, what args would be useful in the callback?

IMO my initial thought is that this probably makes more sense being defined on `FlexTable` because (a) a similar `onRowClick` property is there and (b) it prevents you from having to explicitly attach the callback to each header (assuming there's a good chance you want to know about more than one header).

My thoughts on the signature of this function would be that it should probably only receive a single parameter- the `dataKey` (to identify which header was clicked). I don't think we should pass any sorting related information because (a) the header might not be sortable and (b) the `sort` callback could be used to listen for that type of information anyway.

What are your thoughts?

olslash:
Cool, i can agree on both things. Getting the dataKey in the callback would work for me and is consistent with the signature of `sort` as well. I'll make those edits.

bvaughn:
Thanks so much :)

olslash:
Should be ready to look at now

source/FlexTable/FlexColumn.js
```diff
@@ -45,7 +45,8 @@ export default class Column extends Component {
     cellDataGetter: defaultCellDataGetter,
     cellRenderer: defaultCellRenderer,
     flexGrow: 0,
-    flexShrink: 1
+    flexShrink: 1,
+    onHeaderClick: () => null
```
bvaughn: Looks like this guy got left over from the initial commit. :)

source/FlexTable/FlexTable.test.js
```diff
@@ -325,6 +327,36 @@ describe('FlexTable', () => {
     })
   })
 
+  describe('onHeaderClick', () => {
```
bvaughn: Great tests! :D

bvaughn:
@olslash ~ one minor suggestion and otherwise it looks great

olslash:
fixed, thanks

bvaughn:
Release 3.1.1 is available with this feature. Thanks again @olslash !

PR #61, opened 2016-01-15
pass columnData to onHeaderClick handler

olslash:
I realized it would also be convenient and probably expected to have access to columnData in this handler. For me it's needed because i have more than one column with the same dataKey (one renders a checkbox bound to the item and one renders the item itself)


bvaughn:
LGTM

bvaughn:
Release 4.0.2 has been published with this change. Thanks for the contribution.

Issue #59, opened 2016-01-13
Scrolling not inertial on mobile safari


aquach:
When using VirtualScroll in an iOS app, the scrolling is not inertial. I think this can be fixed by adding `-webkit-overflow-scrolling: touch` to the `VirtualScroll` div, but I thought I'd flag it here so you could investigate it on mobile yourself.

bvaughn:
Great suggestion. Thank you! I will try this. :)

aquach:
Sure thing! I'm now looking into issues relating to https://www.tjvantoll.com/2012/08/19/onscroll-event-issues-on-mobile-browsers/ : the onScroll handler will not fire until the very end of the inertial scroll, which means new elements will not be rendered until the scroll stops. I'm looking into how we can fire onScroll events during the middle of the inertial scroll so that this doesn't happen.

bvaughn:
Initial testing seems promising (at least in the simulator).

aquach:
Cool! Regarding my other issue: I've found that scrollTop is not updated until the end of the scroll when using UIWebView (but not Safari), which makes it impossible to update the visible cells during the inertial scroll :(

bvaughn:
@aquach Can you check out the [issues/53-and-59](https://github.com/bvaughn/react-virtualized/tree/issues/53-and-59) branch?

bvaughn:
I think I'm going to proceed with a point release of this feature only and leave #53 as a separate issue until I hear back from you. :)

aquach:
Awesome, thanks! I'll check out the other one.

PR #62, opened 2016-01-15
Smoother, more natural scrolling on touch devices

### LINKED ISSUE #53: Allow a 'margin' for rendering elements for VirtualScroll
I'm using VirtualScroll on mobile and it's awesome! However, because the renderer often can't keep up with my scrolling, I can see elements pop in as I scroll (as opposed to desktop, where the illusion is perfect). Popping isn't the worst thing, but it would be nice to provide a top and bottom "margin" where elements not visible but close to the visible viewport are still rendered to reduce the popping. At a high level, this would provide a balance parameter between the benefits of virtual scroll (hiding what's not visible for speed) and normal scroll (less popping but slower).

I do have a hacky version of this here (https://github.com/pennyapp/react-virtualized/commit/b019a145304b30f74401c9ccebb5bdfb49942e01) that I'm experimenting with for our project, but I doubt I'll take the time to reify it into something better unless there's strong demand.

bvaughn:
Resolves #53 and #59

WORK IN PROGRESS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 Changelog
 ------------
 
+#### 4.1.0
+Added `-webkit-overflow-scrolling: touch` for smoother inertial scrolling on mobile devices.
+
 #### 4.0.2
 Additional `columnData` parameter passed to `onHeaderClick` callback.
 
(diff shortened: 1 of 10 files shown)


bvaughn:
Closing in favor of PR #64

PR #65, opened 2016-01-18
feat(VirtualScroll): ScrollToIndexFromTop and ScrollToIndexFromBot

ran-y:
Change getUpdatedOffsetForIndex func to return max/min when ScrollToIndexFromTop/ScrollToIndexFromBot props are configured.

Sometimes you want to control the exact line that will be presented when scrolling to a certain line number. This small change allows you to control whether you want it from top or from bottom.
* These props are optional and doesn't change anything if you decide not to use them.

### DIFF
--- a/docs/VirtualScroll.md
+++ b/docs/VirtualScroll.md
@@ -14,6 +14,8 @@ Describes the header and cell contents of a table column
 | rowRenderer | Function | ✓ | Responsbile for rendering a row given an index. Rendered rows must have a unique `key` attribute. |
 | rowsCount | Number | ✓ | Number of rows in list. |
 | scrollToIndex | Number |  | Row index to ensure visible (by forcefully scrolling if necessary) |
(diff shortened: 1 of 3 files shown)


bvaughn:
Hey @ran-y,

Thanks for submitting the PR for discussion. I have misgivings about the new properties though. They seem overly...specific? I prefer the simplicity of the current scroll-to operation.

If you really need to control where a row appears within the viewport, you could always scroll past it and then back to it by invoking the `scrollToRow` method twice.

```js
// Ensure a row is the top row:
this.refs.VirtualScroll.scrollIntoView(rowsCount - 1)
this.refs.VirtualScroll.scrollIntoView(someSpecificRow)

// Ensure a row is the bottom row:
this.refs.VirtualScroll.scrollIntoView(0)
this.refs.VirtualScroll.scrollIntoView(someSpecificRow)
```

ran-y:
Hi ! thanx for the quick response,
i agree that these new properties are overly specific,
though i think that sometimes you do want to control the exact row within the viewport,
which you gave a working solution to that i'm wondering if it's best for then performances.

thanx for this awesome project, very very cool !

PR #90, opened 2016-02-11
Default <FlexTable>'s :headerHeight prop to 0

crzrcn:
So that :headerHeight does not have to be specified when :disableHeader
is true.

### DIFF
--- a/source/FlexTable/FlexTable.js
+++ b/source/FlexTable/FlexTable.js
@@ -44,7 +44,7 @@ export default class FlexTable extends Component {
     /** Optional CSS class to apply to all column headers */
     headerClassName: PropTypes.string,
     /** Fixed height of header row */
-    headerHeight: PropTypes.number.isRequired,
+    headerHeight: PropTypes.number,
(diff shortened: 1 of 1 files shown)


source/FlexTable/FlexTable.js
```diff
@@ -44,7 +44,7 @@ export default class FlexTable extends Component {
     /** Optional CSS class to apply to all column headers */
     headerClassName: PropTypes.string,
     /** Fixed height of header row */
-    headerHeight: PropTypes.number.isRequired,
+    headerHeight: PropTypes.number,
```
bvaughn: This should still be `.isRequired`. Your static property will handle the case when it's not passed in externally.

bvaughn:
Thanks for submitting the PR dude. Added a minor comment.

crzrcn:
Done, back to you.

source/FlexTable/FlexTable.js
liubko: I don't understand why the prop has to be marked as required if there is a default value.. can you elaborate?

PR #110, opened 2016-02-17
Fixes slow (mouse) wheel scrolling

### LINKED ISSUE #2: Extremely slow scrolling with scroll wheel/arrow keys in Firefox
When scrolling with the scroll wheel or the arrow keys in Firefox, it does very small increments, meaning I have to do at least four "scrolls" to scroll a single item (row height == 30).

I'm using the demo at https://bvaughn.github.io/react-virtualized/ , both VirtualScroll and FlexTable are affected, and I'm using Windows 7. Both Firefox 42 and Firefox Developer Edition 44 are affected.

bvaughn:
Resolves #2.

The gist of this fix for the mouse-wheel issue reported against react-virtualized on Firefox is that `componentDidUpdate` no longer adjusts `scrollLeft` / `scrollTop` in reaction to observed scroll events. This was interrupting the scroll animation.


es/Grid/Grid.js
```diff
@@ -282,13 +282,19 @@ var Grid = (function (_Component) {
...
+      var isScrolling = _state.isScrolling;
       var scrollLeft = _state.scrollLeft;
       var scrollTop = _state.scrollTop;
 
       // Make sure any changes to :scrollLeft or :scrollTop get applied
-      if (scrollLeft >= 0 && scrollLeft !== prevState.scrollLeft || scrollTop >= 0 && scrollTop !== prevState.scrollTop) {
-        this.refs.scrollingContainer.scrollLeft = scrollLeft;
-        this.refs.scrollingContainer.scrollTop = scrollTop;
+      // Don't re-apply while a scroll is in progress; this causes slow scrolling for certain OS/browser combinations (eg. Windows and Firefox)
+      if (!isScrolling) {
```
bvaughn: I'm not happy with this approach as-is. I think it would work for almost all cases but since it depends on timing... it has potential to break.

I guess this is more of a proof-of-concept. If it resolves the issue then I'll think of a more elegant way to differentiate between the types of scrolling:
* User scrolls (with or without mouse-wheel) via the browser
* Incoming scroll-to-row/column property
* Incoming scroll left/top via `setScrollPosition` function (or property, as with `ScrollSync`)
* Scroll adjustments based on reduced row/column count (to prevent over-scrolling)

mstange: This sounds much better to me. But actually you just need to differentiate between two sources of scroll position changes: observation of browser scrolling, and intentional adjustments. So instead of isScrolling, I'd call that part of the state scrollPosUpdateReason, with the two possible values "observed" and "requested".

And in the comment, instead of "this causes slow scrolling for certain OS/browser combinations (eg. Windows and Firefox)" I'd write:
"Assigning to scrollLeft / scrollTop tells the browser to interrupt any running scroll animations, and to discard any pending async changes to the scroll position that may have happened in the meantime (e.g. on a separate scrolling thread). So we only set the element's scrollLeft/scrollTop when we really require an adjustment of the scroll position."

I found out that the old code altered scrolling behavior even on OS X with touchpad scrolling (i.e. without any scroll animations), just less noticeably so. What happened was that the scroll event fired, so the current scroll position was recorded in the state. But Grid's _setNextState processes state changes asynchronously, so before componentDidUpdate was called with that state, the scroll position could have changed again, and then we'd overwrite the new scroll position with the one that we recorded in the state.

bvaughn: Thanks for the suggestion @mstange! What you propose sounds solid. :)

PR #117, opened 2016-02-22
use cross-env; win compatibility

mikekidder:


bvaughn:
Thanks @mikekidder  :)

PR #119, opened 2016-02-24
Updated ES6 modules (jsnext:main target) to use es2015-rollup Babel preset

### LINKED ISSUE #118: Library packaging / distribution
Hi,

Thanks for your work on this. Using the library, I notice a few things which can be improved the way it is packaged / distributed to users. From you config in package.json (`main` pointing to the standalone one, and `jsnext:main` pointing to the commonjs files), one distribution seems to be missing:
- `main` should point to commonjs files (currently in `es` folder)
- `jsnext:main` should point to files compiled with preset `es2015-rollup`: basically like commonjs except import / export statements are not transpiled.

If you agree with the points listed above, I can help and open a PR.

bvaughn:
Resolves #118


troch:
That looks about right. The next thing you need to do is to remove `.babelrc` from npm, because it could interfere with rollup.

I would suggest an `.npmignore` file looking like this, to keep it light :smile::

```
docs
source
.*
*.md
*.json
circle.yml
webpack.config.*.js
```

bvaughn:
I don't see how .babelrc could interfere with Rollup. I also don't plane to
git-ignore my webpack configs :)

troch:
I wasn't suggesting to gitignore, but to npmignore. When you use rollup and want to load sources specified by `jsnext:main`, rollup will make sense of the dependencies, tree-shake and then transpile with babel (what is left to transpile, i.e. import / export statements).

When using babel, babel (unless specified otherwise) will use the nearest `.babelrc` file. Let say you use `react-virtualized` in your project, you have your own `.babelrc` file in the root of your repo, and you want to build with rollup: you do not want your `.babelrc` file to be shadowed by the `.babelrc` file of one of your dependencies. This is also why you don't point `jsnext:main` to your `source` directory, because one doesn't want to have to know about all the presets and transform plugins required to transpile dependencies of a project.

bvaughn:
Sorry. I misread/misunderstood your message. It was 5am and my brain wasn't
fully on yet. What you're describing (npmignore) sounds reasonable.

Issue #94, opened 2016-02-12
Investigate replacing `transform` with `transform3d` for supported browsers 


bvaughn:
Things to investigate:
* Does it cause too much GPU strain?
* Does it introduce any rendering artifacts for Safari (or other browsers)
* Should it be opt-in or on by default?

Reference: [detecting `transform3d` support](https://gist.github.com/lorenzopolidori/3794226).

# Original Description

Refer to [this comment](https://github.com/bvaughn/react-virtualized/issues/87#issuecomment-183481441) by James:
> I think I found one potential problem: each row is placed by setting the top CSS property, which apparently makes Chrome redraw the whole viewport. I changed it to transform: translateY(...) and it seems a bit smoother.

Quick test _does_ seem smoother but it could be confirmation bias. Need to test more and confirm that (1) performance really does improve and (2) not at an unacceptable GPU cost.

Will leave this issue open for now and see if more people have opinion or information.

jlongster:
Does Chrome have any way to measure GPU usage? The only potential downside is forcing each row to be handled as a GPU texture (or something like that). I'm not sure what the right pattern here is, but using transforms is probably the way to go, especially since we're only showing a constant number at one time anyway.

I've heard concerns from various browser devs about using too much GPU memory by overusing CSS properties that trigger GPU offloading, but I'm not really sure. What's the "scientific" way we can measure this?

I can see if there's anything in Firefox I can profile. I can't tell any difference in Firefox 46 with the APZ feature making things happen async.

bvaughn:
> Does Chrome have any way to measure GPU usage?

I _think_ the white bar inside of Chrome's Timeline view is GPU but I'm _very_ inexperienced with that type of profiling. :)

jlongster:
I found it: it's in the FPS overlay (???). devtools menu > more tools > rendering settings > show FPS meter. That meter shows the amount of GPU memory.

Observed results while scrolling down through the entire page:

**Using `top`**:

Without any overscan rows: 8-16 MB 
With an overscan of 100: 8->25 MB

**Using `transform`**:

Without any overscan rows: 15->22MB
With an overscan of 100: 30->55MB

The question is if this is acceptable or not. I'm not really sure. But it seems like it could potentially use about twice as much memory on average (although this is all anecdotal, just watching the meter while scrolling).

[...]

PR #124, opened 2016-02-25
Grid: Use key which uniquely identifies children

Arxisos:
When I upgraded to the newest version I suddenly had problems with laggy rendering on Firefox. It turned out that the key to identify the children changed on every re-render, resulting in many unnecessary DOM operations. By using a key which is unique between multiple re-renders I could resolve the problem. With this change my [complex data table](https://github.com/Arxisos/react-rich-table) is scrolling smoothly again.

http://jaero.space/blog/react-performance-1 helped me to understand why the `key` property is so important.

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -425,8 +425,6 @@ export default class Grid extends Component {
       rowStartIndex = overscanRowIndices.overscanStartIndex
       rowStopIndex = overscanRowIndices.overscanStopIndex
 
-      let key = 0
-
       for (let rowIndex = rowStartIndex; rowIndex <= rowStopIndex; rowIndex++) {
         let rowDatum = this._rowMetadata[rowIndex]
 
@@ -437,7 +435,7 @@ export default class Grid extends Component {
 
           child = (
             <div
-              key={++key}
+              key={rowIndex + '_' + columnIndex}
               className='Grid__cell'
               style={{
                 transform,
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -20,15 +20,6 @@ import shouldPureComponentUpdate from 'react-pure-render/function'
 const IS_SCROLLING_TIMEOUT = 150
 
 /**
- * Controls whether the Grid updates the DOM element's scrollLeft/scrollTop based on the current state or just observes it.
- * This prevents Grid from interrupting mouse-wheel animations (see issue #2).
- */
-const SCROLL_POSITION_CHANGE_REASONS = {
-  OBSERVED: 'observed',
-  REQUESTED: 'requested'
-}
-
-/**
  * Renders tabular data with virtualization along the vertical and horizontal axes.
  * Row heights and column widths must be known ahead of time and specified as properties.
  */
@@ -186,7 +177,7 @@ export default class Grid extends Component {
    */
   setScrollPosition ({ scrollLeft, scrollTop }) {
     const newState = {
-      scrollPositionChangeReason: SCROLL_POSITION_CHANGE_REASONS.REQUESTED
+      scrollEventIsCancelable: false
     }
 
     if (scrollLeft >= 0) {
@@ -227,14 +218,14 @@ export default class Grid extends Component {
 
   componentDidUpdate (prevProps, prevState) {
     const { columnsCount, columnWidth, height, rowHeight, rowsCount, scrollToColumn, scrollToRow, width } = this.props
-    const { scrollLeft, scrollPositionChangeReason, scrollTop } = this.state
+    const { scrollLeft, scrollEventIsCancelable, scrollTop } = this.state
 
     // Make sure requested changes to :scrollLeft or :scrollTop get applied.
-    // Assigning to scrollLeft/scrollTop tells the browser to interrupt any running scroll animations,
+    // Assigning to scrollLeft/scrollTop tells the browser to interrupt any running scroll animations if the scroll event is cancelable,
     // And to discard any pending async changes to the scroll position that may have happened in the meantime (e.g. on a separate scrolling thread).
     // So we only set these when we require an adjustment of the scroll position.
     // See issue #2 for more information.
-    if (scrollPositionChangeReason === SCROLL_POSITION_CHANGE_REASONS.REQUESTED) {
+    if (!scrollEventIsCancelable) {
       if (
         scrollLeft >= 0 &&
         scrollLeft !== prevState.scrollLeft &&
@@ -775,7 +766,7 @@ export default class Grid extends Component {
       this._setNextState({
         isScrolling: true,
         scrollLeft,
-        scrollPositionChangeReason: SCROLL_POSITION_CHANGE_REASONS.OBSERVED,
+        scrollEventIsCancelable: event.cancelable,
         scrollTop
       })
     }


bvaughn:
The key change was actually critical to resolving issue #2 and there was a lot of back and forth discussion here and on Twitter (see [here](https://twitter.com/dan_abramov/status/701465968111910913) for example) about it.

bvaughn:
Your change was, essentially, what I had before: https://github.com/bvaughn/react-virtualized/commit/f31a18cbc352ad64e87310eb988e7b9cc287b0ac#diff-43c6f2c20a98e4816031a2f99092ecb4

Arxisos:
Yes, I also didn't have problems with laggy scrolling when I used an older version.

bvaughn:
Right, but unfortunately... I don't know how to solve both the wheel-scrolling problem with Firefox and the scroll-animation interruption problem with Safari if I specify a more "meaningful" key.

For what it's worth, I have not seen "laggy rendering on Firefox". And I don't think I've heard about it from others either. Hm.

Arxisos:
It's not laggy as long the rows don't include complex components. 

I actually experienced the wheel-scrolling issue with an older version, too. The new version correctly resolved that issue. However, wheel-scrolling still works for me on Firefox 44.0.2 after I applied the changes in this pull request. But I'm not sure if I didn't update Firefox in the meantime...

Arxisos:
You just need to add a `console.log('row rendered')` to a row component to see the difference. Without an unique key the `render` method of all visible (pure) row components is called on every scroll. With an unique key the number of `render` calls decreases significantly.

bvaughn:
Yeah, I understand how the keys work. I guess my test harnesses are usually pretty simple in terms of cell-content though.

To be clear, your change wouldn't cause a regression in Firefox- but, it would break the scrolling animation for Safari. (At least, assuming you see what I saw when testing.) If you open a list in Safari and do a really fast scroll- the animation will start and then slam to a halt. Slow scrolling worked okay but fast scrolling was totally broken.

I was able to avoid that behavior by...
* Setting `pointer-events: none` (which obviously isn't an acceptable solution)
* Re-add the `scrollContainer.scrollTop = this.state.scrollTop` line (which breaks wheel scrolling in Firefox)
* Change the keys (like I ended up doing)

Do you have any other ideas by chance?

Arxisos:
I'll look into it and share my findings ASAP.

bvaughn:
Much appreciated. :) Maybe I was just looking at it for too long and missed something obvious.

bvaughn:
So I put together my own test harness for this- and I think the best performance combination seems to come from leaving the `key` as-is but switching the positioning of cells away from `transform: translate` and back to `top` and `left`.

bvaughn:
Check out this updated build:
https://s3.amazonaws.com/storage.briandavidvaughn.com/top-left-transform/index.html

bvaughn:
Closing this PR in favor of #125 which- to the best of my knowledge- addresses the performance issues reported here and on #94 without causing a regression of the scroll-wheel fix from #2

bvaughn:
@Arxisos,

Thank you so much for reporting this issue. I _think_ it should be fixed in version 5.2.3 (just released) but please let me know if that is not the case and I'll continue to work with you on it.

Arxisos:
I compared 5.2.3 and this PR and the PR performs without lags for me, but unfortunately 5.2.3 still has noticeable lags. I'll send you an email with a recording (can't make it public here).

bvaughn:
Does 5.2.3 lag less for you? In my testing (and the test harness I linked to above) it lags _much less_ than before. Maybe your "complex renderer" is significantly different than mine.

Arxisos:
I cannot really (subjectively) see a large improvement between 5.2.2 and 5.2.3 on my real world application. So I would prefer to use my fork for my use case as the lag is really disrupting for the end-user. 

Do you think we could in some way use the new `<ScrollSync />` HOC to have a "fake" container which updates the "real" container, while bypassing all the side-effects we fight with?

bvaughn:
Interesting. I guess it depends on the type of component then. In my test
harnesses I did see an improvement.

Again, I hate dealing with edge-cases like this because if you fix one
thing, you break another. I really don't want to get into the business of
re-implementing scroll animations because each browser handles those
differently and it seems unlikely that I'll handle them as well as any
given browser.

I'm not sure what to do at the moment unfortunately.

bvaughn:
To be clear (in case it sounded like I was just throwing up my hands and
giving up)- I'll try to fix this. I just don't have any ideas at the moment
and so I can't promise a timeline.

I'd welcome any input or information you can provide if you make any
breakthroughs on your end.

Arxisos:
When debugging with `react-addons-perf` I can see that React updates the DOM completely differently.

Unique key (this PR): `set innerHtml`
Current key: `update styles` and `set textContent`

Perhaps Safari does not like the inner html to be changed while smooth scrolling?

bvaughn:
That's an interesting idea.

Arxisos:
I've managed to find a way to get rid of both regressions (Safari and Firefox). Basically I explored differences in the scroll event in Safari and Firefox and I found that `cancelable` is `true` for FF, but `false` for Safari. 

I exploited this behavior in commit https://github.com/Arxisos/react-virtualized/commit/e30c6e61ff743720e874d5b62cc6da48c38b51ef to execute `scrollContainer.scrollTop = this.state.scrollTop` for Safari, but not for Firefox when mouse wheel scrolling.

bvaughn:
Interesting. And this combination for you results in better scrolling performance?

Arxisos:
Yes, I just double-checked it. My real life application is scrolling perfectly smoothly with these changes.

bvaughn:
Unfortunately, patching in that change causes the quick-swipe scrolling in Safari to be interrupted in my test harness.

Arxisos:
Alright, now as you say it I notice it, too...

Arxisos:
Sorry, I misunderstood you. I thought with quick-swipe scrolling you meant a very, very fast swipe (not the regression itself). I still manage to interrupt the scrolling if I try very hard, but it does not happen often.

It seems like that the threshold until the scroll is interrupted is a lot higher after this change (so high that I nearly don't manage to hit the threshold). I am testing using the `<VirtualScroll />` example.

That's why I thought the regression was resolved. 

Can you reproduce what I am talking about (clone my fork, `npm run start` and test on the `<VirtualScroll />` example)? Do you see the difference between the two commits I made?

bvaughn:
> I still manage to interrupt the scrolling if I try very hard, but it does not happen often.

For me it's very easy to reproduce. Especially in my test harness.

Checking out your test harness now.

bvaughn:
Yes, I can very easily reproduce it in your fork.

bvaughn:
Check out the `playground` folder I added in 277e066. If you pull it into your fork you should be able to reproduce the problem in Safari easier.

bvaughn:
There might still be something to the `event.cancellable === false` angle though. I hadn't considered that before you mentioned it.

Either way, we need a solution that addresses all of the following:
* Don't break mouse-wheel scrolling for Firefox
* Don't break quick scrolling for Safari
* Don't clobber performance ("stuttery" scrolling, "janky" rendering, etc)

Arxisos:
OK, I was able to reproduce what you are saying on my fork with the `playground` you added. It is interesting that I hit the problem with the `playground`, but not with the `<VirtualScroll />` example.

bvaughn:
That is interesting. I hit it in both, but it's definitely harder to avoid in the playground example. (I have to scroll _really_ slow for it to work at all in Safari where as in the normal demo site, scrolling medium speed works.)

Arxisos:
I am wondering why `scrollContainer.scrollTop = this.state.scrollTop` does not work anymore, because you said:

> I was able to avoid that behavior by...
> 
> - Setting pointer-events: none (which obviously isn't an acceptable solution)
> - Re-add the scrollContainer.scrollTop = this.state.scrollTop line (which breaks wheel scrolling in Firefox)
> - Change the keys (like I ended up doing)

`scrollContainer.scrollTop = this.state.scrollTop` is what I am doing in my fork by exploiting `event.cancellable === false` (I think so at least).

bvaughn:
Yeah, not sure. I had to stop working on this temporarily (day job) but I'll come back to it soon.

bvaughn:
Just a quick update. I've been trying different combinations of things, and I think that I'm currently happiest with this change (essentially what you recommended):

```diff
diff --git a/source/Grid/Grid.js b/source/Grid/Grid.js
index 9b249dc..7017a1d 100644
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -425,18 +425,17 @@ export default class Grid extends Component {
       rowStartIndex = overscanRowIndices.overscanStartIndex
       rowStopIndex = overscanRowIndices.overscanStopIndex
 
-      let key = 0
-
       for (let rowIndex = rowStartIndex; rowIndex <= rowStopIndex; rowIndex++) {
         let rowDatum = this._rowMetadata[rowIndex]
 
         for (let columnIndex = columnStartIndex; columnIndex <= columnStopIndex; columnIndex++) {
           let columnDatum = this._columnMetadata[columnIndex]
           let child = renderCell({ columnIndex, rowIndex })
+          let key = `${rowIndex}-${columnIndex}`
 
           child = (
             <div
-              key={++key}
+              key={key}
               className='Grid__cell'
               style={{
                 height: this._getRowHeight(rowIndex),
@@ -773,10 +772,14 @@ export default class Grid extends Component {
       this.state.scrollLeft !== scrollLeft ||
       this.state.scrollTop !== scrollTop
     ) {
+      const scrollPositionChangeReason = event.cancelable
+        ? SCROLL_POSITION_CHANGE_REASONS.OBSERVED
+        : SCROLL_POSITION_CHANGE_REASONS.REQUESTED
+
       this._setNextState({
         isScrolling: true,
         scrollLeft,
-        scrollPositionChangeReason: SCROLL_POSITION_CHANGE_REASONS.OBSERVED,
+        scrollPositionChangeReason,
         scrollTop
       })
     }
```

It breaks the quick-scroll use case in Safari but... that seems better than the performance loss.

bvaughn:
By the way, I'm going to move things over to PR #126

bvaughn:
@Arxisos~ Check out 5.2.4 release and see if it doesn't help your performance situation

Arxisos:
I can confirm that it fixed my performance problem and that it does not cause a regression.

bvaughn:
Fantastic :) Thank you.

PR #129, opened 2016-02-27
Split CommonJS and UMD builds; point package main at CommonJS build

bvaughn:


Guria:
Why are you commiting dist versions?

Guria:
It is better add all dist to `.gitignore` and not ignore them in `.npmignore`

bvaughn:
In part, it's a habit left over from publishing repos to Bower. It also gives me a quick-and-easy way to compare what's changed in the built versions of my library. Either way, it shouldn't really impact end users.

PR #133, opened 2016-02-29
Only hide overflow vertically or horizontally

ryanseddon:
I split the if statement into horizontal and vertical checks as I have a grid that uses infinite scrolling so even though my horizontal width was `<=` the available space my vertical height was always false and the overflow never gets hidden.

### Todo
- [x] tests

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -463,8 +463,12 @@ export default class Grid extends Component {
     // Force browser to hide scrollbars when we know they aren't necessary.
     // Otherwise once scrollbars appear they may not disappear again.
     // For more info see issue #116
-    if (totalColumnsWidth <= width && totalRowsHeight <= height) {
-      gridStyle.overflow = 'hidden'
+    if (totalColumnsWidth <= width) {
+      gridStyle.overflowX = 'hidden'
+    }
+
+    if (totalRowsHeight <= height) {
+      gridStyle.overflowY = 'hidden'
     }
 
     return (
--- a/source/Grid/Grid.test.js
+++ b/source/Grid/Grid.test.js
@@ -100,6 +100,18 @@ describe('Grid', () => {
     })
   })
 
+  describe('hide scrollbars based on number of children rendered', () => {
+    it('should set overflow-x on Grid container if columns <= available space', () => {
+      const node = findDOMNode(renderGrid({ columnsCount: 4 }))
+      expect(node.style.overflowX).toEqual('hidden')
+    })
+
+    it('should set overflow-y on Grid container if rows <= available space', () => {
+      const node = findDOMNode(renderGrid({ rowsCount: 5 }))
+      expect(node.style.overflowY).toEqual('hidden')
+    })
+  })
+
   /** Tests scrolling via initial props */
   describe(':scrollToColumn and :scrollToRow', () => {
     it('should scroll to the left', () => {


bvaughn:
Interesting. I was contemplating doing this at lunch earlier (after I made the release). :) Let me smoke test a bit. I think it should be a good change.

bvaughn:
Ah, this change has a (probably?) negative side effect of making the area behind the scrollbar inaccessible in the event that there's only 1 scrollbar. For instance, before this change...
screen shot 2016-02-28 at 5 16 14 pm

After this change:
screen shot 2016-02-28 at 5 16 27 pm

Basically with this change, the bottom ~17 pixels of each cell cannot be viewed because it is hidden behind the bottom scrollbar. The problem is more clearly seen if you have multi-line text, for example:
screen shot 2016-02-28 at 5 19 45 pm

ryanseddon:
That's a pretty rare case though right?

bvaughn:
Well, you can always set this via CSS for grids that you know will always be single row. :) That's why I allow custom `className`.

bvaughn:
To be clear, I'm not completely convinced of what is the right/wrong thing to do here. It's helpful to discuss and debate it.

ryanseddon:
So one idea is to measure the scrollbar height and set `padding-bottom` on the cells?

bvaughn:
I'm not sure that would work though, considering the `Grid`'s `height` and `rowHeight` are specified as fixed values. (It's not a grow-to-fill sort of view.) So if I added the padding... I'd have to subtract from the specified `rowHeight` which would potentially clip a row. Right?

ryanseddon:
yes but then as a user of this library I would see that and perhaps limit my data being displayed or bump the cell height to fit it in

bvaughn:
Hmm... I'm not sure how practical that is. I mean, it's practical (in a sense) for single-row grids. But then at that point you can specify your own CSS to achieve the same thing:

```css
.SingleRowGrid {
  overflow-y: hidden;
}
```
```js
<Grid
  className='SingleRowGrid'
  {...otherProps}
/>
```

But what about for multi-row grids? I'm not sure it would be okay for all uses if the last row was shorter than the others.

ryanseddon:
Well in theory you'd apply that padding to the last row only.

bvaughn:
Ah, I see what you're saying. So the calculation would end up being something a little more complicated like...

```js
if (totalRowsHeight + scrollHeight <= height) {
```

ryanseddon:
yes exactly. i'll give something a try in this pr.

ryanseddon:
So I think my other approach that has similar code to your flextable scrollbarWidth checking is better than setting overflow property. Let me put that up as a PR.

ryanseddon:
So I've made a test case! https://jsbin.com/xibefa/edit?output

This is my scenario, you can see that your current fix to check both horizontal and vertical and then only set overflow to hidden doesn't solve the test cases issue where I don't really care about handling vertical scrollbars but I do horizontal as the testcase should easily fit the 3 column layout.

bvaughn:
Test harnesses are helpful.

It's late but I'll take a look in the morning at some point.

bvaughn:
Man. I am torn on this. I am starting to lean towards the direction you've taken in this PR because I think it gets incredibly complicated to handle it in the _ideal_ way. I think it would involve passing the scrollbar size through to the column sizer function- which would involve forcing `overflow: scroll` to calculate the scrollbar width which... I think isn't worth it.

bvaughn:
Let's beef the tests up slight?

```js
  describe('hide scrollbars based on number of children rendered', () => {
    it('should set overflowX:hidden on scroll-container if columns fit within the available width', () => {
      const node = findDOMNode(renderGrid({ columnsCount: 4 }))
      expect(node.style.overflowX).toEqual('hidden')
    })

    it('should leave overflowX:auto on scroll-container if columns require more than the available width', () => {
      const node = findDOMNode(renderGrid({ columnsCount: 25 }))
      expect(node.style.overflowX).not.toEqual('hidden')
    })

    it('should set overflowY:hidden on scroll-container if rows fit within the available height', () => {
      const node = findDOMNode(renderGrid({ rowsCount: 5 }))
      expect(node.style.overflowY).toEqual('hidden')
    })

    it('should leave overflowY:auto on scroll-container if rows require more than the available height', () => {
      const node = findDOMNode(renderGrid({ rowsCount: 25 }))
      expect(node.style.overflowY).not.toEqual('hidden')
    })
  })
```

bvaughn:
Eh, I'll just merge now and make the change myself :)

bvaughn:
Just released 5.3.2 with this change. Let me know if you see any problems @ryanseddon

ryanseddon:
Thanks @bvaughn will update our project shortly.

PR #137, opened 2016-03-01
Check horizontal scrollbar width and adjust inner scroll container

ryanseddon:
This PR checks the width of the scrollbar and shrinks the internal scroll containers width to remove scrollbarWidth. We avoid setting overflow and potentially cutting off content, a user can set overflow via custom className if they need to hide scrollbars in a fixed grid header scenario.

Alternate to #133

No scrollbars when all columns fit:
screen shot 2016-03-01 at 11 19 39

Scrollbars when columns don't fit:
screen shot 2016-03-01 at 11 19 54

Double line content:
screen shot 2016-03-01 at 11 27 40

screen shot 2016-03-01 at 11 27 56

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -11,6 +11,7 @@ import {
 import cn from 'classnames'
 import raf from 'raf'
 import React, { Component, PropTypes } from 'react'
+import { findDOMNode } from 'react-dom'
 import shouldPureComponentUpdate from 'react-pure-render/function'
 
 /**
@@ -141,7 +142,8 @@ export default class Grid extends Component {
       computeGridMetadataOnNextUpdate: false,
       isScrolling: false,
       scrollLeft: 0,
-      scrollTop: 0
+      scrollTop: 0,
+      scrollbarWidth: 0
     }
 
     // Invokes onSectionRendered callback only when start/stop row or column indices change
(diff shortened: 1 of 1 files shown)


source/Grid/Grid.js
```diff
@@ -522,10 +528,11 @@ export default class Grid extends Component {
 
   _getColumnWidth (index) {
     const { columnWidth } = this.props
+    const { scrollbarWidth } = this.state
 
     return columnWidth instanceof Function
-      ? columnWidth(index)
-      : columnWidth
+      ? columnWidth(index) - scrollbarWidth
```
bvaughn: Why are you subtracting `scrollbarWidth` from each column?

bvaughn:
Hey @ryanseddon,

Thanks for taking the time to submit a PR! :)

I'm a little confused by part of it and so I left a question. This also only seems to handle the horizontal axis (scrollbar width). Seems like any solution we should needs to handle vertical too.

I'm still a little unclear from your 2nd and 3rd screenshots how this differs from the approach currently in master. (Maybe you could show me a case when they'd render something differently?)

ryanseddon:
Hmm I think I may have under thought the issue and this PR doesn't really solve what's going on it just looked like it did in my screenshots.

bvaughn:
No worries dude. It's easy to do. I think this is an issue with a lot of edge-cases.

I have an idea based on our discussion on the other ticket though that I might give a shot. At least it's better than it was at this point :)

Issue #131, opened 2016-02-28
Always-on scrollbar causes offset issues with ScrollSync


bvaughn:
Split off from #116, reported by @ryanseddon; similar to #81 but for multiple `Grid`s.

![scrollsync-offset](https://cloud.githubusercontent.com/assets/143402/13206249/44bba7c6-d94e-11e5-9829-b2eb52a38712.gif)

bvaughn:
Closed via PR #161

PR #155, opened 2016-03-15
Added column and row locking

jquense:
Heyo!

I added the ability to freeze rows and columns a la Excel. I realize this may be a bit high level for a library such as this but I thought it was a good idea for a few reasons:

1. relatively low complexity, just specifying a different offset
2. low to no perf impact on folks that don't use it and (in my tests) really minimal impact on those that do.
3. difficult to implement from the outside, ScrollSync sort of gets you there, but issues like #131 make it a more complicated. More relevant though when freezing columns you really want the scrollbar to span the entire horizonal width, not just the non-frozen section.

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -10,6 +10,7 @@ import {
 } from '../utils'
 import cn from 'classnames'
 import raf from 'raf'
+import getScrollbarSize from 'dom-helpers/util/scrollbarSize'
 import React, { Component, PropTypes } from 'react'
 import shouldPureComponentUpdate from 'react-pure-render/function'
 
@@ -45,6 +46,11 @@ export default class Grid extends Component {
      * Number of columns in grid.
      */
     columnsCount: PropTypes.number.isRequired,
+
+    /**
+     * Number of columns to fix to the left of the grid while scrolling.
+     */
+    columnsLocked: PropTypes.number,
 
     /**
      * Either a fixed column width (number) or a function that returns the width of a column given its index.
@@ -103,6 +109,11 @@ export default class Grid extends Component {
      * Number of rows in grid.
      */
     rowsCount: PropTypes.number.isRequired,
+
+    /**
+     * Number of rows to fix along the top of the grid while scrolling.
+     */
+    rowsLocked: PropTypes.number,
 
     /** Horizontal offset. */
     scrollLeft: PropTypes.number,
@@ -365,9 +376,10 @@ export default class Grid extends Component {
       noContentRenderer,
       overscanColumnsCount,
       overscanRowsCount,
-      renderCell,
       rowsCount,
-      width
+      width,
+      rowsLocked,
+      columnsLocked
     } = this.props
 
     const {
@@ -382,6 +394,7 @@ export default class Grid extends Component {
     if (height > 0 && width > 0) {
       const visibleColumnIndices = getVisibleCellIndices({
         cellsCount: columnsCount,
+        cellsLocked: columnsLocked,
         cellMetadata: this._columnMetadata,
         containerSize: width,
         currentOffset: scrollLeft
@@ -389,6 +402,7 @@ export default class Grid extends Component {
 
       const visibleRowIndices = getVisibleCellIndices({
         cellsCount: rowsCount,
+        cellsLocked: rowsLocked,
         cellMetadata: this._rowMetadata,
         containerSize: height,
         currentOffset: scrollTop
@@ -402,6 +416,7 @@ export default class Grid extends Component {
 
       const overscanColumnIndices = getOverscanIndices({
         cellsCount: columnsCount,
+        cellsLocked: columnsLocked,
         overscanCellsCount: overscanColumnsCount,
         startIndex: this._renderedColumnStartIndex,
         stopIndex: this._renderedColumnStopIndex
@@ -409,6 +424,7 @@ export default class Grid extends Component {
 
       const overscanRowIndices = getOverscanIndices({
         cellsCount: rowsCount,
+        cellsLocked: rowsLocked,
         overscanCellsCount: overscanRowsCount,
         startIndex: this._renderedRowStartIndex,
         stopIndex: this._renderedRowStopIndex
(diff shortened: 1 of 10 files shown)


source/Grid/Grid.js
```diff
@@ -740,10 +781,11 @@ export default class Grid extends Component {
     // This causes a series of rapid renders that is slow for long lists.
     // We can avoid that by doing some simple bounds checking to ensure that scrollTop never exceeds the total height.
     const { height, width } = this.props
+    const scrollbarSize = getScrollbarSize()
     const totalRowsHeight = this._getTotalRowsHeight()
     const totalColumnsWidth = this._getTotalColumnsWidth()
-    const scrollLeft = Math.min(totalColumnsWidth - width, event.target.scrollLeft)
-    const scrollTop = Math.min(totalRowsHeight - height, event.target.scrollTop)
+    const scrollLeft = Math.min(totalColumnsWidth - width + scrollbarSize, event.target.scrollLeft)
+    const scrollTop = Math.min(totalRowsHeight - height + scrollbarSize, event.target.scrollTop)
```
jquense: Without this check the last 17px or so of the scroll isn't called, which causes the fixed columns to shift b/c their offsets aren't updated.

source/Grid/Grid.js
```diff
@@ -10,6 +10,7 @@ import {
 } from '../utils'
 import cn from 'classnames'
 import raf from 'raf'
+import getScrollbarSize from 'dom-helpers/util/scrollbarSize'
```
jquense: I added a dependency bc it seemed better then copying the code, especially since `dom-helpers` is structured in a way that won't bring in extra code. https://github.com/react-bootstrap/dom-helpers/blob/master/src/util/scrollbarSize.js

source/Grid/Grid.js
```diff
@@ -496,6 +496,47 @@ export default class Grid extends Component {
     )
   }
 
+  renderColumns (childrenToDisplay, rowIndex, top) {
+    let { columnsLocked } = this.props
```
jquense: I'm not sure if there was a strong reason to inline all the rendering, but the repetition now was getting out of hand so I moved them to separate functions

source/demo/Application.js
```diff
@@ -22,7 +22,7 @@ const HIGH_ORDER_COMPONENTS = ['AutoSizer', 'ColumnSizer', 'InfiniteLoader', 'Sc
 const list = Immutable.List(generateRandomList())
 
 class Application extends Component {
-  shouldComponentUpdate = shouldPureComponentUpdate
+  shouldComponentUpdate = shouldPureComponentUpdate;
```
jquense: oops not sure why I did this, can remove

jquense:
...and wheel scrolling seems odd, let me look at that.

jquense:
whelp browsers are terrible.  I'm not sure why I'm getting flicker/lag on updating the locked offsets when using the wheel. or scrolling in general in FF (why is perf in general so bad on FF?). 

I'm looking into it, perhaps it's dropping frames, though  that seems weird for a single wheel event.     Any insights would be nice if folks have them :P

bvaughn:
Hey @jquense,

Thanks for taking the initiative to put this PR together. Things are a bit hectic for me this week. (Lots going on at the day-job and I'm also in the process of trying to close on a house.) So I can't promise I'll get to the review as quickly as I'd like, but I'll try and look over it sometime within the next couple of days and give you my thoughts. :)

jquense:
Np :) life gets busy.

jquense:
OK post-mortem from my looking into the scroll wheel issues, and flicker on FF. 

Adding a noop onWheel listener completely fixes wheel flicker for Chrome (on windows) I assume because it forces...something...different in the way it repaints.

FireFox is a lot more difficult. What happens is that the fixed columns/rows update much slower than the actual scroll perceived scroll position so the fixed cells lag behind. I am guessing is due to [async handling of scrolling](https://staktrace.com/spout/entry.php?id=834). I have tried a bunch of different techniques with little or qualified success (why is this so hard?)

1. switching to `transform`/`transform3d`: (perhaps a small improvement, but not a solution)
2. manually updating the DOM nodes styles, thinking that the RAF was too slow for some reason (similar to this approach taken [here](http://adazzle.github.io/react-data-grid/examples.html#/fixed) seems to work, but didn't work here)
3. Splitting the render area into 4 quadrants, and absolutely positioning them, effectively reducing the positioned elements to 4, thinking that maye the large amount of absolute positioning was causing frame drops. (no luck)
4. using `position: sticky` when supported. This __does__ solve the filter, however since at the top of the scroll elements are positioned: `relative`, instead of `absolute` it makes it really hard to layout the cells at that point b/c they are now inline.

What is confusing to me is that this same approach works just fine [here](http://adazzle.github.io/react-data-grid/examples.html#/fixed), but I am having no luck, I am assuming that something I can't see is pushing the scrolling to the main thread (?) in their case.

Ultimately `position: sticky` is the best bet, but my CSS Fu, is too weak to figure out how to handle both frozen rows and columns at the same time, perhaps its just acceptable to only allow fixed columns, which is the common case, and the one not handled sufficiently (to me) by ScrollSync

bvaughn:
Thanks for the detailed post mortem. :)

> What is confusing to me is that this same approach works just fine here, but I am having no luck, I am assuming that something I can't see is pushing the scrolling to the main thread (?) in their case.

The demo you linked to only windows the vertical axis, not both. I think that simplifies things a lot.

> Ultimately `position: sticky` is the best bet, but my CSS Fu, is too weak to figure out how to handle both frozen rows and columns at the same time, perhaps its just acceptable to only allow fixed columns, which is the common case, and the one not handled sufficiently (to me) by ScrollSync

I'm not sure I fully understand this sentence or what you're implying/suggesting. Mind elaborating?

jquense:
> I'm not sure I fully understand this sentence or what you're implying/suggesting. Mind elaborating?

Sure, I'll try. first some more context: https://developer.mozilla.org/en-US/docs/Mozilla/Performance/Scroll-linked_effects

Ok, `position: sticky` is mostly designed to handle these sort of cases. The caveat is that at the "rest" position, (when scrollLeft is < cell `left`, or scrollTop < cell `top`), `position: sticky` acts like `position: relative` which doesn't work with the current  setup because each Cell is positioned absolutely in the container.

The simple way to address that is to render locked cells into a single div. and apply `position: sticky` to the whole set of of cells. However, that is complicated by allowing both frozen columns and rows at the same time, which essentially creates three distinct locked areas to handle positioning.

1. the cells locked on the x axis
2. the cells locked on the y axis
3. the intersection: cells that are locked on both axis's (top left section of the grid)

So now even if you wrap the locked areas in containers, you end up with three containers, which suffer from the inline layout issue originally. 

There may be a way to handle that case gracefully, but my CSS skills failed me when trying to figure it out.

One solution is just to eliminate the complexity, and only allow locking along one axis at a time, I think that's preferable to not supporting this, but a bit disappointing. Perhaps `sticky` isn't the right solution, it seems sort of strange that the scroll behavior of FF just doesn't allow this sort of thing.

> [...] one not handled sufficiently (to me) by ScrollSync

The problem with the ScrollSync approach (beyond dealing with scrollbar padding), is that good UX for fixed columns (not headers) requires that the horizontal scrollbar extend the entire contrainer, not just the section that scrolls. You don't want the frozen sections to feel like tacked on sectyions to the left.

PS:
if you want to take it for a spin: https://github.com/jquense/react-virtualized/tree/locked-translate that branch uses `sticky` if its supported (FF, Safari).

jquense:
more research on the none dev FF suggests the flicker is probably actually do with dropped frames and classic jank, since the async scroll stuff isn't enabled yet (by default?).

My other thought is that perhaps it has something to do with the scroll animations...the `left/top` offsets seem to be updated more incrementally then, say, Chrome...I am way out of my depth at this point, perhaps someone with more experience in these areas as some thoughts? @jlongster?

bvaughn:
I _think_ PR #161 was meant as an alternative to this PR and so I'm going to close this. Please correct me if I'm wrong and we can talk more. :)

PR #159, opened 2016-03-17
Version 6.0

### LINKED ISSUE #156: VirtualScroll takes over arrow keys
I am rendering an input in an item, and the cursor inside the input cannot be controlled with the left/right arrows keys. What is taking over this behavior? Not much more to say about it, I'm sure you know what I'm talking about :) If not I can post a small test case.

Is there a way to turn it off?

bvaughn:
### Completed
* Removed arrow-key snapping behavior from `Grid` and moved it into a new `ArrowKeyStepper` HOC; resolves #156.
* Removed scroll-to-* and set-scroll-* methods from `FlexTable`, `Grid`, and `VirtualScroll`; properties should be used instead of these methods.
* Replaced `react-pure-render` with [`shallowCompare`](https://facebook.github.io/react/docs/shallow-compare.html).

Resolves #156


jquense:
this is sweet!

PR #161, opened 2016-03-17
Clean up Grid to allow more complex ScrollSync

jquense:
Soooo this is a a completely different approach to getting locked rows and columns from #155. There are 3 bits to the PR:

1. a fix to `onScroll` under-reporting scrollLeft and scrollTop at the end of bottom of the container.
2. Skip rendering cells when `renderCell` returns `false`, `null`, or `undefined` avoiding rendering empty cells.
3. ~~Add the `noop` onWheel handler since it seems to reduce scroll lag when updating components based on scroll position (more in #155 on that). Perhaps that is better left to the user tho, since I am unsure _why_ it helps.~~

Together, you can efficiently create complex frozen row,table behavior by overlaying the frozen Grids over the main scrolling one, and adjust their heights to compensate for the scroll-bar. I've updated the `ScrollSync` example to demonstrate this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "prebuild": "npm run lint",
     "postpublish": "npm run deploy",
     "prepublish": "npm run build",
-    "start": "webpack-dev-server --hot --inline --config webpack.config.dev.js",
+    "start": "cross-env NODE_ENV=development webpack-dev-server --hot --inline --config webpack.config.dev.js",
     "test": "npm run lint && npm run test:unit",
     "test:unit": "cross-env NODE_ENV=test karma start",
     "watch": "watch 'clear && npm run lint -s && npm run test -s' source"
@@ -125,6 +125,7 @@
   },
   "dependencies": {
     "classnames": "^2.2.3",
+    "dom-helpers": "^2.4.0",
     "raf": "^3.1.0",
     "react-pure-render": "^1.0.2"
   },
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -10,6 +10,7 @@ import {
 } from '../utils'
 import cn from 'classnames'
 import raf from 'raf'
+import getScrollbarSize from 'dom-helpers/util/scrollbarSize'
 import React, { Component, PropTypes } from 'react'
 import shouldPureComponentUpdate from 'react-pure-render/function'
 
@@ -207,6 +208,8 @@ export default class Grid extends Component {
 
   componentDidMount () {
     const { scrollLeft, scrollToColumn, scrollTop, scrollToRow } = this.props
+
+    this._scrollbarSize = getScrollbarSize()
 
     if (scrollLeft >= 0 || scrollTop >= 0) {
       this.setScrollPosition({ scrollLeft, scrollTop })
@@ -427,6 +430,13 @@ export default class Grid extends Component {
           let columnDatum = this._columnMetadata[columnIndex]
           let renderedCell = renderCell({ columnIndex, rowIndex })
           let key = `${rowIndex}-${columnIndex}`
+
+          // any other falsey value will be rendered
+          // as a text node by React
+          if (renderedCell == null || renderedCell === false) {
+            continue
+          }
+
           let child = (
             <div
               key={key}
@@ -740,10 +750,11 @@ export default class Grid extends Component {
     // This causes a series of rapid renders that is slow for long lists.
     // We can avoid that by doing some simple bounds checking to ensure that scrollTop never exceeds the total height.
     const { height, width } = this.props
+    const scrollbarSize = this._scrollbarSize
     const totalRowsHeight = this._getTotalRowsHeight()
     const totalColumnsWidth = this._getTotalColumnsWidth()
-    const scrollLeft = Math.min(totalColumnsWidth - width, event.target.scrollLeft)
-    const scrollTop = Math.min(totalRowsHeight - height, event.target.scrollTop)
+    const scrollLeft = Math.min(totalColumnsWidth - width + scrollbarSize, event.target.scrollLeft)
+    const scrollTop = Math.min(totalRowsHeight - height + scrollbarSize, event.target.scrollTop)
 
     // Certain devices (like Apple touchpad) rapid-fire duplicate events.
     // Don't force a re-render if this is the case.
--- a/source/ScrollSync/ScrollSync.example.css
+++ b/source/ScrollSync/ScrollSync.example.css
@@ -1,4 +1,5 @@
 .GridRow {
+  position: relative;
   display: flex;
   flex-direction: row;
 }
@@ -9,6 +10,7 @@
 }
 .LeftSideGridContainer {
   flex: 0 0 75px;
+  z-index: 10;
 }
 
 .LeftSideGrid {
--- a/source/ScrollSync/ScrollSync.example.js
+++ b/source/ScrollSync/ScrollSync.example.js
@@ -5,10 +5,10 @@ import { ContentBox, ContentBoxHeader, ContentBoxParagraph } from '../demo/Conte
 import AutoSizer from '../AutoSizer'
 import Grid from '../Grid'
 import ScrollSync from './ScrollSync'
-import VirtualScroll from '../VirtualScroll'
 import shouldPureComponentUpdate from 'react-pure-render/function'
 import cn from 'classnames'
 import styles from './ScrollSync.example.css'
+import scrollbarSize from 'dom-helpers/util/scrollbarSize'
 
 const LEFT_COLOR_FROM = hexToRgb('#471061')
 const LEFT_COLOR_TO = hexToRgb('#BC3959')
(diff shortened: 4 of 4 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
     "prebuild": "npm run lint",
     "postpublish": "npm run deploy",
     "prepublish": "npm run build",
-    "start": "webpack-dev-server --hot --inline --config webpack.config.dev.js",
+    "start": "cross-env NODE_ENV=development webpack-dev-server --hot --inline --config webpack.config.dev.js",
```
jquense: sorry this is just a quirk from my machine that has NODE_ENV set globally :P

source/Grid/Grid.js
```diff
@@ -472,6 +479,7 @@ export default class Grid extends Component {
         className={cn('Grid', className)}
         onKeyDown={this._onKeyPress}
         onScroll={this._onScroll}
+        onWheel={()=>{}}
```
jquense: it may make more sense to just allow the user to listen onWheel that way they can add the empty handler if they want.

jquense:
bah sorry lint issues, will fix. also this should fix #131

source/Grid/Grid.js
bvaughn: I don't understand why you need to specify anything for `onWheel`. Can you explain? Why is `onWheel={()=>{}}` any different/better than omitting the attribute entirely?

jquense: So FF and Chrome seem to exhibit different behavior if an onWheel handler is registered, specifically, without it the new scroll position repaints, _before_ the `onScroll` event fires. Presumably the browser is doing some sort of off-thread optimization for wheel scrolls that it has to bail out of when you listen for the wheel event.

That is just me pulling guesses out of the air though, I couldn't find anything explaining why it works. I do see a noticeable lag in the scroll updates of the other grids though without it.

bvaughn:
Thanks again for the PR buddy. I'm a little anxious about the `onWheel` issue I commented on. (Only because `onWheel` was a source of performance problems in the past.)

Maybe one of the Mozilla guys would be willing to comment on this. I'll ask on Twitter.

bvaughn:
I asked [here](https://twitter.com/brian_d_vaughn/status/710523261633703936). Feel free to chime in if I misrepresented what you observed. :)

jquense:
I understand the anxiousness, I don't actually think its a good default, A better option is probably to just allow the user to specify an `onWheel` prop and make a note of it potentially helping (if anyone can confirm i'm not crazy)

bvaughn:
Let's see if Markus has any insight into it. He's always been really helpful.

source/Grid/Grid.js
```diff
@@ -427,6 +428,12 @@ export default class Grid extends Component {
           let columnDatum = this._columnMetadata[columnIndex]
           let renderedCell = renderCell({ columnIndex, rowIndex })
           let key = `${rowIndex}-${columnIndex}`
+
+          // any other falsey value will be rendered
+          // as a text node by React
+          if (renderedCell == null || renderedCell === false)
```
bvaughn: Curious what prompted this change. Do people actually render empty cells? :) I guess it doesn't hurt to support this use-case. It's just not one I've considered.

source/Grid/Grid.js
```diff
@@ -740,10 +748,11 @@ export default class Grid extends Component {
     // This causes a series of rapid renders that is slow for long lists.
     // We can avoid that by doing some simple bounds checking to ensure that scrollTop never exceeds the total height.
     const { height, width } = this.props
+    const scrollbarSize = getScrollbarSize()
     const totalRowsHeight = this._getTotalRowsHeight()
     const totalColumnsWidth = this._getTotalColumnsWidth()
-    const scrollLeft = Math.min(totalColumnsWidth - width, event.target.scrollLeft)
-    const scrollTop = Math.min(totalRowsHeight - height, event.target.scrollTop)
+    const scrollLeft = Math.min(totalColumnsWidth - width + scrollbarSize, event.target.scrollLeft)
+    const scrollTop = Math.min(totalRowsHeight - height + scrollbarSize, event.target.scrollTop)
```
bvaughn: Nice :)

source/Grid/Grid.js
jquense: Its a bit of an optimization and helpful in the ScrollSync case, you can see in the updated Example the renderBodyCell method doesn't render cells that are "locked" and displayed in the scynced grid overlayed on the left. Not that having the duplication would work, but this is a bit cleaner i think.

source/ScrollSync/ScrollSync.example.css
```diff
@@ -9,6 +10,7 @@
 }
 .LeftSideGridContainer {
   flex: 0 0 75px;
+  z-index: 10;
```
bvaughn: What prompted these CSS changes? (Just curious)

bvaughn: Disregard. I see how you've updated the markup below. :)

bvaughn:
[Markus says](https://twitter.com/markusstange/status/710528325299478528):
>  It will definitely put you on slow paths in many browsers, which may delay the paint in Chrome (it won't in Firefox). In Firefox, it'll delay the start of the scroll (if the page is unresponsive), but we'll still composite before onscroll

source/Grid/Grid.js
```diff
@@ -63,6 +64,12 @@ export default class Grid extends Component {
     noContentRenderer: PropTypes.func.isRequired,
 
     /**
+     * Optional callback invoked on mouse-wheel activity.
+     * (event: SyntheticEvent): void
+     */
+    onWheel: PropTypes.func,
```
bvaughn: I'd like to remove this change for the time being. Markus' response on Twitter made me more anxious. This seems like the sort of change that, if we do, we should try in total isolation. (I'd also want to test it a lot more before I felt comfortable.)

jquense:
o good i'm not crazy :P it does seem like a bad idea though to force the browser into slow scroll path if its avoidable...I wonder what the proposed solution is for this sort of thing, since it seems to be the direction browsers are going in handling scroll events.

source/Grid/Grid.js
jquense: so I changed it to just accept an onWheel prop but not default it to anything, so you aren't getting any de-opt and I think its reasonable for users to want to listen to that event in general (I have for other reasons)

If you think its a bad idea though to allow listening for wheel events at all i can remove it.

bvaughn:
I wish I knew. Focusing on (mostly scrolling related) performance improvements these past few weeks has been mentally exhausting. I'm really appreciative to the Mozilla guys for being so responsive and helpful. I wish I had similar points of contact with the Webkit folks.

source/Grid/Grid.js
bvaughn: Understood. I'm not necessarily against opt-in experimental things like this... it just feels a bit odd. We don't pass-through other event handlers. The only other similar props (`onScroll` and `onSectionRendered`) are used in a totally different way- to notify of changes to the `Grid` and not its inner components.

source/Grid/Grid.js
bvaughn: Thinking about this approach a little more. I think it's safe (based on my testing) but- I want to point out that it doesn't tell us if there is _actually_ a scrollbar or not, just the size of what a scrollbar would be _if_ it were present. (I'm doing a similar check in `FlexTable` but I'm using `offsetWidth` and `clientWidth` directly to determine whether there's a scrollbar.) This approach wouldn't work in `FlexTable` (because it's important whether a scrollbar is _actually_ visible or not).

Okay. Anyway. All of that to say that... since we're not checking whether the scrollbar is actually present or not, only what it's size _would_ be, we might as well just do this calculation only once- in `componentDidMount`. That gets called before the first `onScroll` anyway so it should be safe. We can just cache it as a private property.

jquense:
> performance improvements these past few weeks has been mentally exhausting

I can Imagine, doing two days of it has left my brain really tired. So much of this stuff feels like black magic.

I do really appreciate the work you've being doing though. This stuff is hard, thankless, and very few folks (myself included) produce something with as much, performance, polish and attention to detail as you have, so thank you!

source/Grid/Grid.js
jquense: I agree it doesn't need to be recalcualted over and over, luckily the helper does already cache the result, so we don't need to worry about it here (https://github.com/react-bootstrap/dom-helpers/blob/master/src/util/scrollbarSize.js). I should have noted that, sorry!

I can still move it to a private property if that is helpful, it may just be clearer for folks looking at the code later.

source/Grid/Grid.js
jquense: yeah it is a bit odd, but then again I think the grid _should_ listen to more of the html events. 

In any case, I'm happy to remove it if it's a blocker for you. So you want me to remove?

source/Grid/Grid.js
bvaughn: Ha! So it does. My apologies. :)

It might prevent an incorrect assumption later so yeah. I have a slight preference to cache it for clarity. :)

source/Grid/Grid.js
bvaughn: > I think the grid _should_ listen to more of the html events.

Mind elaborating? Since relevant events (keyboard events, mouse events, etc.) bubble I didn't want to go down the path of having to pass-through all of the listeners. It didn't seem necessary.

bvaughn:
Well thank _you_ :)

source/Grid/Grid.js
bvaughn: I'm very much on the fence about this. My reasons for leaning towards removing it are:

1. It might not _help_ performance; according to what Markus said it might actually make it _worse_.
1. It seems inconsistent (which I care about from an API point of view) to offer a single event-handler pass-through like this.

jquense: Up to you :) there isn't a huge downside to removing it from my perspective, if adding the listener is something we want to do I can attach it manually with a ref.

bvaughn: Let's remove it for now. Baby steps with this sort of thing. I've been experimenting _a lot_ with regard to scrolling performance. This is another combination to add to my list of things to play with.

jquense:
ok! made the scrollbar size caching explicit and removed the onWheel handler. anything else I can do?

bvaughn:
No. Thank you. This looks great.

bvaughn:
Just release 5.5.6 with these changes. :)

bvaughn:
By the way @jquense: In my own testing, I've observed that even adding an empty `onmousewheel=""` attribute to parent DOM element that React is mounted to is sufficient to force Chrome down the path of altered scrolling behavior. (So if you want to experiment more with that attribute, react-virtualized doesn't need to do anything to _directly_ support it.)

PR #163, opened 2016-03-20
Replaced react-pure-render with react-addons-shallow-compare

bvaughn:
* Replaced `react-pure-render` with [`shallowCompare`](https://facebook.github.io/react/docs/shallow-compare.html).

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
     "classnames": "^2.2.3",
     "dom-helpers": "^2.4.0",
     "raf": "^3.1.0",
-    "react-pure-render": "^1.0.2"
+    "react-addons-shallow-compare": "^0.14.7"
   },
   "peerDependencies": {
     "react": "^0.14.0 || ^15.0.0-rc.1",
--- a/source/ArrowKeyStepper/ArrowKeyStepper.example.js
+++ b/source/ArrowKeyStepper/ArrowKeyStepper.example.js
@@ -5,13 +5,11 @@ import { ContentBox, ContentBoxHeader, ContentBoxParagraph } from '../demo/Conte
 import ArrowKeyStepper from './ArrowKeyStepper'
 import AutoSizer from '../AutoSizer'
 import Grid from '../Grid'
-import shouldPureComponentUpdate from 'react-pure-render/function'
+import shallowCompare from 'react-addons-shallow-compare'
 import cn from 'classnames'
 import styles from './ArrowKeyStepper.example.css'
 
 export default class ArrowKeyStepperExample extends Component {
-  shouldComponentUpdate = shouldPureComponentUpdate
-
   static propTypes = {
     list: PropTypes.instanceOf(Immutable.List).isRequired
   }
@@ -76,6 +74,10 @@ export default class ArrowKeyStepperExample extends Component {
         </ArrowKeyStepper>
       </ContentBox>
     )
+  }
+
+  shouldComponentUpdate (nextProps, nextState) {
+    return shallowCompare(this, nextProps, nextState)
   }
 
   _getColumnWidth (index) {
--- a/source/ArrowKeyStepper/ArrowKeyStepper.js
+++ b/source/ArrowKeyStepper/ArrowKeyStepper.js
@@ -1,13 +1,11 @@
 /** @flow */
 import React, { Component, PropTypes } from 'react'
-import shouldPureComponentUpdate from 'react-pure-render/function'
+import shallowCompare from 'react-addons-shallow-compare'
 
 /**
  * This HOC decorates a virtualized component and responds to arrow-key events by scrolling one row or column at a time.
  */
 export default class ArrowKeyStepper extends Component {
-  shouldComponentUpdate = shouldPureComponentUpdate
-
   static propTypes = {
     children: PropTypes.func.isRequired,
     className: PropTypes.string,
@@ -48,6 +46,10 @@ export default class ArrowKeyStepper extends Component {
         })}
       </div>
     )
+  }
+
+  shouldComponentUpdate (nextProps, nextState) {
+    return shallowCompare(this, nextProps, nextState)
   }
 
   _onKeyDown (event) {
--- a/source/AutoSizer/AutoSizer.example.js
+++ b/source/AutoSizer/AutoSizer.example.js
@@ -4,12 +4,10 @@ import React, { Component, PropTypes } from 'react'
 import { ContentBox, ContentBoxHeader, ContentBoxParagraph } from '../demo/ContentBox'
 import AutoSizer from './AutoSizer'
 import VirtualScroll from '../VirtualScroll'
-import shouldPureComponentUpdate from 'react-pure-render/function'
+import shallowCompare from 'react-addons-shallow-compare'
 import styles from './AutoSizer.example.css'
 
 export default class AutoSizerExample extends Component {
-  shouldComponentUpdate = shouldPureComponentUpdate
-
   static propTypes = {
     list: PropTypes.instanceOf(Immutable.List).isRequired
   }
@@ -77,6 +75,10 @@ export default class AutoSizerExample extends Component {
         </div>
       </ContentBox>
     )
+  }
+
+  shouldComponentUpdate (nextProps, nextState) {
+    return shallowCompare(this, nextProps, nextState)
   }
 
   _rowRenderer (index) {
--- a/source/AutoSizer/AutoSizer.js
+++ b/source/AutoSizer/AutoSizer.js
@@ -1,6 +1,6 @@
 /** @flow */
 import React, { Component, PropTypes } from 'react'
-import shouldPureComponentUpdate from 'react-pure-render/function'
+import shallowCompare from 'react-addons-shallow-compare'
 
 /**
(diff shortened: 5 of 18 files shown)


oliviertassinari:
That may be silly, be we can actually have the same behavior with no other dependency:
```js
import shallowEqual from 'fbjs/lib/shallowEqual';
```
At least, you need *npm@3* and *react* to be installed to get it working. E.g: https://github.com/oliviertassinari/react-event-listener/blob/master/src/index.js#L4.

bvaughn:
In order to support React 0.14, all react* dependencies (`react`, `react-dom`, `react-addons-shallow-compare`, etc) are declared as `peerDependencies` instead of dependencies. Npm should nag you to install them (if you haven't already) as part of installing `react-virtualized`. I mention this _briefly_ in the main README. 😁

bvaughn:
The way React chooses to bundle its own code internally is not guaranteed to stay the same between versions for example. They started with that approach initially but IIR plan to migrate modules out into separate projects. (Even if they didn't plan to, they still reserve the right to.)

oliviertassinari:
@bvaughn Alright, sounds fair. I have installed the peer dependency. Thanks for the answer!

bvaughn:
No problem of course! It's a fairly common question.

PS Whatcha workin' on? :wink:

oliviertassinari:
@bvaughn I'm replacing *react-list* with *react-virtualized* on my app. I want to see how it perform :v:. I can think about some soo many application on @material-ui and on @doctolib 😉 .

bvaughn:
Gotcha! I was wondering if you were starting on a material-ui integration yet. Happy to help with that one once you're ready!

In the meanwhile let me know how your experiment goes

oliviertassinari:
@bvaughn I have finally finished the migration from [react-list](https://github.com/orgsync/react-list) to this lib. I'm pretty happy with the final result 😄 .

It's smooth on Cordova Android using the development mode. That's quite the worse environment I can test 🚀 .
sept -26-2016 22-21-39
Still, that was really challenging for me 😁 .
My `render` method looks like this. I feel like this approach isn't documented:
```jsx
    return (
      <ScrollView onScroll={this.handleScroll} fullHeight={true}>
        <LayoutBody fullHeight={true}>
          <AutoSizer>
            {({height, width}) => (
              <div style={wrapperStyle}>
                <Paper
                  rounded={false}
                  transitionEnabled={false}
                  style={{
                    width: width,
                    position: 'relative',
                  }}
                  data-test="ExpenseList"
                >
                  <List
                    autoHeight={true}
                    height={height}
                    width={width}
                    scrollTop={this.state.scrollTop}
                    rowCount={expenses.size}
                    rowRenderer={this.rowRenderer}
                    rowHeight={LIST_ITEM_HEIGHT}
                    sortBy={expenses}
                  />
                </Paper>
              </div>
            )}
          </AutoSizer>
        </LayoutBody>
      </ScrollView>
    );
```


I have two feedbacks soo far:
1. Why exposing the `style` property in the `rowRenderer` function?
My `rowRenderer` function basically returns a `<div style={style} />` with a single child: my row.
2. I feel like the `rowRenderer` is called too often. Couldn't we memoize the function? That's what I'm doing in [react-swipeable-views](https://github.com/oliviertassinari/react-swipeable-views) for my slide children.
On userland, I ended up adding a pure logic for the rows.
That massively speeds up the process. To be even more efficient, I could have memoized the function. But that wasn't necessary.

bvaughn:
Thanks for sharing the video, @oliviertassinari. It looks great!

> My `render` method looks like this. I feel like this approach isn't documented:

It isn't! That's because I haven't really messed around with react-native much. Would you like me to add your example code to the repo? I'd love to! Mind tossing up a quick PR?

> 1. Why exposing the `style` property in the `rowRenderer` function?
My `rowRenderer` function basically returns a `<div style={style} />` with a single child: my row.

I only recently added the `style` and `key` properties (in version 8 actually, you can read more about that decision [here](https://github.com/bvaughn/react-virtualized/issues/386)). Essentially this was done in order to remove an extra/unnecessary DOM element from each cell in a `Grid`. A side effect was also to enable users complete customization/control over cell-styling. Too often before I saw cells that looked like this:

```jsx
<div className='Grid__cell'>
  <div className='userCell'>
    {/* content */}
  </div>
</div>
```

> 2. I feel like the `rowRenderer` is called far too often. Couldn't we memoize the function? That's what I'm doing in [react-swipeable-views](https://github.com/oliviertassinari/react-swipeable-views) for my slide children.
On userland, I ended up adding a pure logic for the rows. That massively speeds up the process.
To be even more efficient, I could have memoized my rows. But that wasn't necessary.

This is a tricky thing to get right. For what it's worth, I _do_ temporarily cache rendered rows when a user is scrolling (which is when performance is most important). I don't cache them when the list is resting though because I think it would cause more confusion and the perf gains would not be worth it (at least not in my own testing). I already get a lot of questions like "my array changed but my list isn't updating" because of react-virtualized using a pure render approach. I think caching would make this worse.

However you _can_ enable caching yourself if you'd like to play around with it! 😁 I'd love to hear your findings if you do! The simplest (albeit a bit hackish) way of doing this would be like so:

```jsx
import { defaultCellRangeRenderer, Grid } from 'react-virtualized'

function cellRangeRenderer (cellRangeRendererProps) {
  const { isScrolling } = cellRangeRendererProps

  function cellRenderer (cellRendererProps) {
    return props.cellRenderer({
      ..cellRendererProps,
      isScrolling // Pass the real scrolling flag through to renderers
    })
  }

  return defaultCellRangeRenderer({
    ...cellRangeRendererProps,
    isScrolling: true, // Trick default renderer into always-caching
    cellRenderer
  })
}

function CustomizedGrid (props) {
  return (
    <Grid
      cellRangeRenderer={cellRangeRenderer}
      {...props}
    />
  )
}
```

oliviertassinari:
> Would you like me to add your example code to the repo?

Sure! I would rather be using a documented pattern 😁 .

> A side effect was also to enable users complete customization/control over cell-styling

Oh right, doesn't sound that bad then 👍 .

> I do temporarily cache rendered rows when a user is scrolling 

I haven't noticed this. Great idea!

> I already get a lot of questions like "my array changed but [...]

I undertand your concern. I think that it's only an option is users embrace immutability.
Would it make sense to add a property to enable an *immutable* mode?

Hum, I'm not sure to understand how I should be using this example 😁 .

bvaughn:
>> I already get a lot of questions like "my array changed but [...]

>I understand your concern. I think that it's only an option if users embrace immutability.
Would it make sense to add a property to enable an immutable mode?

I don't think this is an immutable issue actually. (The same questions come from people using Immutable JS for example.) The real issue is that react-virtualized doesn't have access to the array/list/map/whatever at all. It just has numbers (eg `columnCount`, `rowCount`). If the numbers change, it can infer that _something_ has changed (but not what). If the underlying data changes and the numbers do not, it has no idea at all.

> Hum, I'm not sure to understand. How should I be using this example with a `List` 😁 ?

More or less the same. :smile:

```jsx
import { defaultCellRangeRenderer, List } from 'react-virtualized'

const cellCache = {}

function cellRangeRenderer (cellRangeRendererProps) {
  const { cellRenderer } = cellRangeRendererProps

  function cellRendererWrapper (cellRendererProps) {
    return cellRenderer({
      ...cellRendererProps,
      isScrolling: false
    })
  }

  return defaultCellRangeRenderer({
    ...cellRangeRendererProps,
    cellCache, // Provide your own, permanent cache
    isScrolling: true, // Trick default renderer into always-caching
    cellRenderer: cellRendererWrapper
  })
}

function CustomizedList (props) {
  return (
    <List
      cellRangeRenderer={cellRangeRenderer}
      {...props}
    />
  )
}
```

oliviertassinari:
> The real issue is that react-virtualized doesn't have access to the array/list/map/whatever at all

Isn't what the `sortBy` property is for? Well, there is a small limitation. You can only provide **one** source of data.

> More or less the same.

Thanks! After doing some measurement (going back and forth as fast as possible), I couldn't find a way to consistently improve the metrics.
I couldn't measure the impact of `isScrolling` nor of a memoization on my `side`.
I'm not properly tooled to say anything here 😁 .

bvaughn:
`sortBy` is only a `Table` prop (not `List` or `Grid`) so it only applies in some situations. 😄

Unfortunately I know what you mean. I think RV is kind of at the point of diminishing returns with optimizations. I'm sure there are more that could be made- and maybe some that would make substantial impact- but it seems like most of the things I try lately make negligible or indeterminate impact. 😅

oliviertassinari:
Oh, I though `sortBy` was always available!
Well, I'm gonna rename this property `immutableBinding` that has the same effect, still solve my cache update issue and is more explicit 👍 .

bvaughn:
Nope it's `Table` only, used by the default header renderer to render the correct sort icons. 😄

Sounds good!

oliviertassinari:
@bvaughn Thanks for your help! At least now, I have a better understanding of the API exposed and how it's could be integrated with Material-UI.
We are rewriting all of our component in the `next` branch. It's time to do it right.

bvaughn:
That's so exciting!

Ping me if you run into issues or want a second opinion or anything. I'm very eager to see RV and Material-UI work together. :)

PR #175, opened 2016-04-05
Fixed several a11y issues for FlexTable

globexdesigns:
When using the FlexTable component with react-a11y (https://github.com/reactjs/react-a11y) enabled, we see a ton of warning messages about bad a11y practices.

This PR addresses some low-hanging fruit:

- Adds support for passing through the `aria-label` prop to the Grid child component
- Adds support for passing through the `aria-label` prop to each column header (and will try to set a default value if one isn't provided by using either the `label` or the `dataKey` value)
- Sets `role` and `tabIndex` values for column headers and rows since those must be defined for any element with `onClick`

### DIFF
--- a/source/FlexTable/FlexColumn.js
+++ b/source/FlexTable/FlexColumn.js
@@ -85,6 +85,9 @@ export default class Column extends Component {
   }
 
   static propTypes = {
+    /** Optional aria-label value to set on the column header */
+    'aria-label': PropTypes.string,
+
     /** Optional CSS class to apply to cell */
     cellClassName: PropTypes.string,
 
--- a/source/FlexTable/FlexTable.js
+++ b/source/FlexTable/FlexTable.js
@@ -199,6 +199,7 @@ export default class FlexTable extends Component {
         )}
 
         <Grid
+          aria-label={this.props['aria-label']}
           ref='Grid'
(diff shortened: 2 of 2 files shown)


bvaughn:
Hey @globexdesigns. Thanks for submitting this PR!

Give me a little time to mull over it. I think I might actually like to make a change or two to `VirtualScroll` and the underlying `Grid` as well now (to add the proper `role` attributes and to support the `aria-label`) that you've pointed this out.

bvaughn:
I'm going to go ahead and merge this and then build on top of it to add a few more roles. I'll hopefully make a bugfix release with this change at some point this evening. :)

globexdesigns:
Thanks a lot for the fast turnaround!

bvaughn:
[6.0.5](https://github.com/bvaughn/react-virtualized/blob/master/CHANGELOG.md#605) was just released with this change. Gave you a shout-out in the release notes. Thanks again!

bvaughn:
PS. Pushed a small update in [6.0.6](https://github.com/bvaughn/react-virtualized/blob/master/CHANGELOG.md#606) that conditionally adds the aria properties only when click (or sort) handlers are present. Just a small optimization but it seemed worth doing.

PR #179, opened 2016-04-06
fix InfiniteLoader example

olslash:
Just a little thing i noticed while setting this up -- list was initialized {} which doesn't have a "length" property


bvaughn:
Ha! Thanks :)

PR #180, opened 2016-04-07
Fix dependency ranges

cesarandreu:


bvaughn:
Thanks! 6.0.8 just released with this updated package.

PR #184, opened 2016-04-11
Move shallow compare addon to peer dependencies

cesarandreu:
https://twitter.com/theridgway/status/719627580731629568

I think if you have it in dependencies you can end up with multiple versions?

PR #185, opened 2016-04-12
Add missing devDependency

cesarandreu:
If you run `npm i`, it won't get you react-addons-shallow-compare


bvaughn:
Haha, right. Broke in #184

PR #192, opened 2016-04-17
Fixed sortBy field name in FlexTable example

FelixWeis:
Current demo does not use Immutable.List.sortBy correctly.

https://bvaughn.github.io/react-virtualized/?component=FlexTable


bvaughn:
Good catch. Thanks. :)

PR #206, opened 2016-04-29
Version 7

bvaughn:
### Backwards incompatible changes
- [x] [Clean up property names](https://github.com/bvaughn/react-virtualized/wiki/Version-7-Roadmap#clean-up-property-names-codemod)
- [x] [Use named function arguments](https://github.com/bvaughn/react-virtualized/wiki/Version-7-Roadmap#use-named-function-arguments)

### Backwards compatible changes
- [x] [Cells are cached while scrolling](https://github.com/bvaughn/react-virtualized/wiki/Version-7-Roadmap#cells-are-cached-while-scrolling)
- [x] [Cell renderers receive an `isScrolling` parameter](https://github.com/bvaughn/react-virtualized/wiki/Version-7-Roadmap#cell-renderers-receive-an-isscrolling-parameter)
- [x] [Defer cell measurement until necessary](https://github.com/bvaughn/react-virtualized/wiki/Version-7-Roadmap#defer-cell-measurement-until-necessary)
- [x] [Support just-in-time calculated cell sizes](https://github.com/bvaughn/react-virtualized/wiki/Version-7-Roadmap#support-just-in-time-calculated-cell-sizes)
- [X] [Support custom inline styles (in addition to custom CSS)](https://github.com/bvaughn/react-virtualized/wiki/Version-7-Roadmap#more-flexible-styling)

### Miscellaneous
- [ ] [facebook/jscodeshift](https://github.com/facebook/jscodeshift) cod-mods for easy upgrade
- [x] Document intended usage and limitations of `CellMeasurer` component


MikeMcElroy:
Big +1 on named arguments -- couldn't find a way to comment on that on the wiki.

MikeMcElroy:
Would you accept PRs on some of these conversions for this branch?  I'm already dinking around in FlexTable & FlexColumn for work stuff, so if you wouldn't mind, I could take on those methods you have slated for updating in the [wiki](https://github.com/bvaughn/react-virtualized/wiki/Version-7-Roadmap).

bvaughn:
Hey @MikeMcElroy! Thanks for the input and the offer to help! That's awesome. :)

I'm working my way through the named parameters this evening. Just finished `FlexColumn#cellDataGetter`.

I'd be happy to share the load though I wonder if we might step on each other's toes a bit. If you're working this evening, feel free to send me some PRs into this branch- maybe work your way from the bottom of the table up? :)

bvaughn:
We can chat in Gitter.

source/CellMeasurer/CellMeasurer.js
```diff
@@ -0,0 +1,222 @@
...
+  getColumnWidth ({ index }) {
+    if (this._cachedColumnWidths[index]) {
+      return this._cachedColumnWidths[index]
+    }
+
+    const { rowCount } = this.props
+
+    let maxWidth = 0
+
+    for (var rowIndex = 0; rowIndex < rowCount; rowIndex++) {
```
kof: maybe also let?

source/CellMeasurer/CellMeasurer.js
```diff
@@ -0,0 +1,222 @@
...
+
+    const rendered = cellRenderer({
+      columnIndex,
+      rowIndex
+    })
+
+    this._div.innerHTML = ReactDOMServer.renderToString(rendered)
+
+    const height = this._div.clientHeight
+    const width = this._div.clientWidth
```
kof: not sure how expensive those getters are, as you need only height or width, maybe it would be better that you also query dom for only what you need

source/CellMeasurer/CellMeasurer.js
```diff
@@ -0,0 +1,222 @@
...
+      this._div.style.zIndex = -1
+
+      this._updateDivDimensions(this.props)
+
+      this._containerNode = this._getContainerNode(this.props)
+      this._containerNode.appendChild(this._div)
+    }
+  }
+
+  _unrenderAndDestroy () {
```
kof: unrender sounds weird) destroy is also not really what happens here ...

removeContainer, unmountContainerNode would be better.

source/CellMeasurer/CellMeasurer.js
```diff
@@ -0,0 +1,222 @@
...
+    rowIndex
+  }) {
+    const { cellRenderer } = this.props
+
+    const rendered = cellRenderer({
+      columnIndex,
+      rowIndex
+    })
+
+    this._div.innerHTML = ReactDOMServer.renderToString(rendered)
```
kof: I have a strong feeling that this is very bad:
- renderToString should be slower than rendering  the element via ReactDom.render()
- renderToString should def. consume memory a lot, because it will create DOM nodes all over the time, no DOM reuse at all here. Basically this means that everything is rendered twice. 

Why not to .render() the element, measure it, then unmount it and then use the same element in the visible area? Like I did here https://gist.github.com/kof/eefccc63442bf6748165311e953fc68a#file-autorowheight-js-L13

MoOx:
Hi, I want to use this project and saw this PR.
Do you think I can directly start with this (v7) to avoid a upgrade in a few days? :)

bvaughn:
@MoOx: Beginning with this branch and then switching over to v7 once it's released would probably be the least effort for you, if you're able to do that. I hope to get v7 out sometime this weekend if nothing comes up so you wouldn't have too long to wait.

source/CellMeasurer/CellMeasurer.js
bvaughn: I also have big reservations about this line. I actually tried a couple of render approaches like you mention- but React threw an error about functions (eg. `getRowHeight`) having side effects that trigger rendering.

I think the main problem is that I want the width/height getter functions to be synchronous, not callback based like yours. Because (a) I don't want to force everyone to use the `InfiniteLoader` for the callback and (b) I think the callback approach you're using might cause a lot of `forceUpdates`.

I'd love to improve this bit though. Let's talk more about ideas?

source/CellMeasurer/CellMeasurer.js
bvaughn: Sure. This was initially doing something else (with my first attempt, which used `.render` instead of `.renderToString`). I can rename it though.

source/CellMeasurer/CellMeasurer.js
bvaughn: I _think_ you pay the cost once, regardless of whether you call one or both of them.

source/CellMeasurer/CellMeasurer.js
bvaughn: Sure. `for (var` habit dies hard.

source/CellMeasurer/CellMeasurer.js
kof: How do you know this? My assumption that layout calc happens on such things. And also they are likely to be slow anyways.

source/CellMeasurer/CellMeasurer.js
kof: Yes, because the measuring should have happened before. Not at the moment it needs to render. React does a good job with this warnings.

I am not sure measuring heights is 100% guaranteed in sync, need to do some research to be sure CSS is applied in sync with dom insertion. 

Also there are other edge cases like fonts and images which will result in a different heights asynchronously. For e.g. a font-family could trigger an async font load and when its loaded text would be rerendered .... same for images... 

I am not sure this should be supported, however we could do that if we have an async callback and  user could decide how long to wait.

kof: If async def. can't be used, we should try to find a way to make React.render to render sync.

source/CellMeasurer/CellMeasurer.js
kof: https://gist.github.com/paulirish/5d52fb081b3570c81e3a

source/CellMeasurer/CellMeasurer.js
kof: Found the line which is responsible to trigger actual rendering:

https://github.com/facebook/react/blob/master/src/renderers/dom/client/ReactMount.js#L495

Using this "private" methods one can make render sync.

But I think it is better to make the whole rendering in the visible area at the next tick, so that we don't need hacks at all, we could render in the hidden container using pure react flow and then use the state to trigger children updates.

source/CellMeasurer/CellMeasurer.js
bvaughn: My understanding of what's "expensive" about these calls is that they force the browser to do a sync layout. Once that's been done once (for querying `clientHeight` for example) why would the browser need to do it again when `clientWidth` was queried? Such an implementation would seem massively inefficient.

kof: I am not sure browser can efficiently cache those calculations and won't go check everything again when you ask for clientWidth ....

source/CellMeasurer/CellMeasurer.js
bvaughn: My plan was just to document the limitations of this measurer (eg. it can't be used to measure async-loaded images, fonts that haven't yet been loaded, etc.). I think the async measurer with callback approach you initially used poses a lot more challenges for the library itself though in terms of avoiding unnecessary renders and callbacks.

Let's say we have an RV list that is large enough to display 20 rows at a time. If each one measures async and we have no guarantee what order their callbacks will be invoked in, then we can't know the offset of row 20 until rows 1-19 have completed measuring. Not so awful when we're only talking about 20 rows but if we're talking about hundreds I think it would be very inefficient to have to run through the entire list of sizes and offsets and constantly tweak them as new values came in.

The above problem is way more complex when it comes to `Grid` since we would need to measure every column in a given row to know the rows height (and vice versa to know a given column's width).

source/CellMeasurer/CellMeasurer.js
kof: We can benchmark it.

kof: so bad jsperf.com is offline, need to start local instance ...

bvaughn: It's fine. I don't mind changing it to only query for the one we need just to be on the safe side.

source/CellMeasurer/CellMeasurer.js
bvaughn: I'm leaning towards a naive initial implementation at least that doesn't try to solve every use case but also doesn't add tons of complexity to RV. Then we can iterate on it in a way that's backwards compatible.

For instance, as of version 7 RV will use named parameters for everything. That means that while a `rowHeight` getter only takes a named `index` parameter now, we could update (in a backwards compatible way) it to also accept an optional `callback` parameter if we wanted to add async support. I _think_ async measuring support would require a substantial re-thinking about the layout logic though and I don't think I have the bandwidth for it now.

MoOx:
Ha, this repo can't be used in npm since source are ignored :(

MoOx:
Trying this as a postinstall hook `cd node_modules && git clone https://github.com/bvaughn/react-virtualized.git && cd react-virtualized && git checkout version-7 && npm i && npm run build` :D

bvaughn:
> Ha, this repo can't be used in npm since source are ignored :(

Ah, you mean version-7 can't be used without building. Yup, that makes sense. I'll publish an RC at some point soon but I have a couple of more TODO items (listed in the description) first.

bvaughn:
FYI @MoOx~ I've just published version `7.0.0-rc.1` to NPM. I'd like to do more performance testing before a final release candidate, but this should let you start using the changes I've done so far if you'd like.

bvaughn:
Performance tests aren't showing anything alarming- but it's gotten too late to feel comfortable with a release. Look for 7.0 in the morning. :)

PR #209, opened 2016-05-02
CellSizeAndPositionManager utility for JIT calculating cell sizes and positions

bvaughn:


edulan:
This is awesome! :clap: 

I'm looking forward to see it in action. Is it supposed to work with Collection component too?

bvaughn:
Thanks :)

Unfortunately this particular utility won't help with `Collection` since there's no guarantee that the order of its data correlates with the positioning. I've optimized `Collection` as best I could think of currently via the [`SectionManager`](https://github.com/bvaughn/react-virtualized/blob/master/source/Collection/SectionManager.js) component. But I'd welcome ideas for improving it. :D

PR #216, opened 2016-05-05
Fix initial scroll on Collection component

edulan:
Setting initial scroll on a Collection component is not working properly. Cells are rendered in the right place but scrolling container remains the same. This PR fix that.

I've tried to duplicate the same logic found in Grid component, in which initial scrolls works as expected.

### DIFF
--- a/source/Collection/CollectionView.js
+++ b/source/Collection/CollectionView.js
@@ -135,6 +135,10 @@ export default class CollectionView extends Component {
 
     this._scrollbarSize = getScrollbarSize()
 
+    if (scrollLeft >= 0 || scrollTop >= 0) {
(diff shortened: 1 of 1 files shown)


bvaughn:
Thanks for the contribution :)

My preference would be to move this check to an else-if following `if (scrollToCell >= 0)` (a little further down in the same method) just to avoid potentially updating state twice.

bvaughn:
I'll just make my own suggested change post-merge.

bvaughn:
Just release 6.3.2 with this fix. Gave you a shoutout in the changelog and git tag.

Thanks again!

edulan:
Thanks a lot @bvaughn :+1:

Issue #116, opened 2016-02-21
Always on scrollbar causing offset issues with ColumnSizer and ScrollSync


ryanseddon:
I created two gifs to show what's happening on your demo pages, these seem like a similar issue to #81 in that they both need to take into account if a scrollbar is present and make adjustments based on that. 

I don't have capacity to dig into this and do a PR until later in the week so if you're busy I'll take a look when I can.

`<ColumnSizer />`:

![scrollbar-sizer](https://cloud.githubusercontent.com/assets/143402/13206245/33863d36-d94e-11e5-9ec2-b59a773a3f1a.gif)

`<ScrollSync />`:

![scrollsync-offset](https://cloud.githubusercontent.com/assets/143402/13206249/44bba7c6-d94e-11e5-9829-b2eb52a38712.gif)

ryanseddon:
These gifs were recorded in Chrome 48.0.2564.109, I don't see the column sizer issue in Firefox but I see the scrollsync across the board.

bvaughn:
Thanks for the detailed report @ryanseddon. This seems related to another issue I've filed but haven't had the time to address yet, #101.

bvaughn:
My wedding was this weekend and I have family visiting so I probably won't get to this for another few days. If you beat me to it, more power to you. Else I'll take a look when I can. :)

ryanseddon:
Oh! Congrats! Step away from the computer and go enjoy your wedding festivities.

bvaughn:
Thanks thanks. ;) Was just taking a brief rest and catching my breath. Cheers!

Guria:
Just for reference screenshot of ColumnSizer demo from windows:
![](http://i.imgur.com/lHINPVV.png)

Don't sure if it supposed to look like that.

bvaughn:
If scrollbars are turned to always-on, then... yeah it would have to look like that.

Guria:
@bvaughn scrollbar always on is default behaviour for all browser under windows and linux. There is no easy way to get mac os like behaviour.

bvaughn:
I don't think that changes my response. I'm still not sure (in the above screenshot) what you would expect to see differently, given always-on scrollbars (regardless of whether the scrollbars are on because of the operating system, or the fact that the user is using a mouse instead of trackpad, or something else).

Guria:
@bvaughn my expectation that with setting like on next screenshot is enough room to scrollbars not appear at all. But they always appears except on mobile or mac os:
[...]

PR #234, opened 2016-05-13
fix #222 scrollbar covered cell content

### LINKED ISSUE #222: scrollbar covered cell content
When columns width === grid width and there's a vertical scrollbar, the scrollbar will cover the last cell of every row.
Similarly, when rows Height === grid height and theres a horizontal scrollbar, it will cover the last cell of every column.

here's an example: http://plnkr.co/edit/lV9gaaS34xoybhfiWipJ

a
b

Seems the problem was from Grid.js  [here](https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/Grid.js#L436) and [here](https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/Grid.js#L440)

May be change it to something like this can solve the problem?

```js
    if (totalColumnsWidth + (totalRowsHeight > height ? this._scrollbarSize : 0) <= width) {
      gridStyle.overflowX = 'hidden'
    }

    if (totalRowsHeight + (totalColumnsWidth > width ? this._scrollbarSize : 0) <= height) {
      gridStyle.overflowY = 'hidden'
    }
```

anjianshi:


source/Grid/Grid.js
```diff
@@ -204,6 +204,8 @@ export default class Grid extends Component {
 
     // See defaultCellRangeRenderer() for more information on the usage of this cache
     this._cellCache = {}
+
+    this._scrollbarSize = getScrollbarSize()
```
bvaughn: I don't believe this is a good idea. I think it could cause problems with server-side rendering.

bvaughn: Elaborating a bit- I think the `inDom` check for `dom-helpers` will prevent scrollbar size from ever being set in that case. Because at the time the component is _created_ it won't be in the DOM.

Hmm... maybe we could address this concern by:
* Adding a fallback to `render` to handle a potentially undefined size (eg `const scrollbarSize = this._scrollbarSize || 0`)
* Adding a check in `componentDidMount` for an undefined `this._scrollbarSize` and re-initializing if necessary

bvaughn:
Friendly bump @anjianshi. If you're willing to make the suggested PR changes for this~ great! If not let me know and I'll close the PR and tackle the issue when I get some time. :)

anjianshi:
Oh, sorry! these days I was very busy, tomorrow I will try to improve the PR.

bvaughn:
No problem! I understand being busy. :)

source/Grid/Grid.js
anjianshi: @bvaughn  I think this is not good. Because the first render will not be correct. May be we can skip set the overflow style in first render? I think this will not affect #116.

source/Grid/Grid.test.js
```diff
@@ -120,43 +120,49 @@ describe('Grid', () => {
   describe('shows and hides scrollbars based on rendered content', () => {
     const scrollbarSize = getScrollbarSize()
 
+    function renderTwice (markup) {
```
bvaughn: Why is this helper necessary? (Why is rendering twice required by these tests?)

anjianshi: Because in first render, `overflowX` or `overflowY` always won't be set
(because `this._scrollbarSize` was `undefined`)

anjianshi: Now in first render, overflowX and overflowY won't be set, so it can produce right result, whatever server-side rending or client-side rending.
And in follow up renders, because `this._scrollbarSize` was set, `overflowX` and `overflowY` will be set normally, to prevent #116 's problem

bvaughn: This is why I previous suggested what I did- initializing it in `componentWillMount` and then...
> * Adding a check in `componentDidMount` for an undefined `this._scrollbarSize` and re-initializing if necessary

This would handle the server-side rendering concern.

We should not require external code (test code or production code) to call `forceUpdate` in order for the correct layout behavior. `Grid` should update itself if it detects that condition.

anjianshi: but without this helper, the test won't pass.

bvaughn: Re-read my suggestion above? :)

anjianshi: In first render, overflowX or Y always won't be set, so there's noting to test.
The purpose of call forceUpdate() is testing the second render's result.
Or there's other way to do this?

Or we should `adding a fallback to render to handle a potentially undefined size (eg const scrollbarSize = this._scrollbarSize || 0)` as you say?
But this will cause the first render got wrong result about overflowX / overflowY.

sorry, my english is not very well. :)

bvaughn: It's not acceptable for react-virtualized users to have to call `forceUpdate` to get correct layout behavior. The server-side rendering edge-case we're talking about can be detected and handled by `Grid` automatically by checking for `this._scrollbarSize == null`. At that point your tests- (and more important, users using `Grid`- won't have to call `forceUpdate` themselves.

Does this make better sense?

bvaughn: What I originally suggested was to...
* Measure scrollbar size in [`componentWillMount`](https://facebook.github.io/react/docs/component-specs.html#mounting-componentwillmount) so that client-side content will only be measured once, right before rendering the first time.
* Check for an undefined scroll bar size in [`componentDidMount`](https://facebook.github.io/react/docs/component-specs.html#mounting-componentdidmount) so that server-side rendered content will be re-measured for scrollbar size at the first opportunity. In this second case, if scrollbar isn't measured until `componentDidMount`, then the `Grid` should `forceUpdate` (or `setState`) itself to ensure that it renders one more time with the adjusted scrollbar size.

bvaughn: I have a little time today while flying to just pick this up. I'll base it on what you've done and make the changes I've suggested. I'll give you credit for the idea/suggestion in release notes. :)

anjianshi: Understand. Happy to see these improve, and thanks for your excellent work!

PR #238, opened 2016-05-16
Only run codecov when CI environment variable is set

cesarandreu:
`npm run test` makes `codecov` run. That should probably only happen when it's in a CI environment. This PR changes it to be a noop when `CI` environment variable is missing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "lint": "standard",
     "prebuild": "npm run lint",
     "postpublish": "npm run deploy",
-    "posttest": "codecov",
+    "posttest": "[ -z \"$CI\" ] || codecov",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
     "lint": "standard",
     "prebuild": "npm run lint",
     "postpublish": "npm run deploy",
-    "posttest": "codecov",
+    "posttest": "[ -z \"$CI\" ] || codecov",
```
Guria: platform specific command, would fail on windows and alternate shells

bvaughn: True. Unfortunately there are no official collaborators besides myself so it doesn't pose any problems currently.

PR #252, opened 2016-05-23
Add center option for scrollToAlignment

edulan:
Adds "center" option to `scrollToAlignment` for being able to scroll to a cell and display it centered on the container.

This is specially useful when scrolling horizontally and want to show a given cell in the middle of the screen (in our use case is the today column in a gantt chart)

### DIFF
--- a/source/Grid/Grid.test.js
+++ b/source/Grid/Grid.test.js
@@ -314,7 +314,7 @@ describe('Grid', () => {
       // 100 rows * 20 item height = 2,000 total item height
       // We first scroll past the specified cell and then back.
       // The minimum amount of scrolling then should leave the specified cell in the middle (just scrolled into view).
-      // Since alignment is set to "middle" we should scroll past this point until the cell is aligned center.
+      // Since alignment is set to "center" we should scroll past this point until the cell is aligned center.
       expect(grid.state.scrollLeft).toEqual(1075)
       expect(grid.state.scrollTop).toEqual(920)
     })
--- a/source/Grid/Grid.test.js
+++ b/source/Grid/Grid.test.js
@@ -299,6 +299,25 @@ describe('Grid', () => {
       expect(grid.state.scrollLeft).toEqual(1050)
       expect(grid.state.scrollTop).toEqual(900)
     })
+
+    it('should scroll to the correct position for :scrollToAlignment "center"', () => {
+      render(getMarkup({
+        scrollToColumn: 99,
+        scrollToRow: 99
+      }))
+      const grid = render(getMarkup({
+        scrollToAlignment: 'center',
+        scrollToColumn: 24,
+        scrollToRow: 49
+      }))
+      // 100 columns * 50 item width = 5,000 total item width
+      // 100 rows * 20 item height = 2,000 total item height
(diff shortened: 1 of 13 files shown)


source/Grid/Grid.test.js
```diff
@@ -299,6 +299,25 @@ describe('Grid', () => {
...
+      const grid = render(getMarkup({
+        scrollToAlignment: 'center',
+        scrollToColumn: 24,
+        scrollToRow: 49
+      }))
+      // 100 columns * 50 item width = 5,000 total item width
+      // 100 rows * 20 item height = 2,000 total item height
+      // We first scroll past the specified cell and then back.
+      // The minimum amount of scrolling then should leave the specified cell in the middle (just scrolled into view).
+      // Since alignment is set to "middle" we should scroll past this point until the cell is aligned center.
```
bvaughn: "center" (not "middle")

source/VirtualScroll/VirtualScroll.test.js
```diff
@@ -115,6 +115,19 @@ describe('VirtualScroll', () => {
...
+
+    it('should scroll to the correct position for :scrollToAlignment "center"', () => {
+      render(getMarkup({
+        scrollToIndex: 99
+      }))
+      const rendered = findDOMNode(render(getMarkup({
+        scrollToAlignment: 'center',
+        scrollToIndex: 49
+      })))
+      // 100 items * 10 item height = 1,000 total item height; 10 items can be visible at a time.
```
bvaughn: This comment's a bit misleading I guess, since 11 items are actually visible now (the first and last item are only partially visible).

source/utils/getUpdatedOffsetForIndex.js
```diff
@@ -19,12 +19,15 @@ export default function getUpdatedOffsetForIndex ({
 }) {
   const maxOffset = cellOffset
   const minOffset = maxOffset - containerSize + cellSize
+  const midOffset = maxOffset - (containerSize + cellSize) / 2
```
bvaughn: No need to calculate this here. Just return it inline down below, in the `'center'` case.

min and max offsets are only calculated here because they're used by multiple cases below.

bvaughn:
Nice update! Thanks for the PR. Left some small comments.

Docs will also need to be updated (but I can do that if you'd prefer to just focus on the code.)

source/VirtualScroll/VirtualScroll.test.js
edulan: Yep, I'll update that

edulan:
Thanks @bvaughn. I'll take care of the docs too :wink:

source/utils/getUpdatedOffsetForIndex.js
edulan: :ok_hand:

andresgutgon:
📯 👍

bvaughn:
Thanks for putting together this PR :)

bvaughn:
Version 7.1.2 just released with this feature.

edulan:
Thank you so much @bvaughn for being so fast :rocket:!

The work you're doing with react-virtualized is totally awesome.

PR #259, opened 2016-05-27
Added new method to Grid/FlexTable/VirtualScroll for forced measuring of all cells

### LINKED ISSUE #258: Render entire container scroll height for VirtualScroll
Hi Brian,

I'm currently using `VirtualScroll` where I am manually changing the `scrollTop` when the user clicks on a button that should scroll them to a particular row. The reason I'm not using the provided `scrollToIndex` prop is because the scroll that I need to do requires animation.

However this becomes an issue if there are tons of rows being rendered and I need to animate the `scrollTop` to something not yet visible (ie. the last row). Because `VirtualScroll` only renders what is visible to the user, The container has a inaccurate scroll height representation; and only when the user begins to scroll further down, does the scroll height begin to increase and increase.

If I already know the expected scroll height of the container, would there be a way to add a prop for me to pass it down, so that the container would be the correct scroll height, but still render only the visible rows, the rest of the scroll area would just be whitespace.

In the `Grid` component that `VirtualScroll` uses, there is a `totalRowsHeight` variable to set the height of the of the div with the class `Grid__innerScrollContainer`.

`const totalRowsHeight = this._rowSizeAndPositionManager.getTotalSize()`;

Ideally, it would be great if we could so something like:

`const totalRowsHeight = this.props.totalRowsHeight || this._rowSizeAndPositionManager.getTotalSize();`

PS. I love this library, it's slick as hell!

Best,

- Johnny

bvaughn:
Resolves #258

### DIFF
--- a/docs/FlexTable.md
+++ b/docs/FlexTable.md
@@ -10,6 +10,7 @@ This component expects explicit width, height, and padding parameters.
 | children | [FlexColumn](FlexColumn.md) | ✓ | One or more FlexColumns describing the data displayed in this table |
 | className | String |  | Optional custom CSS class name to attach to root `FlexTable` element. |
 | disableHeader | Boolean |  | Do not render the table header (only the rows) |
+| estimatedRowSize | Number |  | Used to estimate the total height of a `FlexTable` before all of its rows have actually been measured. The estimated total height is adjusted as rows are rendered. |
 | headerClassName | String |  | CSS class to apply to all column headers |
 | headerHeight | Number | ✓ | Fixed height of header row |
 | headerStyle | Object |  | Optional custom inline style to attach to table header columns. |
@@ -35,6 +36,12 @@ This component expects explicit width, height, and padding parameters.
 | width | Number | ✓ | Width of the table |
 
 ### Public Methods
+
+##### measureAllRows
+Pre-measure all rows in a `FlexTable`.
+
+Typically rows are only measured as needed and estimated heights are used for cells that have not yet been measured.
+This method ensures that the next call to getTotalSize() returns an exact size (as opposed to just an estimated one).
 
 ##### recomputeRowHeights
 Recompute row heights and offsets.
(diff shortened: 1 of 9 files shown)


mwilliamson:
Would it make sense to allow this behaviour to be set as a property? Otherwise, I think you have to remember to call the method again each time the rows (or rather, their heights) change.

I might be demonstrating my ignorance of React here, but is there a way to call the method before the first render?

bvaughn:
The behavior _could_ be set as a property I suppose, but then I'd have to add another check to `componentWillUpdate` which I'd kind of like to avoid. I also kind of like the symmetry of having `recomputeGridSize` and `measureAllCells` as methods.

I don't think the first render matters too much in this case but please correct me if I'm wrong. The first render will automatically measure enough rows/columns to fill the viewport. Calling `measureAllCells` after that render (eg in a `componentDidMount` type method or similar) won't force a re-render so that's okay. And the only time the total height comes into play is when a user begins scrolling (or when you begin adjusting/animating a scroll offset) at which point- a render would be triggered and the newly-measured columns/rows would have an affect.

mwilliamson:
The total height makes a difference if the user is trying to scroll to the end, for instance by hitting Ctrl-End, or by hitting the bottom of the scrollbar background (in UI toolkits where hitting the scrollbar background takes you to that scroll position, rather than scrolling you a fixed distance in that direction).

bvaughn:
Yeah, definitely. But what I was saying was that it doesn't make a different _until_ the user scrolls or until you programmatically scroll. So whether the measure-all functionality has happened when the initial rendering occurs doesn't matter.

mwilliamson:
I'm not quite sure I follow. If the height is underestimated, then when the user tries to scroll to the bottom (by hitting Ctrl-End or whatever), they won't actually scroll to the bottom, which can be a jarring user experience.

bvaughn:
That's why I added the new `measureAllCells` method. Use it if you want the height to be _accurate_ instead of _estimated_. You expressed concern that you can't actually call that method until the component has been mounted/rendered once. I was just pointing out that I don't think this _actually_ matters because- until a user has interacted with the list in some way, the total size isn't used for any thing.

The use-case of using a keyboard shortcut to jump to the end is one I haven't considered to be honest.

PR #261, opened 2016-05-28
Handle larger scroll sizes than browser supports natively

### LINKED ISSUE #219: Handle larger scroll sizes than browser supports natively
Browsers have maximum scroll offsets, after which the user is unable to scroll further _or_ programmatically increase the offset.

To see this, go to the [`VirtualScroll` test harness](https://bvaughn.github.io/react-virtualized/?component=VirtualScroll), set the number of rows to 1,000,000, and scroll to the bottom. Somewhere in the middle (eg row 559,239) the browser will stop scrolling.

We should be able to work around this limitation by doing something like:
* Choosing a size ceiling (eg 10,000,000 pixels) for the scrolling container
* Calculate an offset based on the size of the ceiling divided by the total cell size (assuming this is greater)
* Shift rendered cells slightly to account for this discrepancy

This way we could cram more cells in a given area than would be normal without it being noticeable to an end user.

This will probably be a bit tricky to implement since it impacts several areas (finding the visible cell range for a given offset, rendering a range of cells, etc.) and everything will need to account for the shift.

bvaughn:
Resolves #219


tbranyen:
Can you throw up a demo when you get a chance? I wanna test out the scrolling behavior, and btw nice work!

bvaughn:
Of course. It's here: https://bvaughn.github.io/react-virtualized/

Just bump the number of rows in the VirtualScroll up to 10,000,000 or something.

PR #267, opened 2016-06-05
How to expand a row on click (repro case in playground)

anjdreas:
Hi,

I'm doing a log viewer and want to expand a row upon mouse click, to show more details in that entry. I tried a naive approach by updating `selectedRowIndex` state of parent of `FlexTable`, and the row does change height, but the neighbor rows do not update so they overlap each other. 

If I initialize the state so `selectedRowIndex: 2`, it does correctly adjust all rows on initial render, but when clicking subsequent rows it fails to render properly.

Any idea how I can accomplish this?

Repro case: https://github.com/anjdreas/react-virtualized/blob/expand-bug-repro/playground/expand.js

Summary of implementation:
```
            ReactVirtualized.FlexTable,
            {
              onRowClick: ({ index}) => {
                self.setState({ selectedRowIndex: index })
              },
              rowHeight: ({index}) => index === self.state.selectedRowIndex ? 100 : 20,
```

Screenshot:
image

### DIFF
--- a/playground/expand.html
+++ b/playground/expand.html
@@ -1,0 +1,45 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="initial-scale=1, maximum-scale=1">
+  <title>foo</title>
+  <link rel="stylesheet" type="text/css" href="../styles.css" />
+  <style type="text/css">
+    body, html, #mount {
+      width: 100%;
+      height: 100%;
+      margin: 0;
+      padding: 0;
+      font-family: -apple-system, BlinkMacSystemFont,  "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell",  "Fira Sans", "Droid Sans", "Helvetica Neue",  sans-serif;
+      font-size: 12px;
+      overflow: hidden
+    }
+    * {
+      box-sizing: border-box;
+    }
+    .container {
+      width: 100%;
+      height: 100%;
+    }
+    .expand {
+      border: 1px solid #eee;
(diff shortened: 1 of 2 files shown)


bvaughn:
Hey @anjdreas,

The key problem is that you're not telling `FlexTable` that the heights of some of its rows has changed. For all it knows from the outside, nothing has changed. It still has the same `rowCount` as before so it does not remeasure anything.

The easiest way to do that would be like this:
```diff
diff --git a/expand.js b/expand.js
index e812453..71577de 100644
--- a/expand.js
+++ b/expand.js
@@ -46,6 +46,8 @@ var App = React.createClass({
           }
 
           mostRecentWidth = autoSizerParams.width
+          
+          var flexTable
 
           return React.createElement(
             ReactVirtualized.FlexTable,
@@ -54,7 +56,9 @@ var App = React.createClass({
               onRowClick: ({ index}) => {
                 console.debug('rowClick: ' + index)
                 self.setState({ selectedRowIndex: index })
+                flexTable.recomputeRowHeights()
               },
+              ref: (ref) => { flexTable = ref },
               rowHeight: ({index}) => index === self.state.selectedRowIndex ? 100 : 20,
               rowGetter: ({index}) => chatHistory[index],
               rowCount: chatHistory.length,
```

By the way, next time it would probably be easier to just share your repro as a Plunker, forked from this: https://plnkr.co/edit/6syKo8cx3RfoO96hXFT1

anjdreas:
Thanks, makes sense. Plunker looks great by the way.

anjdreas:
Sorry about creating a PR, that was not intended.

bvaughn:
No worries. :)

Issue #70, opened 2016-01-29
scroller on window.onscroll


harrybabu:
Currently the scroller is visible on the container div and the scrolling logc works for the particular div, is it possible to emulate the logic on window.onscroll.

bvaughn:
I don't think I understand. Even if your virtual list/table was 100% width
and height of the viewport, you wouldn't be able to use window scroll
because the scrollable region of items would be inside of the list/table.
(That's how the virtualization is done.)

Maybe you could explain in a different way?

harrybabu:
I want to use the entire window as a scroll container instead of just a single div.

bvaughn:
But what would be the difference if the div was the full width and height
of the window?

bvaughn:
Here's a visual I created of how react-virtualized works:
![react-virtualized-2](https://cloud.githubusercontent.com/assets/29597/12694042/4baad0f2-c6d4-11e5-87ab-76e83c5d3398.png)

In this example, a list is 300 pixels tall. It contains 10 items, each with a row height of 60 pixels. This means that at any time only 5 or 6 rows will be visible. react-virtualized only renders those 5-6 rows though, but it creates a scrollable container that's 600 pixels tall (10 rows x 60 pixels each).

If the list was the entire height of the browser window, it would still work the same. (There would be a scrollable container large enough to contain all of the list items- but only enough to fill the visible viewport would be rendered.)

So basically... there would be no difference. :)

Going to close this issue. Feel free to ask a follow-up question if you have any.

harrybabu:
Yeah i understood that

Assume that you have a pagelayout of header, footer, left sidebar and content (react-vrtualized list) is in the middle. Currenlty the scrollbar is coming only for the content wrapper div.

My usecase is, i want the scrollbar to be at the window level. Meaning the content wrapper should be window instead of wrapper div. 

Any help appreciated :+1:

harrybabu:
Something like this

https://github.com/seatgeek/react-infinite#user-content-using-the-window-to-scroll-usewindowasscrollcontainer-mode

bvaughn:
I don't understand why you'd want that, but maybe once you have it working you can show me and I would? :) Anyway... you could accomplish that by doing the following:

1. Use a `fixed` position for your content so that it won't scroll with the `window`.
1. Tell the `window` how much scrollable height it has so a scrollbar appears. (This is probably as simple as `heightHeight + footerHeight + numRows * rowHeight`.)
[...]

PR #270, opened 2016-06-09
WindowScroller HOC

minheq:
Hi,
I have been on development of this HOC for a while but here is its purpose:

1. Use Grid/VirtualScroll that is not contained by fixed height, but stretches across the window (innerScrollContainer height)
2. Use _window_ to handle the scroll event instead of Grid's onScroll handler

Ultimately, it is to achieve something like Facebook newsfeed, e-commerce site's product lists or Pinterest feed. And since company I work for also wanted this, I went on to develop this HOC. Also #70 wants this

**I added WindowScroller HOC example into demo application**, so you can see how it works in the example given.

Let me know if this could be further improved, and how. Thanks!

**Note:**
For this component to work as intended, _removeHeightContainer_ is required to be passed in to Grid/VirtualScroll component, wrapped by the HOC.

P/S: This is my first ever pull request, so hopefully you can provide some general feedback so I can improve my contributions further in future :)


source/Grid/Grid.js
```diff
@@ -122,6 +122,12 @@ export default class Grid extends Component {
     overscanRowCount: PropTypes.number.isRequired,
 
     /**
+     * Removes fixed height from the scrollingContainer so that the total height
+     * of rows can stretch the window. Intended for use with WindowScroller
+     */
+    removeHeightContainer: PropTypes.bool,
```
bvaughn: Can we change the name of this? Maybe something like `autoHeight`?

source/Grid/Grid.js
```diff
@@ -551,7 +558,8 @@ export default class Grid extends Component {
               height: totalRowsHeight,
               maxWidth: totalColumnsWidth,
               maxHeight: totalRowsHeight,
-              pointerEvents: isScrolling ? 'none' : 'auto'
+              pointerEvents: isScrolling ? 'none' : 'auto',
+              position: 'relative'
```
bvaughn: Static styles (like `position`) should go in `styles.css` to be more consistent with other styling stuff.

source/Grid/Grid.test.js
```diff
@@ -546,7 +548,7 @@ describe('Grid', () => {
     })
   })
 
-  describe('styles and classeNames', () => {
+  describe('styles and classNames', () => {
```
bvaughn: hah 👍

source/Grid/Grid.test.js
```diff
@@ -991,4 +993,34 @@ describe('Grid', () => {
...
+      const props = {
+        removeHeightContainer: true
+      }
+      const grid = render(getMarkup(props))
+      const rendered = findDOMNode(render(getMarkup(props)))
+
+      expect(rendered.querySelector('.Grid__innerScrollContainer').style.height).toEqual('2000px') // 100 rows * 20px rowHeight
+      expect(grid._rowSizeAndPositionManager.getTotalSize()).toEqual(2000)
+    })
+  })
```
bvaughn: Nice tests :)

source/VirtualScroll/VirtualScroll.js
```diff
@@ -141,6 +142,7 @@ export default class VirtualScroll extends Component {
           stopIndex: rowStopIndex
         })}
         overscanRowCount={overscanRowCount}
+        removeHeightContainer={removeHeightContainer}
```
bvaughn: Could you add this (or rather, the renamed `autoHeight`) pass-thru property to `FlexTable` as well?

source/WindowScroller/WindowScroller.js
```diff
@@ -0,0 +1,76 @@
...
+    window.addEventListener('scroll', this._onScroll, false)
+  }
+
+  shouldComponentUpdate (nextProps, nextState) {
+    return shallowCompare(this, nextProps, nextState)
+  }
+
+  componentWillUnmount () {
+    window.removeEventListener('scroll', this._onScroll, false)
+  }
```
bvaughn: tiny nit: Alpha-sort the React lifecycle methods? :)

source/WindowScroller/WindowScroller.js
```diff
@@ -0,0 +1,76 @@
...
+          height,
+          scrollTop
+        })}
+      </div>
+    )
+  }
+
+  _onScroll (event) {
+    this._setNextState({
+      scrollTop: Math.max(0, window.scrollY - Math.max(0, this._positionFromTop))
```
bvaughn: Could remove the `Math.max(0, this._positionFromTop)` check from this method (which will trigger many times) and put it into `componentDidMount` where the value gets set initially (so it only runs once).

source/WindowScroller/WindowScroller.js
```diff
@@ -0,0 +1,76 @@
...
+      height: 0
+    }
+
+    this._onScroll = this._onScroll.bind(this)
+  }
+
+  componentDidMount () {
+    this._positionFromTop = ReactDOM.findDOMNode(this).getBoundingClientRect().top
+    this.setState({ height: window.innerHeight })
+    window.addEventListener('scroll', this._onScroll, false)
```
bvaughn: We also need to watch for changes to `window.innerHeight`. If you start with a very small height and then it grows larger (because a user resizes) then it currently leaves the last few rows of the `Grid` empty.

source/WindowScroller/WindowScroller.test.js
```diff
@@ -0,0 +1,68 @@
...
+  it('inherits the window height and passes it to child component', () => {
+    window.innerHeight = 500
+    const component = render(getMarkup())
+
+    expect(component.state.height).toEqual(window.innerHeight)
+    expect(component.state.height).toEqual(500)
+
+    const rendered = findDOMNode(component)
+    expect(rendered.textContent).toContain('height:500')
+  })
```
bvaughn: Might be nice to add a test for updated height after resize events :)

bvaughn:
Thanks for this PR! Looks great for the most part. Left some minor comments in the code. :)

One thing I did notice was that it breaks if you swap `VirtualScroll` for `Grid` because horizontal scrolls cause `scrollTop` to be reset. I suppose we could just document this as a limitation of the component. But it would be nice if it worked for vanilla `Grid` too. Any thoughts?

A bit more concerning, the component demo doesn't seem to work on IE10. I haven't spent any time looking into why this is yet (because I have to leave for work now) but IE10 support is kind of something I've committed to. Any chance you could take a look? :)

minheq:
Thanks for the comments! Glad that you are OK with it  😬

With that, I will sort all the comments and update the branch. And regarding your concerns and thoughts, I will address them too. Getting to it now 💯  Hopeful that this could get merged to the codebase 💃  WIP!

Guria:
@minheq :fire: great hoc, I think it would be helpful

source/Grid/Grid.js
```diff
@@ -745,12 +752,14 @@ export default class Grid extends Component {
...
-    const { height, width } = this.props
+    const { autoHeight, height, width } = this.props
     const scrollbarSize = this._scrollbarSize
     const totalRowsHeight = this._rowSizeAndPositionManager.getTotalSize()
     const totalColumnsWidth = this._columnSizeAndPositionManager.getTotalSize()
     const scrollLeft = Math.min(Math.max(0, totalColumnsWidth - width + scrollbarSize), event.target.scrollLeft)
-    const scrollTop = Math.min(Math.max(0, totalRowsHeight - height + scrollbarSize), event.target.scrollTop)
+    const scrollTop = this.props.scrollTop && autoHeight
+      ? this.props.scrollTop
+      : Math.min(Math.max(0, totalRowsHeight - height + scrollbarSize), event.target.scrollTop)
```
minheq: This part is responsible for supporting vanilla Grid, such that when we scroll horizontally it still persists the scrollTop from the prop.

Not sure if this is the best solution though, but it was the most straightforward one I found. And also I am not really fond of putting extra ternary expressions here, it might affect the performance a little bit, but we could always put this into component's limitation like u suggested instead. This is just a proposed solution to your comment :)

bvaughn: Gotcha. I think I'd probably rather document the limitation (for now) than do this. Thanks for proposing the option though. :)

source/demo/Application.js
```diff
@@ -35,7 +36,7 @@ class Application extends Component {
     const matches = window.location.search.match('component=(.+)')
     const activeComponent = matches && (COMPONENTS.includes(matches[1]) || HIGH_ORDER_COMPONENTS.includes(matches[1]))
       ? matches[1]
-      : 'VirtualScroll'
+      : 'WindowScroller'
```
bvaughn: Whoops, meant to mention this before. We shouldn't change the demo default landing page, I don't think. No big deal though, I can always change this post-merge if needed. :)

source/WindowScroller/WindowScroller.example.js
```diff
@@ -0,0 +1,76 @@
...
+            {({ height, scrollTop }) => (
+              <VirtualScroll
+                autoHeight
+                className={styles.VirtualScroll}
+                height={height}
+                rowCount={list.size}
+                rowHeight={30}
+                rowRenderer={this._rowRenderer}
+                scrollTop={scrollTop}
+                width={460}
```
bvaughn: This will cause it to overflow on narrow browsers (eg iPhone). Suggestion: Use `AutoSizer` here too?

```diff
diff --git a/source/WindowScroller/WindowScroller.example.js b/source/WindowScroller/WindowScroller.example.js
index 1c0a8ca..8901b18 100644
--- a/source/WindowScroller/WindowScroller.example.js
+++ b/source/WindowScroller/WindowScroller.example.js
@@ -2,6 +2,7 @@
 import Immutable from 'immutable'
 import React, { Component, PropTypes } from 'react'
 import { ContentBox, ContentBoxHeader, ContentBoxParagraph } from '../demo/ContentBox'
+import AutoSizer from '../AutoSizer'
 import WindowScroller from './WindowScroller'
 import VirtualScroll from '../VirtualScroll'
 import shallowCompare from 'react-addons-shallow-compare'
@@ -38,16 +39,20 @@ export default class AutoSizerExample extends Component {
         <div className={styles.WindowScrollerWrapper}>
           <WindowScroller>
             {({ height, scrollTop }) => (
-              <VirtualScroll
-                autoHeight
-                className={styles.VirtualScroll}
-                height={height}
-                rowCount={list.size}
-                rowHeight={30}
-                rowRenderer={this._rowRenderer}
-                scrollTop={scrollTop}
-                width={460}
-              />
+              <AutoSizer disableHeight>
+                {({ width }) => (
+                  <VirtualScroll
+                    autoHeight
+                    className={styles.VirtualScroll}
+                    height={height}
+                    rowCount={list.size}
+                    rowHeight={30}
+                    rowRenderer={this._rowRenderer}
+                    scrollTop={scrollTop}
+                    width={width}
+                  />
+                )}
+              </AutoSizer>
             )}
           </WindowScroller>
         </div>
```

bvaughn:
Let me know when I should dive back into this PR again by the way. I see you've made a lot of changes but it's not clear if it's back to me or still waiting on input from you. No hurry of course! Just want to make sure I'm quick in getting back to you. :)

source/demo/Application.js
minheq: Ahhh, yes! my bad, I'll change it back :)

minheq:
Most of the changes are done, only IE10 support is left to work on so right after I fix it, I'll let you know and mention you in a comment :) Thanks for taking your time in this!

minheq:
@bvaughn Hey, sorry, only today found time to complete the changes. Added couple more changes, and this component now also works for IE10+ :D

It is ready for your review, let me know what you think :) And thanks for your time! 😬

minheq:
@bvaughn Hey, sorry, only today found time to complete the changes. Added couple more changes, and this component now also works for IE10+ :D

It is ready for your review, let me know what you think :) And thanks for your time! 😬

bvaughn:
Actually it works for IE9+ ;) so even better!

Thanks for the changes. I'm reviewing them this morning. I don't mind writing the documentation myself if necessary. You've already done a lot of work on this feature. I think this is easily the largest PR I've received! :)

bvaughn:
Looks good to me. Merging!

bvaughn:
7.5.0 has just been released with this new addition! Gave you a big shout-out in the [CHANGELOG](https://github.com/bvaughn/react-virtualized/blob/master/CHANGELOG.md#750) and [tag release notes.](https://github.com/bvaughn/react-virtualized/releases/tag/7.5.0)

Thanks again for your hard work on this feature! You're now the 2nd highest contributor for this project ;)

minheq:
Wow! Thanks a lot for your help @bvaughn along the way, really glad I could contribute to this really awesome project :)

PR #272, opened 2016-06-12
Added rowClassName to Grid

nicholasrq:
To bring more flexibility at styling Grid i've added `rowClassName` setting to Grid and VirtualScroll components. It will help to assign dynamic custom class names to `Grid__cell` elements.

`rowClassName` can be a string or a function represented by following signature: `({rowIndex: number, columnIndex: number}): PropTypes.string`.

Not sure about using `_wrapSizeGetter` to wrap `rowClassName` function but don't want to copy that method to stick to DRY paradigm.


nicholasrq:
Added some extra tests to be sure that no previous functionality is broken

bvaughn:
Hey @nicholasrq,

Thanks for the PR! This seems like a reasonable addition. Couple of suggestions though:
* The property in `Grid` should be named `cellClassName` as you say.
* A `rowClassName` property should be added to both `VirtualScroll` and `FlexTable` that adapts the arguments (similar to how `FlexTable`'s `_getRowHeight` method does now).
* Looks like some test-only changes were made to `VirtualScroll.example.js` that should be reverted

nicholasrq:
Hey! Thank's for your comment.

As i can see `FlexTable` already has an option named `rowClassName`. It's not very clear where i should add classes in case where `FlexTable`'s rows wrapped with `Grid`'s `Grid__cell` element. Should i continue to add classes to `Grid__cell` itself or i need to add class names to `FlexTable__rowColumn` instead? Or both to of them?

nicholasrq:
Hey @bvaughn! Me again.

Accordingly to your comment i've made changes to both `Grid` and `VirtualScroll`.

For `Grid` property now named `cellClassName`.
For `VirtualScroll` i've made an adapter to convert `rowIndex` to `index` and fixed docs accordingly to these changes.

I've also unified `_wrapSizeGetter` by adding `_wrapPropertyGetter` method which is doing similar things but is more abstract. Both `_wrapSizeGetter` and `_wrapCellClassNameGetter` are now using `_wrapPropertyGetter`.

Waiting for your answer to my previous comment about `FlexTable`.

bvaughn:
Hm, at first I thought we could just make use of the new `Grid` property and just remove the wrapper `FlexTable__row` element. This would also probably simplify custom CSS styles for IE9 support. Unfortunately I was not considering the a11y or mouse click/over/out methods attached to that wrapper element. Need to think on that a bit.

The general pattern I've been trying to follow for `Grid` and its 2 HOCs- `FlexTable` and `VirtualScroll` is:
* If a property exists in one, it exists in all 3 (although sometimes with a slight name variation- eg "cell" becomes "row").
* If a custom class property can be set, a similar, sibling custom style property should also be supported.

I'd like to continue following that pattern with this property although I'm a bit stumped initially with regard to what it should be called in `FlexTable`.

source/Grid/Grid.js
```diff
@@ -687,10 +692,18 @@ export default class Grid extends Component {
...
+
   _wrapSizeGetter (size) {
-    return typeof size === 'function'
-      ? size
-      : () => size
+    return this._wrapPropertyGetter(size)
+  }
+
+  _wrapCellClassNameGetter (className) {
+    return this._wrapPropertyGetter(className)
```
bvaughn: Couple of minor nits:
* Alpha-sort private methods and properties. (This just helps keep things tidy and easier to read/maintain.)
* Add a `cellStyle` property. (I try to mirror class-name and style object properties b'c there are a reasonable number of people who prefer inline styles over css classes.)

I like the generic `_wrapPropertyGetter` change. 👍

docs/VirtualScroll.md
```diff
@@ -16,6 +16,7 @@ This component renders a virtualized list of elements with either fixed or dynam
 | rowHeight | Number or Function | ✓ | Either a fixed row height (number) or a function that returns the height of a row given its index: `({ index: number }): number` |
 | rowRenderer | Function | ✓ | Responsbile for rendering a row given an index. Signature should look like `({ index: number, isScrolling: boolean }): React.PropTypes.node` |
 | rowCount | Number | ✓ | Number of rows in list. |
+| rowClassName | String or Function |  | Optional custom CSS class name to attach to `Grid__cell` element. If function given then signature should be look like: ({index: number}): PropTypes.string |
```
bvaughn: nit: Alpha-sort doc properties too if you would. Makes them easier for people to scan visually when they're looking for a property.

source/Grid/Grid.test.js
```diff
@@ -564,6 +566,59 @@ describe('Grid', () => {
...
+        cellClassName: ({rowIndex, columnIndex}) => {
+          return `col-${rowIndex}-${columnIndex}`
+        }
+      })))
+      const rows = Array.from(rendered.querySelectorAll('.Grid__cell')).map(row => {
+        return row.className.split(' ')[1]
+      })
+      expect(rows.length).toEqual(4)
+      expect(rows).toEqual(['col-0-0', 'col-0-1', 'col-1-0', 'col-1-1'])
+    })
```
bvaughn: 👍 for the tests

source/Grid/defaultCellRangeRenderer.js
```diff
@@ -16,6 +17,7 @@ export default function defaultCellRangeRenderer ({
   rowSizeAndPositionManager,
   rowStartIndex,
   rowStopIndex,
+  cellClassName,
   scrollLeft,
   scrollTop,
   verticalOffsetAdjustment
```
bvaughn: nit: Alpha-sort these props

source/VirtualScroll/VirtualScroll.js
```diff
@@ -68,6 +68,9 @@ export default class VirtualScroll extends Component {
     /** Number of rows in list. */
     rowCount: PropTypes.number.isRequired,
 
+    /** Optional custom CSS class for individual rows */
+    rowClassName: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
```
bvaughn: Ditto for alpha-sort

source/VirtualScroll/VirtualScroll.js
```diff
@@ -151,6 +156,7 @@ export default class VirtualScroll extends Component {
         autoHeight={autoHeight}
         rowHeight={rowHeight}
         rowCount={rowCount}
+        cellClassName={rowClassName}
         scrollToAlignment={scrollToAlignment}
         scrollToRow={scrollToIndex}
         scrollTop={scrollTop}
```
bvaughn: Ditto for alpha-sort of class props and JSX `<Grid>` props.

nicholasrq:
@bvaughn ok then. I'll try to experiment with `FlexTable` and will notify you if i'll have any results. Also if you'll have any thoughts about that problem i will happy to read your feedback.

bvaughn:
Maybe `rowWrapperClassName` and `rowWrapperStyle` for `FlexTable` ? and just keep the `rowClassName` (and `rowStyle`) for `VirtualScroll` ?

bvaughn:
There don't seem to be any obvious solutions.

bvaughn:
PS. How are you intending to use this new property? Just curious.

nicholasrq:
@bvaughn `rowWrapperClassName ` seems more legit for me. I thought about this kind of solution.

About leaving `cellClassName` for `VirtualScroll` will not be very obvious because there are no "cells" actually, just rows although they named `Grid__cell`.

p.s. in my case there is grouped list of items, so i flatten them into simple array, but some items still "group headers" so i need to style them differently. also "groups" and "items" should have different `z-index` which is not possible with absolutely positioned parent and i can achieve it by styling children of `Grid__cell`, so i have to add styles directly to `Grid__cell` but now there is no difference between "group header" and "item" in my case. with custom classes on `Grid_cell` now i can separate headers and items to apply different styles to them.

bvaughn:
Let's do `rowWrapperClassName` and `rowWrapperStyle` for `FlexTable` and stick with the current name for `VirtualScroll` then.

nicholasrq:
@bvaughn i've added `cellStyle` and `rowStyle` to `Grid` and `VirtualScroll`. Also fixed docs and methods ordering accordingly to your comments.

Changes to `FlexTable` will appear in separate commit in case of complexity of this component.

nicholasrq:
@bvaughn looks like i've made it. Waiting for your feedback.

bvaughn:
Looks good. Merging.

nicholasrq:
Yay!

bvaughn:
I'll be pushing out a release within the next few.

Thanks for your contribution. :)

nicholasrq:
Thanks! Waiting for release =)

bvaughn:
7.6.0 is out. You got a shout-out in the release notes. Thanks again!

nicholasrq:
Looks great =) Going to come back if any interesting thoughts will come.

Cheers!

PR #273, opened 2016-06-13
Fix FlexTable doc table format

VincentBel:
Fix some of the wrong positioned rows in FlexTable doc:
image


bvaughn:
Thanks kindly :)

Issue #279, opened 2016-06-15
Preventing VirtualScroll contents from going under scroll bar


trevorr:
I've noticed that because the width of the `Grid__innerScrollContainer` is set to the same value as the `Grid`, the contents of a `VirtualScroll` go under the scroll bar. This is causing an issue for me with focusable rows because the right focus border is hidden by the scroll bar:

<img width="431" alt="screen shot 2016-06-15 at 1 58 07 pm" src="https://cloud.githubusercontent.com/assets/928766/16095533/a17d9c9c-330a-11e6-9648-0c84fb3e6fca.png">

Components like `react-virtualized-select` sidestep this issue by using background color to indicate focus, rather than an outline, though the rows still extend behind the scroll bar:

<img width="415" alt="screen shot 2016-06-15 at 1 42 28 pm" src="https://cloud.githubusercontent.com/assets/928766/16095551/b2f155ea-330a-11e6-97ed-ac55e5dff201.png">

For visual consistency with other (Bootstrap-based) components, I'd like to keep its focus styling, rather than just using background color. I thought about using right padding inside the rows, but that has several problems, including that the scroll bar may not always be present and, when it is, its width will vary across browsers. I think the ideal way to fix this is to specify `width="100%"` (rather than the fixed `Grid` width) for the `Grid__innerScrollContainer` and the contained rows/cells. Hacking these changes in at runtime yields the desired appearance:

<img width="430" alt="screen shot 2016-06-15 at 1 59 30 pm" src="https://cloud.githubusercontent.com/assets/928766/16095564/c000714e-330a-11e6-82df-99aae8cd6fc5.png">

Now I realize that because `VirtualScroll` is based on `Grid`, which supports bidirectional scrolling and multiple columns, achieving this may be tricky. Any thoughts on how it might be done? I can attempt to submit a PR if we can come up with an acceptable approach.

bvaughn:
This is somewhat complicated, as you guessed, because of the inner `Grid` and its constraints. See related issues #222 (the most recent and so the most relevant), #154, #131, #116, #101, #80.

I'd welcome suggestions or ideas from you on how to improve this aspect of RV, so long as we could fix without a regression of one of the issues mentioned above.

trevorr:
[...]

PR #280, opened 2016-06-16
Allow Grid column/row sizes to be specified as a percentage and use a…

trevorr:
… 100% column for VirtualScroll

### DIFF
--- a/source/Grid/Grid.example.css
+++ b/source/Grid/Grid.example.css
@@ -33,6 +33,7 @@
   flex-direction: column;
   justify-content: center;
   padding: 0 .5em;
+  overflow: hidden;
 }
 .cell {
   border-right: 1px solid #e0e0e0;
@@ -64,4 +65,11 @@
   justify-content: center;
   font-size: 1em;
   color: #bdbdbd;
+}
+
+.checkboxLabel {
+  margin-left: .5rem;
+}
+.checkboxLabel:first-of-type {
+  margin-left: 0;
 }
--- a/source/Grid/Grid.example.js
+++ b/source/Grid/Grid.example.js
@@ -27,11 +27,13 @@ export default class GridExample extends Component {
       rowCount: 1000,
       scrollToColumn: undefined,
       scrollToRow: undefined,
-      useDynamicRowHeight: false
+      useDynamicRowHeight: false,
+      usePercentColumnWidth: false
     }
 
     this._cellRenderer = this._cellRenderer.bind(this)
     this._getColumnWidth = this._getColumnWidth.bind(this)
+    this._getPercentColumnWidth = this._getPercentColumnWidth.bind(this)
     this._getRowClassName = this._getRowClassName.bind(this)
     this._getRowHeight = this._getRowHeight.bind(this)
     this._noContentRenderer = this._noContentRenderer.bind(this)
@@ -53,7 +55,8 @@ export default class GridExample extends Component {
       rowCount,
       scrollToColumn,
       scrollToRow,
-      useDynamicRowHeight
+      useDynamicRowHeight,
+      usePercentColumnWidth
     } = this.state
 
     return (
@@ -66,7 +69,8 @@ export default class GridExample extends Component {
 
         <ContentBoxParagraph>
           Renders tabular data with virtualization along the vertical and horizontal axes.
-          Row heights and column widths must be calculated ahead of time and specified as a fixed size or returned by a getter function.
+          Row heights and column widths must be calculated ahead of time and specified as
+          a fixed size/percentage or returned by a getter function.
         </ContentBoxParagraph>
 
         <ContentBoxParagraph>
@@ -79,6 +83,17 @@ export default class GridExample extends Component {
               onChange={event => this._updateUseDynamicRowHeights(event.target.checked)}
             />
             Use dynamic row height?
+          </label>
+
+          <label className={styles.checkboxLabel}>
+            <input
+              aria-label='Use percentage column widths?'
+              checked={usePercentColumnWidth}
+              className={styles.checkbox}
+              type='checkbox'
+              onChange={event => this._updateUsePercentColumnWidth(event.target.checked)}
+            />
+            Use percentage column widths?
           </label>
         </ContentBoxParagraph>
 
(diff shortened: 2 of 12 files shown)


bvaughn:
Hey @trevorr,

Thanks for the contribution. I appreciate the time you've spent. I'm going to need to give this some careful thought and consideration (which I probably won't have the time to do until the weekend). I'm initially very wary of introducing another forking path here WRT cell measurements because it increases my maintenance effort in the future. But I'll give it some thought. :)

trevorr:
I understand. An alternative approach I considered was adding a "single column mode" to `Grid` that `VirtualScroll` enables. It would set the container and cell widths to 100%. It would have been less code and in some ways easier to understand, but I chose the other approach because "single column mode" felt like a special case hack that might not play well with other features. It also begged the question of why not "single row mode" too, which probably then would have netted the same amount of code, but with lots of `if` statements instead of polymorphism.

bvaughn:
Okay. Back to thinking about this a little.

I feel like percentage widths for columns would be better as a HOC - something similar to [`ColumnSizer`](https://github.com/bvaughn/react-virtualized/blob/master/source/ColumnSizer/ColumnSizer.js). I'm really reluctant to fork the cell measuring logic and add more complexity to `Grid`.

What do you think about a HOC approach? We could even use the `dom-helpers/util/scrollbarSize` util to measure the scrollbar size and take that into account when it comes to calculating the column widths.

bvaughn:
Actually I wonder if we could even roll this functionality _into_ `ColumnSizer` ?

bvaughn:
Actually- thinking more about the specific problem you're trying to solve here- I wonder if we couldn't help things along a bit more by passing scrollbar width as a new name parameter to `cellRenderer`.

bvaughn:
Although if all you want is support for 100% column width in `VirtualScroll` ... you might not even need the `ColumnSizer` work. You could just rely on the new `horizontalScrollBarSize` property passed to `cellRenderer` and size your cell accordingly?

bvaughn:
Just ignore the above ramblings. ;) I'm going to take a quick shot at this myself this morning. I have a few minutes before work. Will check back in shortly.

bvaughn:
It's been a few days since I've been able to think about this. I had another idea.

A couple of small adjustments to `Grid`...
```diff
diff --git a/source/Grid/Grid.js b/source/Grid/Grid.js
index 655d51f..c631311 100644
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -449,6 +449,7 @@ export default class Grid extends Component {
 
   render () {
     const {
+      autoHeight,
       cellClassName,
       cellRenderer,
       cellRangeRenderer,
@@ -459,7 +460,6 @@ export default class Grid extends Component {
       noContentRenderer,
       overscanColumnCount,
       overscanRowCount,
-      autoHeight,
       rowCount,
       style,
       tabIndex,
@@ -580,7 +580,7 @@ export default class Grid extends Component {
           <div
             className='Grid__innerScrollContainer'
             style={{
-              width: totalColumnsWidth,
+              width: columnCount === 1 ? 'auto' : totalColumnsWidth,
               height: totalRowsHeight,
               maxWidth: totalColumnsWidth,
               maxHeight: totalRowsHeight,
diff --git a/source/Grid/defaultCellRangeRenderer.js b/source/Grid/defaultCellRangeRenderer.js
index 713ceb1..c7d9918 100644
--- a/source/Grid/defaultCellRangeRenderer.js
+++ b/source/Grid/defaultCellRangeRenderer.js
@@ -68,11 +68,11 @@ export default function defaultCellRangeRenderer ({
           key={key}
           className={cn('Grid__cell', className)}
           style={{
-            ...cellStyleObject,
             height: rowDatum.size,
             left: columnDatum.offset + horizontalOffsetAdjustment,
             top: rowDatum.offset + verticalOffsetAdjustment,
-            width: columnDatum.size
+            width: columnDatum.size,
+            ...cellStyleObject
           }}
         >
           {renderedCell}
```
...would allow this to be accomplished by specifying a `rowClassName` or `rowStyle`:

### Style
```js
<VirtualScroll
  rowStyle={{ width: '100%' }}
  {...otherProps}
/>
```

### Class name

```js
<VirtualScroll
  rowClassName={styles.rowWrapper}
  {...otherProps}
/>
```
```css
.rowWrapper {
  width: 100% !important;
}
```

This approach feels a _little_ hacky as well but it avoids the 150+ line layout and measurement fork.

Thoughts? Reactions?

trevorr:
I think that seems fine. Just a few questions:

- Does this still allow a `Grid` that is not part of a `VirtualScroll` to have a single column wider than the `Grid` and scroll it?
- I've been wondering, is setting `maxWidth` and `maxHeight` to be the same as `width` and `height` (respectively) on the container not redundant? Similarly, what is the effect of setting `width: auto` and `maxWidth` equal to the `Grid` width? Could that result in hiding part of the row again in some cases, or does row `width: 100%` prevent that and make the container `maxWidth` irrelevant?
- Would you have `VirtualScroll` work this way by default, or would any user wanting this behavior have to override the row width as shown? I would think the proposed behavior would always be desired, but maybe there's a compatibility concern?

bvaughn:
> Does this still allow a Grid that is not part of a VirtualScroll to have a single column wider than the Grid and scroll it?

Assuming the cell/column width is greater than the `Grid` width then yeah, it should still be scrollable.

> I've been wondering, is setting maxWidth and maxHeight to be the same as width and height (respectively) on the container not redundant? Similarly, what is the effect of setting width: auto and maxWidth equal to the Grid width? Could that result in hiding part of the row again in some cases, or does row width: 100% prevent that and make the container maxWidth irrelevant?

I don't recall if I specifically need the max width/height or if I'm just setting them out of caution. Setting the width/height is important though and 100% wouldn't work. The size of the inner container is what makes the `Grid` scrollable, and then based on scroll position- RV decides which columns to render. If we used 100% or _just_ max width/height, the `Grid` wouldn't be scrollable b'c it would only contain the initially rendered cells. (Or if it was scrollable it would only be scrollable in so much as the overscan rows/columns made it scrollable.)

> Would you have VirtualScroll work this way by default, or would any user wanting this behavior have to override the row width as shown? I would think the proposed behavior would always be desired, but maybe there's a compatibility concern?

I think I'd probably make `VirtualScroll` and `FlexTable` default to this behavior.

trevorr:
Sounds good to me then. I'll close this PR and we can go with your change for fixing #279.

PR #283, opened 2016-06-20
Avoid setting scrollTop when autoHeight prop present

### LINKED ISSUE #281: WindowScroller degraded performance due to assigned scrollTop
Related: #270 cc @minheq 

I've been trying to incorporate the newly-introduced WindowScroller component into our application, and while I was successful in doing so, I noted that scrolling the page did not feel very smooth.

After some profiling, I found that this set of lines is triggered very frequently, causing many forced repaints:

https://github.com/bvaughn/react-virtualized/blob/19bcb10/source/Grid/Grid.js#L298-L318

Since with WindowScroller, the element being scrolled is the window, not the rendered grid, `this.refs.scrollingContainer` will always have a `scrollTop` of `0`, and so this assignment happens much more frequently than I imagine is expected. Removing this set of lines seems to improve the experience a bit, but obviously this would regress non-WindowScroller usage. Generally speaking, most of the `Grid.js` file appears to assume that scrolls will occur within the rendered container (understandable since WindowScroller was a later addition).

It's not immediately obvious to me how this could be solved, since Grid can't know whether the received `scrollTop` is from a WindowScroller or a ScrollSync (where the latter would want to assign the `scrollTop` of the `scrollingContainer`). Perhaps an extra prop like `windowContext` to avoid the `scrollTop` being assigned, or one to more generally indicate that overflow is visible, not scrolled?

Code: https://github.com/Automattic/wp-calypso/blob/update/cpt-infinite-list-2/client/my-sites/post-type-list/index.jsx
Demo: https://calypso.live/types/post?branch=update/cpt-infinite-list-2 (assumes active WordPress.com user session with a site containing enough posts for scrolling; unfortunately a decent amount of effort involved if not available to you already)

aduth:
Fixes #281 

This pull request seeks to avoid setting `scrollTop` in the `Grid` component `componentDidUpdate` when `autoHeight` is true. This is especially relevant in the context of `WindowScroller`, where because the height of the `Grid` expands automatically, `scrollTop` will always be zero.

__This pull request is in-progress. Specifically in consideration of:__

- The tests are failing. `scrollTop` on the rendered element always seems to be zero in tests, even when we'd expect it to be assigned. Perhaps I'm not hooking in at the right point in lifecycle? Or something with the test environment setup.
- Includes a commit for experimental `pointer-events` assigning on the `body` element when scrolling under `WindowScroller`. There's a bit of duplication between here and `Grid` that might warrant some clean-up, and totally lacks tests. Okay with dropping this commit altogether, but figured it was worth exploring.

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -292,7 +292,7 @@ export default class Grid extends Component {
    * 1) New scroll-to-cell props have been set
    */
   componentDidUpdate (prevProps, prevState) {
-    const { height, scrollToAlignment, scrollToColumn, scrollToRow, width } = this.props
+    const { autoHeight, height, scrollToAlignment, scrollToColumn, scrollToRow, width } = this.props
     const { scrollLeft, scrollPositionChangeReason, scrollTop } = this.state
 
     // Make sure requested changes to :scrollLeft or :scrollTop get applied.
@@ -300,7 +300,7 @@ export default class Grid extends Component {
     // And to discard any pending async changes to the scroll position that may have happened in the meantime (e.g. on a separate scrolling thread).
     // So we only set these when we require an adjustment of the scroll position.
     // See issue #2 for more information.
-    if (scrollPositionChangeReason === SCROLL_POSITION_CHANGE_REASONS.REQUESTED) {
+    if (!autoHeight && scrollPositionChangeReason === SCROLL_POSITION_CHANGE_REASONS.REQUESTED) {
       if (
         scrollLeft >= 0 &&
         scrollLeft !== prevState.scrollLeft &&
--- a/source/Grid/Grid.test.js
+++ b/source/Grid/Grid.test.js
@@ -478,15 +478,26 @@ describe('Grid', () => {
   describe(':scrollLeft and :scrollTop properties', () => {
     it('should render correctly when an initial :scrollLeft and :scrollTop properties are specified', () => {
       let columnStartIndex, columnStopIndex, rowStartIndex, rowStopIndex
-      render(getMarkup({
+      render(getMarkup())
+      const rendered = findDOMNode(render(getMarkup({
         onSectionRendered: params => ({ columnStartIndex, columnStopIndex, rowStartIndex, rowStopIndex } = params),
         scrollLeft: 250,
         scrollTop: 100
-      }))
+      })))
+      expect(rendered.scrollTop).toEqual(100)
       expect(rowStartIndex).toEqual(5)
       expect(rowStopIndex).toEqual(9)
       expect(columnStartIndex).toEqual(5)
       expect(columnStopIndex).toEqual(8)
+    })
+
+    it('should not set :scrollLeft or :scrollTop when using :autoHeight', () => {
+      const rendered = findDOMNode(render(getMarkup({
+        autoHeight: true,
+        scrollLeft: 250,
+        scrollTop: 100
+      })))
+      expect(rendered.scrollTop).toEqual(0)
     })
 
     it('should render correctly when :scrollLeft and :scrollTop properties are updated', () => {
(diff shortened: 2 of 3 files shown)


source/WindowScroller/WindowScroller.js
```diff
@@ -95,6 +102,18 @@ export default class WindowScroller extends Component {
     onResize({ height })
   }
 
+  _enablePointerEventsAfterDelay () {
+    if (this._disablePointerEventsTimeoutId) {
+      clearTimeout(this._disablePointerEventsTimeoutId)
+    }
+
+    this._disablePointerEventsTimeoutId = setTimeout(() => {
```
bvaughn: In the spirit of "everything counts" I've been trying to avoid creating throw-away functions for things that happen frequently (like scrolling). For instance I recently updated this same bit of code in `Grid` to reuse a single function:
https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/Grid.js#L604 n

I think we should do the same here.

source/WindowScroller/WindowScroller.js
```diff
@@ -107,6 +126,12 @@ export default class WindowScroller extends Component {
 
     this._setNextState({ scrollTop })
 
+    if (this._originalBodyPointerEvents === null) {
+      this._originalBodyPointerEvents = document.body.style.pointerEvents
+      document.body.style.pointerEvents = 'none'
```
bvaughn: I wonder if this has any weird layout/repainting side effects (changing body style)?

source/Grid/Grid.js
```diff
@@ -292,15 +292,15 @@ export default class Grid extends Component {
...
+    const { autoHeight, height, scrollToAlignment, scrollToColumn, scrollToRow, width } = this.props
     const { scrollLeft, scrollPositionChangeReason, scrollTop } = this.state
 
     // Make sure requested changes to :scrollLeft or :scrollTop get applied.
     // Assigning to scrollLeft/scrollTop tells the browser to interrupt any running scroll animations,
     // And to discard any pending async changes to the scroll position that may have happened in the meantime (e.g. on a separate scrolling thread).
     // So we only set these when we require an adjustment of the scroll position.
     // See issue #2 for more information.
-    if (scrollPositionChangeReason === SCROLL_POSITION_CHANGE_REASONS.REQUESTED) {
+    if (!autoHeight && scrollPositionChangeReason === SCROLL_POSITION_CHANGE_REASONS.REQUESTED) {
```
bvaughn: On the one hand- this feature request and PR makes total sense. On the other hand it still sits oddly with me. It feels like poor coupling between `Grid` and `WindowScroller`. I need to give this a little more thought.

bvaughn:
Thanks for contributing this PR @aduth. I don't feel great about the coupling between `Grid` and `WindowScroller` but I understand the important of avoiding forced layout/reflow so... I'll go ahead and make my own suggested PR changes and merge this.

source/Grid/Grid.test.js
```diff
@@ -478,17 +478,28 @@ describe('Grid', () => {
...
       let columnStartIndex, columnStopIndex, rowStartIndex, rowStopIndex
-      render(getMarkup({
+      render(getMarkup())
+      const rendered = findDOMNode(render(getMarkup({
         onSectionRendered: params => ({ columnStartIndex, columnStopIndex, rowStartIndex, rowStopIndex } = params),
         scrollLeft: 250,
         scrollTop: 100
-      }))
+      })))
+      expect(rendered.scrollTop).toEqual(100)
```
bvaughn: Hm. This test doesn't pass for me (or [CI](https://circleci.com/gh/bvaughn/react-virtualized/797)).

aduth:
Oh, awesome! Thanks @bvaughn for the quick work here and for the improvements in a1fa6c13ce9cf8a996f64020d794deef8d0f686a. Gonna take the new version for a spin in our app.

Issue #292, opened 2016-06-23
Webpack UMD config and `react-addons-shallow-compare`


ducky427:
Hi. 
I am using the UMD build of this library. It seems that the UMD webpack config file doesn't list `react-addons-shallow-compare` as an external package. So it is included in the final build file. I think this shouldn't be the case. But I don't know enough by webpack to suggest a fix. 

Any help would be appreciated. 

Thanks

bvaughn:
Hey @ducky427,

Interesting. I think I agree with you that it shouldn't be included in the UMD build.

The fix for this is pretty simple. I just need to add a new externals entry for `react-addons-shallow-compare` [here](https://github.com/bvaughn/react-virtualized/blob/master/webpack.config.umd.js#L16).

bvaughn:
Hm, not as straight forward as I hoped. Not sure how to make Webpack play nicely with something from `React.addons`. Tried a couple of things without any luck. Will take another look at this in the evening. In the meanwhile, help appreciated if anyone knows the answer. :)

ducky427:
I think I've figured it out. PR sent.

bvaughn:
7.11.1 release is available with this UMD fix. Thanks for your help!

ducky427:
Thanks a lot for the quick release!

👍

PR #293, opened 2016-06-23
Do not package shallow-compare for UMD

ducky427:
This addresses #292 hopefully.

Inspiration from https://github.com/callemall/material-ui/issues/3176


bvaughn:
Hm, trying to rebuild local UMD target with this change and then running one of my playground files errors with:
```
external "React.addons.shallowCompare":1
Uncaught TypeError: Cannot read property 'shallowCompare' of undefined
```

bvaughn:
Oh! No! I forgot I had reverted a change in my playground include. This does work! :)

ducky427:
cheers!

PR #295, opened 2016-06-24
Load React UMD Build with add-ons for playground

clauderic:
Since `react-addons-shallow-compare` is an external dependency, I believe you need to load the React UMD build with add-ons.


bvaughn:
Is your local checkout a little behind master by chance?

https://github.com/bvaughn/react-virtualized/blob/master/playground/helper.js#L38

clauderic:
Derp.

PR #298, opened 2016-06-27
New: first implementation of selection in FlexTable

Dogild:
Previously, when using the FlexTable developers needed to define their own system to handle the selection.
Now it is integrated in the component FlexTable.
This PR adds the following feature for the selection :
    - Handler onRowSelect(rowIndex) -> called when a row is selected
    - Handler onRowDeselect(rowIndex) -> called when a row is deselected
    - Handler onSelectionIndexesForProposedSelection(indexes) -> called to confirm or change the future selected rows, in this callback you could get the set 1,2,3 but returns 5,6 to select the rows 5 and 6. This is useful to cancel a selection, for instance to show a confirmation panel.
    - Added the propTypes allowsMultipleSelection, multiple selection will work only when using the mask key shift
    - Added the propTypes selectedRows to preselect the rows
    - The function rowWrapperStyle will give the boolean isSelected in the given object to specify if the row is selected or not.

Unit tests have been added in FlexTable.test.js


Dogild:
Let me know what was your idea with the HOC!

bvaughn:
Hey @Dogild,

Thanks for sharing this PR. :)

I still prefer for this functionality to not live within `FlexTable` though. I think `FlexTable` should be all about layout and stateful things like selection should be done in a HOC. (That way `FlexTable` doesn't get more complicated and harder to maintain and optimize.)

You could move the properties you added to `FlexTable`, more or less, to a new HOC that decorates a `FlexTable` _or_ a `VirtualScroll` and uses the existing row-click event to manage selected state.

I'm still on the fence a bit about whether this should be in core react-virtualized or if it's too specific a use-case.

Dogild:
Let me know what you think about this version :)

bvaughn:
Hey @Dogild,

This approach looks more consistent with the other components/HOCs. So that's 👍

I'm still undecided about whether it belongs in core or if it should be an external NPM project that depends on react-virtualized. I've found myself wishing I'd made some other components external libraries (eg `ScrollSync` and `CellMeasurer`) to keep react-virtualized core lighter and easier to maintain.

I'll leave a couple of thoughts on the PR. But I will need to give it a bit more consideration before I decide whether to merge or not. :)

source/SelectableGridComponent/SelectableGridComponent.js
```diff
@@ -0,0 +1,123 @@
...
+    * Optional callback when a row is deselected
+    * ({ index: number }): void
+    */
+    onRowDeselect: PropTypes.func,
+
+    /**
+     * Callback invoked when a user clicks on a table row.
+     * ({ index: number }): void
+     */
+    onRowClick: PropTypes.func,
```
bvaughn: I don't think you need to expose a redundant `onRowClick` callback here. Same goes for `rowWrapperStyle`. If users want to _also_ listen for clicks separately from select/deselect events- (which I'm not sure of)- or provide additional custom style- they could do so this:

```js
<SelectableGridComponent {...selectableGridComponentProps}>
  {({ rowWrapperStyle, onRowClick }) => (
    <FlexTable
      onRowClick={() => {
        usersOnClickHandler()
        onRowClick()
      }}
      rowWrapperStyle={{
        ...usersCustomStyle,
        ...rowWrapperStyle
      }}
      {...flexTableProps}
    >
      {...flexColumns}
    </FlexTable>
  )}
</SelectableGridComponent>
```

docs/SelectableGridComponent.md
```diff
@@ -0,0 +1,77 @@
...
+  { name: 'Brian Vaughn', description: 'Software engineer' }
+  // And so on...
+];
+
+// Render your table
+ReactDOM.render(
+  <SelectableGridComponent
+    allowsMultipleSelection
+    onRowSelect={(index) => console.error('Selected Index : ' + index.index)}
+    onRowDeselect={(index) => console.error('Deselected Index : ' + index.index)}
```
bvaughn: Should be...
```js
onRowSelect={({ index }) => console.log('Selected Index : ' + index)}
onRowDeselect={({ index }) => console.log('Deselected Index : ' + index)}
```

source/SelectableGridComponent/SelectableGridComponent.js
```diff
@@ -0,0 +1,123 @@
...
+
+    /** Optional custom CSS class for individual rows */
+    rowWrapperStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
```
bvaughn: Seems like users might want to react in more ways than just the `rowWrapperStyle`. Some users use external CSS (class names) instead of inline styles (eg `rowWrapperClassName`). Others may be using `rowClassName` or `rowStyle`. (The wrapper exists for kind of a specific purpose. I don't think it's the most commonly used one.)

Come to think of it, users might actually want to render different content (not just CSS changes).

It might be easier to satisfy all use-cases by allowing users to query `SelectableGridComponent` directly for its `selectedRows` value. Maybe something like...

```js
class MyComponent extends Component {
  render () {
    <SelectableGridComponent
      ref={(ref) => this._selectableGridComponent = ref}
      {...selectableGridComponentProps}
    >
      {({ onRowClick }) => (
        <FlexTable
          onRowClick={onRowClick}
          {...flexTableProps}
        >
          <FlexColumn
            {...flexColumnProps}
            rowClassName={({ index }) => {
              if (this._isRowSelected(index)) {
                return 'selectedRow'
              } else {
                return 'deselectedRow'
              }
            }}
            cellRenderer={
              ({ cellData, columnData, dataKey, rowData, rowIndex }) => {
                if (this._isRowSelected(index)) {
                  return <div>selected content</div>
                } else {
                  return <div>deselected content</div>
                }
              }
            }
          />
        </FlexTable>
      )}
    </SelectableGridComponent>
  }

  _isRowSelected (index) {
    return (
      this._selectableGridComponent &&
      this._selectableGridComponent.isRowSelected(index)
    )
  }
}
```

Dogild:
I have corrected the implementation following your instruction

bvaughn:
Sorry for kind of letting this hang. I went to Japan for a few weeks for work and it kind of slipped my mind. :/

I was on the fence about this PR, but I think this component would be better as a dependent component (eg [react-infinite-calendar](https://github.com/clauderic/react-infinite-calendar), [react-sortable-hoc](https://github.com/clauderic/react-sortable-hoc), [react-virtualized-checkbox](https://github.com/emilebres/react-virtualized-checkbox), [react-virtualized-select](https://github.com/bvaughn/react-virtualized-select)). That way if it's useful to others, they can benefit from it, but I won't have to manage the additional complexity. (I've been pretty strapped for time lately.)

I'd be happy to link to it from the main README as I do with the other components I mentioned above. I'd also be happy to answer any questions you have about getting a repo setup or getting the component published to NPM (if you haven't done it before).

I appreciate the contribution and the suggestion. Happy to help get it up and out for others to use. Let me know if I can do anything. :D

remon-georgy:
That feature is exactly what I need now. Thanks @bvaughn for the great work! Have you managed to package that component into a separate npm project?

Dogild:
Hi @remon-georgy, no I didn't put the component in a separate package. The company where I work doesn't have a open source policy yet...We will release that in the future, but I don't now when...

remon-georgy:
@Dogild fair enough. I'll still like to use your PR for inspiration :).

PR #303, opened 2016-07-02
Fix docs examples that use rowRenderer

stevenjob:
Small thing i found while reading docs.

Great Repo. Keep it up. :beer:


bvaughn:
Thank you! It's so difficult to keep docs updated properly. :)

PR #312, opened 2016-07-15
Forced inline style to overwrite calculated ones if present

dzannotti:
Currently inline styles passed to Grid overwrites passed styles via prop, i think the expected behaviour is to always respect the user's styles


dzannotti:
On second thoughts i added the same behaviour everywhere it would overwrite style

bvaughn:
Thanks for the PR @dzannotti! This actually helps with the discussion [here](https://github.com/bvaughn/react-virtualized/commit/e99279e753eaf4d63b57de0a8c45fffe9d24fbcd#commitcomment-18183136) so great timing. :)

PR #313, opened 2016-07-15
Fix cellRangeRenderer link in docs

sidd:


bvaughn:
Oh hey @sidd. Thanks for the fix! :)

PR #314, opened 2016-07-19
Use renderSubtreeIntoContainer in CellMeasurer

jquense:
:wave: howdy, This is a bit more intrusive then I was hoping but it does work :)

The original issue with rSIC is that it react doesn’t like when you do things like render new components while a `render()` method is firing: “render() should be pure”. To address this I moved the childrenToDisplay calculations in Grid to right _before_ each render which avoids the warning. This seems safe to me and my understanding of how the update cycle works.

I didn’t make any attempt to adjust Collection to handle this, I am assuming you probably don’t want to use this HoC in that context anyway yeah?

### DIFF
--- a/source/CellMeasurer/CellMeasurer.js
+++ b/source/CellMeasurer/CellMeasurer.js
@@ -1,7 +1,6 @@
 /** @flow */
 import React, { Component, PropTypes } from 'react'
 import ReactDOM from 'react-dom'
-import ReactDOMServer from 'react-dom/server'
 
 /**
  * Measures a Grid cell's contents by rendering them in a way that is not visible to the user.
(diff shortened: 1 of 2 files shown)


bvaughn:
Hey @jquense,

Thanks for the PR! Interesting. :)

I'm a tiny bit wary about introducing a dependency on one of the "unstable_*" methods, for reasons like this...
screen shot 2016-07-20 at 7 43 31 pm
...but I understand the `context` concern so it seems worth doing.

I've done a tiny bit of testing and perf profiling and nothing obvious seems to be negatively impacted, as would be expected. :)

Let me test it a bit more but I think it looks good.

jquense:
thanks! and for the unstable API,  I think that it's pretty safe for now, I've had some costs with the team about it and why the API may change (nothing on the horizon that I know of right now) the concept of a subtree is something that is integral to a few important/heavily used patterns in the react ecosystem system, so it'll be around in some form!

bvaughn:
Gotcha. Yeah, I assumed as much from Ben's tweet. I was just a bit wary about having to fork to support multiple, different APIs (should they change it). But we can cross that bridge if we get there. :)

bvaughn:
7.12.3 just released. Props to you in the CHANGELOG and on Twitter ;)

PR #324, opened 2016-07-25
Add property forceRecomputeRowHeights to VirtualScroll

Sediug:
Virtual Scroll doesn't update row heigh properly when using a function. 

This happens only when you are using a function in the rowHeight prop and you change something in that function. For example when the row height is dinamic and the function changes when the user do something. I tried to find out how to make this work properly and finally I found that the easiest way for me is to add a new property forceRecomputeRowHeights and, if that prop is true, call the method recomputeRowHeights on componentWillReceiveProps. I know isn't the best way but from now is my best idea, with that change it works properly updating the render when I want. I have something like this in my component:

```
componentWillMount () {
	this.forceRecomputeRowHeights = false
}

// When the function data changes I set that to true this.forceRecomputeRowHeights = true
// Then on render

render () {
   let forceRecomputeRowHeights = this.forceRecomputeRowHeights

   if  (forceRecomputeRowHeights)  {
         this.forceRecomputeRowHeights = false
   }

   <VirtualScroll
          ...
	  forceRecomputeRowHeights={forceRecomputeRowHeights}
    />
}
```

What's do you think? Thanks

### DIFF
--- a/source/VirtualScroll/VirtualScroll.js
+++ b/source/VirtualScroll/VirtualScroll.js
@@ -62,7 +62,14 @@ export default class VirtualScroll extends Component {
      */
     rowHeight: PropTypes.oneOfType([PropTypes.number, PropTypes.func]).isRequired,
 
-    /** Responsbile for rendering a row given an index; ({ index: number }): node */
+    /**
+     * Force to recompute row height on component receive new props. It may be usefull when you have a
+     * function in rowHeight property and you change internal values in that function and need's to rerender
+     * the component with the new height getter.
+     */
+    forceRecomputeRowHeights: PropTypes.bool,
+
+    /** Responsible for rendering a row given an index; ({ index: number }): node */
     rowRenderer: PropTypes.func.isRequired,
 
     /** Optional custom CSS class for individual rows */
@@ -100,6 +107,7 @@ export default class VirtualScroll extends Component {
     onScroll: () => null,
     overscanRowCount: 10,
     scrollToAlignment: 'auto',
+    forceRecomputeRowHeights: false,
     style: {}
   }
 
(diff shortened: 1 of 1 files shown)


bvaughn:
Hey @Sediug,

There are 2 relevant things to consider (although you may already be aware of both):
* [`react-virtualized` components are "pure"](https://github.com/bvaughn/react-virtualized#pure-components) in order to be as performant as possible. This means that unless properties change, nothing will re-render. This is why I expose a [`forceUpdateGrid` method](https://github.com/bvaughn/react-virtualized/blob/master/docs/VirtualScroll.md#forceupdategrid).
* If `rowHeight` is a function, changes to it are ignored. This is because people often define anonymous functions in `render`. Without this optimization, `react-virtualized` would be constantly throwing away its cache and recalculating. This is why I expose a [`recomputeRowHeights` method](https://github.com/bvaughn/react-virtualized/blob/master/docs/VirtualScroll.md#recomputerowheights-index-number).

Your change would work around the 2nd issue but I'm afraid the cost would be too great. Any time an external prop changed- anything- it would throw away the measurement cache. For dynamic height rows- especially ones measured by `CellMeasurer`- this would be a big performance hit.

In addition to that, all virtualized components share an informal interface. This means this method would need to be added to `Collection`, `FlexTable`, and `Grid` in addition to `VirtualScroll`. Particularly with `Grid` and `FlexTable` though this would be an even more expensive change. (`FlexTable` because sortable rows would clear the cache; `Grid` because both column and row caches would be cleared- being doubly expensive.)

I think the best solution here would just be for you to create a HOC that decorates a `VirtualScroll`. For example...

```js
class HighOrderComponent extends Component {
  componentWillReceiveProps (newProps) {
    if (this.props.rowHeight !== newProps.rowHeight) {
      this._virtualScroll.recomputeRowHeights(0)
    }
  }

  render () {
    return (
      <VirtualScroll
        {...this.props}
        ref={(ref) = this._virtualScroll = ref}
      />
    )
  }
}
```

bvaughn:
I appreciate the contribution! I really do. :)

But I'm extra wary of adding anything that might be accidentally used in a way that's costly to performance, so I'm going to pas on this change set. Hope you understand!

Sediug:
Yes, ofc. Thanks so much for your time :+1: I really apreciate your answer :). The HOC it's a nice idea, thanks you very much!

PR #327, opened 2016-07-27
Add resetRowMeasurements() and resetColumnMeasurements() to CellMeasurer

rhysd:
We need to reset measurements of specific row/column when the cell is resized.  Resize event occurs in children components and `CellMeasurer` cannot detect it.  So we need to notify `CellMeasurer` should reset measurements for the cell.  These APIs can be used for it.

I didn't know how to add test case because these are related to cache.  Cache is an internal state of component and I didn't know how to access inner state of the component.


bvaughn:
Hi @rhysd,

Sorry for how long it's taken me to review this. I try to respond to issues right away but PRs take a bit more thought, and I've been busy. :) I'll take a look now...

bvaughn:
I made a small rename for (subjective) consistency:
* `resetMeasurementForColumn`
* `resetMeasurementForRow`

And I added test coverage. :)

This release will go out momentarily as version `7.17.0`. Thank you for your contribution!

rhysd:
@bvaughn Thank you so much :smile:

PR #330, opened 2016-07-30
Fix FlexTable table formatting

maxthyen:
Minor, but the table formatting for one of the tables in `FlexTable.md` was off.

Before / After:

image

Thanks!


bvaughn:
Oh boy. I need to read my own docs more carefully when I update them.

Thanks for the fix :)

PR #340, opened 2016-08-08
Reset grid rendered memoizer when underlying data changes

sakari:
I have a virtual scrolling table that displays an array of data. The array contains regular row data and possibly multiple loading indicators. When a loading indicator becomes visible (ie. I get onRowsRendered callback) I make a fetch and replace the single loading indicator row with the resulting data rows. This works surprisingly well.

However as Grid memoizes the `_onSectionRendered` calls I do not always get the the callback to `onRowsRendered` when I insert a new loading indicator row in the visible range.

This pull request provides a `backingData` prop that the Grid can look at to determine whether the memoization should be forgot and generally indicate that the data backing the indexes has changed.

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -414,6 +414,11 @@ export default class Grid extends Component {
    * 3) Cells-count or cells-size has changed, making previous scroll offsets invalid
    */
   componentWillUpdate (nextProps, nextState) {
+    if (nextProps.backingData !== this.props.backingData) {
+      // reset memoization so that load status will be queried again
+      this._onGridRenderedMemoizer = createCallbackMemoizer()
+      this._onScrollMemoizer = createCallbackMemoizer(false)
+    }
     if (
       nextProps.columnCount === 0 &&
       nextState.scrollLeft !== 0 ||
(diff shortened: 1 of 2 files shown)


bvaughn:
Maybe I'm missing something, but... why is this necessary?

Sounds like there are 2 cases you care about:
* New rows have been loaded as a result of user scrolling
* New data has loaded as the result of a fetch

It seems to me like the existing Api should enable this. Basically, assuming you've passed an `onRowsRendered` callback to your `FlexTable` (or whatever component you're using) then...

```js
class MyComponent extends React.Component {
  _onRowsRendered ({ startIndex, stopIndex }) {
    // Cache the most recently rendered rows
    this._startIndex = startIndex
    this._stopIndex = stopIndex

    this._handleTheEventYouCareAbout()
  }

  _onFetchComplete (response) {
    // Process fetch `response` however you want, then...
    this._handleTheEventYouCareAbout()
  }

  _handleTheEventYouCareAbout () {
    // The currently-visible rows are this._startIndex...this._stopIndex
  }
}
```

sakari:
Something like that and yes that is a possibility yes. However I feel it would be nicer if this detail would be hidden from the client of the virtualscroller/infiniteloader combo. Anyway this caching behaviour was surprising at least to me.

If you feel that this is better done outside react-virtualized then feel free to close this PR.

bvaughn:
> Anyway this caching behaviour was surprising at least to me.

Understood. I will update the documentation to make this clearer.

I would prefer to solve this in user-land because my suspicion is that your use-case is not a very common one. We can always re-visit this PR if I turn out to be wrong.

Thank you for the contribution :)

PR #346, opened 2016-08-12
[WindowScroller] correct initial _positionFromTop if page scrolled

igorbt:
In the case when page already have a scrolling on component mount, _positionFromTop should take this into consideration when using getBoundingClientRect for calculations. The most common use-case is when navigating back in history page is rendered directly with scroll.

Because PhantomJS seems to ignore window scrolling and return same getBoundingClientRect for elements, all tests are passing without any changes and I coundn't find any way to improve them to test the fix. Maybe setting a negative top or translateY for documentElement would work to reproduce this use-case in unit-tests, but that's a bit weird solution.

### DIFF
--- a/source/WindowScroller/WindowScroller.js
+++ b/source/WindowScroller/WindowScroller.js
@@ -46,7 +46,8 @@ export default class WindowScroller extends Component {
   }
 
   componentDidMount () {
-    this._positionFromTop = ReactDOM.findDOMNode(this).getBoundingClientRect().top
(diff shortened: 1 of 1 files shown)


bvaughn:
Hey @igorbt,

Afraid I don't understand the description of this bug, or the fix. Why would the root `<html>` element have a non-zero `top`?

bvaughn:
Ah. I was unable to reproduce the issue you reported in Chrome, but I can reproduce it in Firefox. Nice catch. :)

bvaughn:
Thanks for the PR! Fix went out as 7.19.2.

PR #347, opened 2016-08-12
[WindowScroller] restore pointerEvents on body on unmount

igorbt:
Restoring pointerEvents on body right away on unmount, without a timeout, will also prevent react warning about setState on not mounted component. Basically, that warning was the trigger for this PR.

Added 2 tests about pointerEvents on scrolling and on unmounting.

BTW, @bvaughn I had a hard time writing these unit-tests partially because little experience with all the tooling, but also because running the tests took me a lot of time. I would love to see something like [karma-webpack-with-fast-source-maps](https://www.npmjs.com/package/karma-webpack-with-fast-source-maps) used here. This would help future contributions on this great project. What do you think? Can you do this?

### DIFF
--- a/source/WindowScroller/WindowScroller.js
+++ b/source/WindowScroller/WindowScroller.js
@@ -8,7 +8,7 @@ import raf from 'raf'
  * Specifies the number of miliseconds during which to disable pointer events while a scroll is in progress.
  * This improves performance and makes scrolling smoother.
  */
-const IS_SCROLLING_TIMEOUT = 150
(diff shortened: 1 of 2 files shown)


bvaughn:
Thanks for the PR @igorbt. Looks like a nice addition. :)

I agree with you that the tests take a while to run. I'm not familiar with the fork you pointed at though. If you'd be interested in taking a pass at speeding up the tests- I'd welcome another PR! :)

bvaughn:
This change has been released as version 7.19.1

PR #348, opened 2016-08-12
[Grid] render inner scroll container based on rowCount

igorbt:
Rendering inner scroll container based on rowCount and columnCount rather than based on childrenToDisplay should keep current behavior in most use-cases, but it fixed one particular use-case, that was important for my app:

When going back to a page that renders a Grid wrapped with WindowScroller it should come back to the same scroll position in grid. 
Instead, when the previous scroll was big enough (in my case > 10000px) it failed to return to that scroll. This happen basically because Grid didn't render the scroll inner container fast enough and browser saw there is not enough height in the document and just didn't scroll. I'm not sure if this would be easy to reproduce on a simple example, but it was 100% reproducible in my project. Again, I'm not sure if it's a specific browser buggy behavior or a kind of a race condition my app has hit, but this change fixed the issue for me and hopefully will help others too.

I didn't add any unit-tests because it would probably be really really hard to reproduce the bug in a testing environment.

Beside that, the change basically is more aligned to the docs:

```
| noContentRenderer | Function |  | Optional renderer to be rendered inside the grid when either `rowCount` or `columnCount` is empty: `(): PropTypes.node` |
```

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -482,7 +482,9 @@ export default class Grid extends Component {
       noContentRenderer,
       style,
       tabIndex,
-      width
+      width,
+      rowCount,
+      columnCount
     } = this.props
 
     const { isScrolling } = this.state
@@ -516,7 +518,7 @@ export default class Grid extends Component {
     const childrenToDisplay = this._childrenToDisplay
 
     const showNoContentRenderer = (
-      childrenToDisplay.length === 0 &&
+      (rowCount === 0 || columnCount === 0) &&
       height > 0 &&
       width > 0
     )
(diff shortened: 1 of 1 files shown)


bvaughn:
The `_childrenToDisplay` instance property is calculated on `componentWillMount` and `componentWillUpdate`. This calculation takes `columnCount` and `rowCount` into consideration, so this change should not be necessary. It's possible that there's a bug- (in which case I'd love to know what's in your `childrenToDisplay` array)- but I think the bug should be fixed in `CellSizeAndPositionManager` if so.

bvaughn:
Added a sanity-check test to cover this functionality (either `rowCount` or `columnCount` changing to/from 0) in commit 1bc73fd. Looks like it's working as intended, which makes me think this PR is not necessary.

igorbt:
@bvaughn , I found the root cause of the problem in my use-case. It was because of `WindowScroller` components that only sets a height on `componentDidMount`. This was causing `_childrenToDisplay` to be empty. I will close this PR and will open another one with the fix.

PR #360, opened 2016-08-25
Collection: cache cell position/size information while scrolling

coluccini:
To avoid unnecessary rendering I've implement a cache for cell's position information in almost the same way the Grid component does

### DIFF
--- a/source/Collection/Collection.js
+++ b/source/Collection/Collection.js
@@ -5,6 +5,7 @@ import getUpdatedOffsetForIndex from '../utils/getUpdatedOffsetForIndex'
 import shallowCompare from 'react-addons-shallow-compare'
 import type { ScrollPosition, SizeInfo } from './types'
 
+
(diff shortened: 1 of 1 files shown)


bvaughn:
Huzzah! Thanks! I'll find time to review this change sometime this morning and- hopefully- push out a bugfix release. :)

bvaughn:
It doesn't seem the rendered cell cache is ever being cleared after scrolling stops (which will cause memory leaks). I will need to think about this a bit.

bvaughn:
Version 7.20.0 has cell caching in `Collection`. Thanks ;)

coluccini:
Awesome! Thanks a lot to you not only because the amazing work you've done with the component but, specially, because how you actively responde to any doubt/request from the developers using it. Your the best!

bvaughn:
Aw, thanks :)

PR #362, opened 2016-08-27
Add CellMeasurer size cache strategy for uniform-yet-unknown row heights

guigrpa:
As discussed in 9839c52d96b6cec72a58cbf550bc78c57728a97b, I have added a new size cache strategy for the case in which we know that all rows have the same height, even though this height is unknown. This use case can be quite common (e.g. custom drop-down menus). The new class is exported as `CellMeasurerCellSizeCacheUniformHeight`.

I have updated the related documentation and included a unit test.

### DIFF
--- a/docs/CellMeasurer.md
+++ b/docs/CellMeasurer.md
@@ -52,11 +52,11 @@ class CellSizeCache {
 }
 ```
 
-The [default caching strategy](https://github.com/bvaughn/react-virtualized/blob/master/source/CellMeasurer/defaultCellSizeCache.js) is exported as `defaultCellMeasurerCellSizeCache` should you wish to decorate it.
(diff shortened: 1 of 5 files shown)


bvaughn:
Look at that! You even added tests and docs- and corrected some of my typos in existing docs. <3

Excited to review this shortly.

bvaughn:
FYI, I'm going to make a few changes to the uniform cache so that it can be used for fixed width _or_ height. This won't change its behavior for your use case but will hopefully make it more flexible for others. :)

bvaughn:
Available in 7.21.0~ exported as `uniformSizeCellMeasurerCellSizeCache`. Thanks for your contribution! :)

guigrpa:
Thanks to you for sharing this extremely useful library, and for your time!

guigrpa:
Hmmm, I believe the current implementation forces the user to have uniform row heights *and* column widths. Maybe we can come up with a cache implementation that is configured upon instantiation and works in the general case: uniform heights *and/or* uniform widths. Let me have a go at it — is it possible to commit to an already merged PR or should I open a new one?

bvaughn:
Ah, I see. I believe I misunderstood the use-case. Should be an easy fix. I can make it configurable via a constructor arg. 😄 Unless you'd rather tackle it?

bvaughn:
I'll hop on it this morning. Should be fast. Give me a few.

guigrpa:
Oops, we crossed each other. See #363

PR #363, opened 2016-08-28
Add CellMeasurer size cache strategy for uniform-yet-unknown row heights (follow-up)

guigrpa:
This is a follow-up on #362. In the merged solution, the size cache imposes that both row height and col width are uniform, which might not be what the user wants. In this PR, the cache can be configured upon construction (defaults being `uniformWidth: true, uniformHeight: true`).

### DIFF
--- a/source/CellMeasurer/uniformSizeCellSizeCache.js
+++ b/source/CellMeasurer/uniformSizeCellSizeCache.js
@@ -3,43 +3,58 @@
  * Permanently caches cell sizes until explicitly cleared.
  */
 export default class UniformSizeCellSizeCache {
+  constructor ({ uniformWidth = true, uniformHeight = true } = {}) {
+    this.uniformWidth = uniformWidth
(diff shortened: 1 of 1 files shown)


bvaughn:
Hey, looks like we crossed paths. My bad. 😄

I've got a local branch of this that adds more tests and also updates `defaultCellSizeCache` to use the new, configurable `uniformCellSizeCache`~ so I'm going to just roll with that. It's almost ready for release. I'll tag you again here in a few.

guigrpa:
OK, great!

bvaughn:
Okay! Check out release 7.22.0 😄 Hopefully this is more flexible.

[Release notes](https://github.com/bvaughn/react-virtualized/blob/master/CHANGELOG.md#7220) and [`CellMeasurer` docs](https://github.com/bvaughn/react-virtualized/blob/master/docs/CellMeasurer.md#customizing-cellsizecache) have been updated.

PR #378, opened 2016-09-02
Run tests using jest

spicyj:
AutoSizer, CellMeasurer, and WindowScroller fail because jsdom doesn't do measurement stuff properly and you need a real browser. The other tests work with `./node_modules/.bin/jest` though.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -59,9 +59,6 @@
     },
     "test": {
       "comments": false,
-      "plugins": [
-        "__coverage__"
-      ],
       "presets": [
         "es2015",
         "react",
--- a/package.json
+++ b/package.json
@@ -78,11 +78,17 @@
       "jasmine"
     ]
   },
+  "jest": {
+    "setupFiles": [
+      "source/jest-setup.js"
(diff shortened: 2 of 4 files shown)


spicyj:
I'm guessing you don't want to take this because you can't test real browser layout stuff or any measurements with this. Figured I'd send it so you can see what it would be though – it was pretty easy to make the others pass.

bvaughn:
Oh hey Ben. This is great! Very minimal changes. Thanks for sharing! 😁

I'm working on react-select today but I'll give this branch a spin ASAP.

bvaughn:
This looks promising! Test time drops from ~28 seconds to ~22 without any other changes.

bvaughn:
Wow! Test time drops from ~28 seconds to ~9. Unfortunately 18 tests fail. But still! 😮 Seems worth pursuing. 😁

bvaughn:
Took a few months- but Jasmine has been replaced entirely by Jest. Thanks for starting this discussion Ben! 😄

PR #390, opened 2016-09-14
Utilize babel-plugin-transform-runtime

raizyr:
Allow `react-virtualized` to be usable without requiring `babel-external-helpers` to be included in the application code.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -41,6 +41,9 @@
       ]
     },
     "es": {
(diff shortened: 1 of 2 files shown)


bvaughn:
Thanks @raizyr!

As I said before, I'm not a Babel expert- but this looks like it fixes the react-boilerplate issue. So thanks!

Issue #402, opened 2016-09-20
List docs use `rowIndex` in example `rowRenderer`


IanVS:
The [List example](https://github.com/bvaughn/react-virtualized/blob/master/docs/List.md#examples) in the docs uses the proposed-but-not-implemented `rowIndex`:

```js
function rowRenderer ({ key, rowIndex, style}) {
  return (
    <div
      key={key}
      style={style}
    >
      {list[rowIndex]}
    </div>
  )
}
```

bvaughn:
Toss up a PR to fix it? :)

bvaughn:
I'm not at my computer atm.

IanVS:
> Toss up a PR to fix it?

Of course!

bvaughn:
Thanks! :grin:

PR #403, opened 2016-09-20
Update docs for List

IanVS:
Changing `index` to `rowIndex` was decided against, so this reverts that change in the example.

This also adds `key` and `style` to the `rowRenderer` function signature.

Fixes: #402


IanVS:
Don't merge yet, it's in a few other places too.

bvaughn:
Thanks dude! :)

PR #404, opened 2016-09-20
Fix a few more docs examples

IanVS:
#403 got merged a bit [too quick](https://github.com/bvaughn/react-virtualized/pull/403#issuecomment-248195711) ;).  Here are a few more places in the docs which needed to be changed back to `index` as well.

### DIFF
--- a/docs/AutoSizer.md
+++ b/docs/AutoSizer.md
@@ -34,13 +34,13 @@ const list = [
   // And so on...
 ];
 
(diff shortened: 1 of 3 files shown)


IanVS:
Feel free to squash the commits.  I was lazy and didn't change the commit message (had already squashed locally)

bvaughn:
Thanks again :)

Issue #398, opened 2016-09-19
Support passive scroll listeners


romulof:
Make `Grid` use passive scroll listeners (when available) for huge performance improvements.
Reference: https://developers.google.com/web/updates/2016/06/passive-event-listeners

Since React does not support passive event listeners (yet: https://github.com/facebook/react/issues/6436), we have first to test for browser support using the following module (coded in es7 + flowtype):

```es6
/* ListenerOptions.js
 * @flow
 */
let passive: boolean = false;

const testOptions = {
  get passive() {
    passive = true;
  },
};

try {
  // $FlowIssue: addEventListener type hint does not support options yet
  document.addEventListener('test', null, testOptions);
} catch (e) {
  // ignore
}

export default {
  hasPassive: passive,
};

export function addPassiveEvent(element: HTMLElement, eventName: string, handler: EventHandler) {
  const options = passive ? { passive: true } : false;
  // $FlowIssue: addEventListener type hint does not support options yet
  element.addEventListener(eventName, handler, options);
}
```

Then instead of attaching using react's `onScroll` component prop, get the elements ref and add the event handler directly (or using the helper function in `ListernerOptions`).

```es6
/* @flow */
import React from 'react';
import { addPassiveEvent } from './ListenerOptions';


export default class PassiveScroller extends React.Component {
  onScrollerRef = (scrollerElement:HTMLElement) => {
    if (!scrollerElement) {
      if (this.scrollerElement) {
        // Prevent memory leaks
        this.scrollerElement.removeEventListener('scroll', this.onScroll);
      }

      this.scrollerElement = null;
      return;
    }

    this.scrollerElement = scrollerElement;
    addPassiveEvent(scrollerElement, 'scroll', this.onScroll);
  };

  onScroll:EventHandler = (event) => {
    // Normal scroll handler
  };

  scrollerElement: ?HTMLElement = null;

  render() {
    return (
      <div
        ref={this.onScrollerRef}
      >
        {this.props.children}
      </div>
    );
  }
}
```

I can submit the PR if there are no objections.

romulof:
Browser support:
- chrome: shipped in 51
- firefox: shipped in 49
- webkit: shipped in june: iOS 10 and Safari 10 (to be released)

bvaughn:
Hey @romulof! Thanks for the proposal. I'm definitely interested, if you'd be willing to put together a POC.

I have some initial concerns but we should test it out and see. My concerns are:

[...]

PR #407, opened 2016-09-21
Passive event listeners

dxtr026:
adding {passive : true} if passiveEventListeners are supported else let the capture flag default to false.

https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md
https://github.com/GoogleChrome/ui-element-samples/pull/32

### DIFF
--- a/source/WindowScroller/utils/onScroll.js
+++ b/source/WindowScroller/utils/onScroll.js
@@ -1,3 +1,5 @@
+import { isPassiveSupported } from './supportsPassive'
+
 let mountedInstances = []
 let originalBodyPointerEvents = null
 let disablePointerEventsTimeoutId = null
@@ -47,7 +49,7 @@ function onScrollWindow (event) {
 
 export function registerScrollListener (component) {
   if (!mountedInstances.length) {
-    window.addEventListener('scroll', onScrollWindow)
+    window.addEventListener('scroll', onScrollWindow, (isPassiveSupported ? {passive: true} : false))
   }
   mountedInstances.push(component)
 }
@@ -55,7 +57,7 @@ export function registerScrollListener (component) {
 export function unregisterScrollListener (component) {
   mountedInstances = mountedInstances.filter(c => (c !== component))
   if (!mountedInstances.length) {
-    window.removeEventListener('scroll', onScrollWindow)
+    window.removeEventListener('scroll', onScrollWindow, (isPassiveSupported ? {passive: true} : false))
     if (disablePointerEventsTimeoutId) {
       clearTimeout(disablePointerEventsTimeoutId)
       enablePointerEventsIfDisabled()
--- a/source/WindowScroller/utils/onScroll.js
+++ b/source/WindowScroller/utils/onScroll.js
@@ -49,7 +49,7 @@ function onScrollWindow (event) {
 
 export function registerScrollListener (component) {
   if (!mountedInstances.length) {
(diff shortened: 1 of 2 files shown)


bvaughn:
Hi @dxtr026. 😄

I have a branch locally where I've been testing passive event listeners with `Grid` but I haven't noticed _any_ performance improvements with my own test harness and Chrome's Timeline view. I wonder if you've seen improvements with the `WindowScroller` component? And if so, how much of an improvement?

I'm open to doing something non-standard (like using a passive listener instead of the synthetic scroll event) but only if I'm confident that it offers non-trivial performance gains. 😁

dxtr026:
@bvaughn same here, couldn't find any concrete results to actually see the performance improvement. I tried doing tests but I think Chrome auto handles passive listeners and that's the reason we are not able to capture any difference.

bvaughn:
Ok. I plan to look into it more, perhaps over the weekend. As soon as I find time.

bvaughn:
Closing this (for now) as it adds complication without measurably improving performance. If you find anything else related to this effort though please let me know. I'm happy to continue looking into passive listeners. I think there's potential there. I just don't want to make the move until there's concrete gains.

See my comments on issue #398 for more.

azizhk:
Example: https://rbyers.github.io/scroll-latency.html
So Chrome assumes passive by default. And if you preventDefault a scroll event for any further events, Chrome will run your function before paint.
Check it out in the example, initially if you scroll, you should get 60fps, after to tick the preventDefault button, scroll, untick and then scroll, you will get degraded performance. Then you need to set passive:true to get good performance.

So in Chrome you will get the best performance out of the box. if you don't preventDefault, you don't need to set passive:true

Plus timeline will actually not tell you what frame rate you are getting. You will need to use the FPS Meter. For example, I get 60fps while scrolling a long PR in github (https://github.com/openstreetmap/iD/pull/8/files) But timeline tells me that its actually slow.
FPS Meter for Passive Event Listeners
 
Timeline Example for Passive Event Listeners

bvaughn:
I think the Timeline is pretty trustworthy (and more detailed than the Frame Rate viewer). Looking at the timeline view you've pasted- there's a ton of repainting going on. That would make for some long frames.

Scrolling the page at the PR you posted _does_ feel pretty janky to me at first (and the Timeline view kind of validates this). Not sure why the Frame Rate view doesn't pick up on it. Things level out after I've scrolled for a bit.

azizhk:
Ok ditch the github example.
Take the first example: https://rbyers.github.io/scroll-latency.html
Apply 50ms Handler Jank, And then tick passive.
Now scroll in the page and record timeline and see the FPS meter.
Timeline will tell you that you are getting a slow frame rate while you are getting smooth scroll.
screen shot 2016-10-10 at 12 03 04 am

> Closing this (for now) as it adds complication without measurably improving performance.

The point I'm trying to make is that because Passive Event Listeners are a new feature, they are not correctly recorded in the Timeline, but there is significant performance gain.

bvaughn:
This is interesting. Point recognized.

I have a lot on my plate right now, preparing to start a new job on Monday and presenting at a conference next weekend- so I probably won't be able to pick this thread back up for a bit. But it's on my list. 😄

azizhk:
Thanks Brian,
No issues. Take your time. Best of Luck with your new job and conference. Rock it.
Sorry for my rant.

PR #411, opened 2016-09-21
match InfiniteLoader onRowsRendered callback with Grids onSectionRendered signature

gravitypersists:
Currently, if you want to use Grid with InfiniteLoader, you can map `onRowsRendered` to `onSectionRendered` like so:

```
<InfiniteLoader>
  {({ onRowsRendered }) => (
    <Grid onSectionRendered={(x) => onRowsRendered({startIndex: x.rowStartIndex, stopIndex: x.rowStopIndex})} />
  )}
</InfiniteLoader>
```

This just adapts the signature for `InfiniteLoader`s callback to read `startIndex` first, and then `rowStartIndex` second if that was null or undefined.

```
<InfiniteLoader>
  {({ onRowsRendered }) => (
    <Grid onSectionRendered={onRowsRendered})} />
  )}
</InfiniteLoader>
```

I'd be happy to write a test for this, but please let me know first if this is a welcome change. 

(sidenote: I am really enjoying using this library, it's one of the best I've had a pleasure to use, but the differences between Tables and Grids seems to be a bit of a pain point, in my view).

### DIFF
--- a/source/InfiniteLoader/InfiniteLoader.js
+++ b/source/InfiniteLoader/InfiniteLoader.js
@@ -108,18 +108,18 @@ export default class InfiniteLoader extends Component {
     })
   }
 
-  _onRowsRendered ({ startIndex, stopIndex }) {
+  _onRowsRendered ({ startIndex, stopIndex, rowStartIndex, rowStopIndex }) {
     const { isRowLoaded, minimumBatchSize, rowCount, threshold } = this.props
 
-    this._lastRenderedStartIndex = startIndex
-    this._lastRenderedStopIndex = stopIndex
+    this._lastRenderedStartIndex = (startIndex != null) ? startIndex : rowStartIndex
+    this._lastRenderedStopIndex = (stopIndex != null) ? stopIndex : rowStopIndex
 
     const unloadedRanges = scanForUnloadedRanges({
       isRowLoaded,
       minimumBatchSize,
       rowCount,
-      startIndex: Math.max(0, startIndex - threshold),
-      stopIndex: Math.min(rowCount - 1, stopIndex + threshold)
+      startIndex: Math.max(0, this._lastRenderedStartIndex - threshold),
+      stopIndex: Math.min(rowCount - 1, this._lastRenderedStopIndex + threshold)
     })
 
     // For memoize comparison


bvaughn:
Interesting. I had considered doing this a time or two but it felt a _little_ hacky. Still... I see the appeal. Let me give it some more thought. I'm still on the fence. 😄 I do appreciate the PR and the offer to write tests (which I'll definitely take you up on if we move forward)!

Thank you for the compliment too.

bvaughn:
Sorry for letting this sit for so long. I've actually come back to this PR a few times, read the code, and still felt undecided. I thought I'd share my thinking rather than continue to remain silent.

On the one hand it's a simple, safe change. On the other though- I worry about the precedent it may set. There are other connection points between a `List`/`Grid` and the HOCs that decorate them (eg `onRowsRendered` vs `onSectionRendered`, `rowRenderer` vs `cellRenderer`, `scrollToRow` vs `scrollToIndex`, etc).

I'm concerned that by making it "easier" (getting rid of the adapter layer) to mix and match `InfiniteLoader`- we may actually be making things _more_ confusing when matching the other HOCs. That is, unless we go and update all possible properties to be cross-compatible, which is more forking that I feel comfortable with.

gravitypersists:
Another option would be to pass `onSectionRendered` out from InfiniteLoader, though I'm not sure this eases your concerns. It does seem cleaner than my first approach, and doesn't set precedent of Grid/Table having the same connection points, but also begs the method `loadMoreColumns`.

bvaughn:
I'm not sure what you mean by "pass `onSectionRendered` out". Are you suggesting that `InfiniteLoader` could pass its child function a callback named `onSectionRendered` as well as the current `onRowsRendered` ? If so, you're right. That just kind of begs us to do the same with `loadMoreRows`.

I _think_ that... given the relatively small amount of adapter code the current version requires to make the 2 play together... I might be leaning towards leaving things as-is for now.

I temporarily considered renaming `List` and `Table` properties to use `rowIndex` instead of `index` so that all 3 would match. But once I started implementing the rename- I changed my mind because the new names felt unnecessarily long.

gravitypersists:
Okay. But I'm going to end with stating, again, that I think a huge place for improvement here is in more unifying of Table and Grid!

bvaughn:
Thank you for understanding _and_ for proposing ideas and discussing their pros and cons. I really appreciate both of those, especially being a solo maintainer. 🙇

I will continue to think about this and I'd love for you to share any additional suggestions or ideas you have as well.

PR #412, opened 2016-09-21
Adding natural sorting to FlexTable

sriverag:
Now sorting can be reset to its original order (natural), after clicking a sortable column header.

The new order of sorting direction is: 
NAT, ASC, DESC, NAT... 

Previously it was: 
NAT, ASC, DESC, ASC, DESC...

### DIFF
--- a/source/Table/SortDirection.js
+++ b/source/Table/SortDirection.js
@@ -1,5 +1,11 @@
 const SortDirection = {
   /**
+   * Sort items in natural (original) order.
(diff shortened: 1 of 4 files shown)


bvaughn:
Thanks for the PR!

We chatted via email and agreed to keep this in application code for now so I'm going to close this. We can reopen it later if things change.

sriverag:
No problem, thanks for your help.

PR #413, opened 2016-09-23
Fix blocked safari scrollbar

fritz-c:
The `detectElementResize.js` library has a bug in Safari in which an invisible element covers the scrollbar and prevents clicks from going through to the control.
before

This is documented in [this issue](https://github.com/sdecima/javascript-detect-element-resize/issues/31) and I got the idea for the fix from [this commit](https://github.com/poscar/javascript-detect-element-resize/commit/0a80f3ec58ab6de6e5491898bcfe263b04032ae6).

The result:
after


bvaughn:
Thanks, @fritz-c !

PR #415, opened 2016-09-23
Update package.json - add "files" field

tkh44:
Added "files" field to package.json to prevent the playground files from being downloaded when doing npm install.

I'm pretty sure the dist folder is all you need since everything lives in there on build.  package.json, changelog, readme, and license are always included as per https://docs.npmjs.com/files/package.json#files

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,9 @@
     "type": "git",
     "url": "https://github.com/bvaughn/react-virtualized.git"
   },
+  "files": [
+    "dist"
+  ],
   "keywords": [
     "react",
     "reactjs",


bvaughn:
I have been using `.npmignore` for this purpose. But you're right, I had missed the `playground` folder. Unfortunately your PR also misses the `styles.css` file which is a key component of the dist so I'm going to close this PR. I've addressed the playground folder issue in master (and it's not included in the latest release, 8.0.8)

Thanks~

tkh44:
No problem, just wanted to point it out and do something 👍

bvaughn:
Much appreciated :)

Issue #423, opened 2016-09-29
Demo site has IE11 flexbox height bug


lRawd:
**Edit for clarity**:
This bug affects IE11 only.

The main body/content of the demo site is clipped due to an IE11 flexbox height bug. Adding `height: 100%` to a handful of the CSS classes fixes this problem but breaks the `WindowScroll` demo in Chrome (as well as making the height too tall in Safari.)

Screenshot from https://bvaughn.github.io/react-virtualized/#/components/List
![react2](https://cloud.githubusercontent.com/assets/17622807/18971481/ce83245c-865a-11e6-92ea-179df21c0f7d.png)

bvaughn:
Bleh. IE11 specific CSS issue. 😦

Thanks for letting me know. It's been fixed.

bvaughn:
Unfortunately the fix broke the `WindowScroller` demo in Chrome so I have reverted it.

Would welcome assistance on this one. I can only run IE in a crappy VM _and_ I'm on vacation this week so I don't have a lot of time to look at it. Any chance you would be willing to lend a hand? :)

PR #425, opened 2016-10-03
Flexbox height fix for IE, based on https://github.com/philipwalton/f…

lRawd:
Fixes for issue #423 

Based on https://github.com/philipwalton/flexbugs#3-min-height-on-a-flex-container-wont-apply-to-its-flex-items


bvaughn:
Thanks so much! 😄

Issue #424, opened 2016-09-30
VirtualScroll (List) with dynamic item height scrolling not smooth and jumping


EllisShen:
I have been tuning the VirtualScroll (List) component for almost whole day but no luck.
I'm building a web based chatting application in which uses the react-virtualized List to display the chatting messages. Since message may have different content and different height, I use **react-measure** to calculate the item height and issue the **recomputeRowHeights** in rowRenderer.

The result is bad, VirtuallScroll List will jump around whenever I stopped the scrolling. For example, when I scrolled to the half of browser, I should see the middle of the messages, but it always suddenly shift the offset. Please take a look at the recorded video:
https://drive.google.com/file/d/0B_W64UoqloIkcm9oQ08xS09Zc1k/view?usp=sharing

Since I only use the List and Autosizer component, I only adapt the required css file into my project which is like
```
.VirtualScroll {
    width: 100%;
    outline: none;
}
```

For the render method, I nested a lot of flex components inside the rowRender:
Here is the code:
```
    render() {
        const inChat = this.context.store.getState().inChat;
        const {conversationList} = this.state;
        const imgUrl = 'img/builtin-wallpaper-1.jpg';
        const backgroundStyle = {
            backgroundImage: 'url(' + imgUrl + ')',
            backgroundRepeat: 'no-repeat',
            backgroundSize: 'cover',
            backgroundPosition: 'top left'
        };

        // console.log('inChat');
        // console.log(inChat);

        if (inChat.id === this.id && inChat.status === 'FETCHING'){
            return (
                <Box column center height="80%">
                    <CircularProgress />
                </Box>
            );
        } else if (inChat.id === this.id && inChat.status === 'FETCHED'){
            return (
                <Box column flex="1 0 100%" style={backgroundStyle}>
                    <HorizontalToolBar/>
                    <AutoSizer disableHeight={true}>
                        {({ width }) => (
                            <List
                                ref={(element) => {this.VirtualScroll = element;}}
                                className='VirtualScroll'
                                height={window.innerHeight - toolbarHeight - textAreaHeight}
                                overscanRowCount={10}
                                noRowsRenderer={this._noRowsRenderer.bind(this)}
                                rowCount={conversationList.length}
                                rowHeight={i => {
                                    return (Measured_Heights[i.index] | 20);  // default Height = 58
                                }}
                                rowRenderer={this._rowRenderer}
                                scrollToIndex={undefined} // scroll to latest item
                                width={width}
                            />
[...]

PR #428, opened 2016-10-08
Emphasize new params in List docs

bhj:
See #424 

Also went ahead and removed that stylesheet from the basic List example since it wasn't doing anything as of 8.x.


bvaughn:
Thanks :)

PR #432, opened 2016-10-11
Fix an issue with testing in Enzyme

DevinClark:
Enzyme, or upstream of Enzyme, aliases window methods (like getComputedStyle) to their own methods. When these are called like `getComputedStyle()` rather than `window.getComputedStyle()` they are interpreted as `global.getComputedStyle()` which does not exist.

We were fixing this by adding this to our Enzyme setup file.
```js
global.getComputedStyle = window.getComputedStyle;
```

Tests and linting passed locally. I also played around with the playground to make sure everything still worked there.

### DIFF
--- a/source/AutoSizer/AutoSizer.js
+++ b/source/AutoSizer/AutoSizer.js
@@ -104,7 +104,7 @@ export default class AutoSizer extends Component {
     const height = boundingRect.height || 0
     const width = boundingRect.width || 0
 
-    const style = getComputedStyle(this._parentNode)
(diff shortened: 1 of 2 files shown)


bvaughn:
Thanks @DevinClark !

bvaughn:
Just released 8.0.13 with this change.

Shoutout/thanks to you [in the `CHANGELOG`](https://github.com/bvaughn/react-virtualized/blob/master/CHANGELOG.md#8013) as well. 😄

DevinClark:
Happy to help, @bvaughn! 😃

PR #436, opened 2016-10-13
Remove the animation event listener while removing the resize listener

yp:
When mounting and unmounting very quickly a complex component on IE11 and a quite old computer as a child of AutoSizer, the `resetTriggers` function gets called *after* the `removeResizeListener` has been called. Hence `resetTriggers` fails because `element.__resizeTriggers__` is `false`.
Since such a trigger is called by the animation event listener, this pull request tracks and removes such a listener.

PS: The test case highlighting this problem is quite complex and it could be (I believe, at least) time-dependant, but I could spend some time if it is needed.

### DIFF
--- a/source/vendor/detectElementResize.js
+++ b/source/vendor/detectElementResize.js
@@ -143,10 +143,15 @@ var addResizeListener = function(element, fn){
       element.addEventListener('scroll', scrollListener, true);
 
       /* Listen for a css animation to detect element display/re-attach */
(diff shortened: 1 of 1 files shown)


bvaughn:
Thanks @yp. This looks like a nice fix. 😄 I'll include it with the next release, 8.1.0~ hopefully going out shortly.

bvaughn:
This change has been released with version 8.1.0. Special note of thanks to you in the CHANGELOG. 😄

Thanks again @yp!

yp:
Thanks for the quick release!

PR #438, opened 2016-10-14
Remove proptypes in production

mbrevda:
As react doesn't check proptypes in production, there is no point in keeping them in the build. Removing them make the build size ever so slightly smaller.

~~While I didn't see tests passing locally, they don't seem to be passing on master, either.~~ If CI says its good, then its good!

### DIFF
--- a/source/Table/Column.js
+++ b/source/Table/Column.js
@@ -8,16 +8,6 @@ import defaultCellDataGetter from './defaultCellDataGetter'
  * Describes the header and cell contents of a table column.
  */
 export default class Column extends Component {
-
-  static defaultProps = {
-    cellDataGetter: defaultCellDataGetter,
-    cellRenderer: defaultCellRenderer,
-    flexGrow: 0,
-    flexShrink: 1,
-    headerRenderer: defaultHeaderRenderer,
-    style: {}
-  }
-
   static propTypes = {
     /** Optional aria-label value to set on the column header */
     'aria-label': PropTypes.string,
@@ -75,5 +65,16 @@ export default class Column extends Component {
 
     /** Flex basis (width) for this column; This value can grow or shrink based on :flexGrow and :flexShrink properties. */
     width: PropTypes.number.isRequired
+  }
+}
+
+if (process.env.NODE_ENV !== 'production') {
+  Column.defaultProps = {
+    cellDataGetter: defaultCellDataGetter,
+    cellRenderer: defaultCellRenderer,
+    flexGrow: 0,
+    flexShrink: 1,
+    headerRenderer: defaultHeaderRenderer,
+    style: {}
   }
 }
--- a/source/ArrowKeyStepper/ArrowKeyStepper.js
+++ b/source/ArrowKeyStepper/ArrowKeyStepper.js
@@ -6,13 +6,6 @@ import shallowCompare from 'react-addons-shallow-compare'
  * This HOC decorates a virtualized component and responds to arrow-key events by scrolling one row or column at a time.
  */
 export default class ArrowKeyStepper extends Component {
-  static propTypes = {
-    children: PropTypes.func.isRequired,
-    className: PropTypes.string,
-    columnCount: PropTypes.number.isRequired,
-    rowCount: PropTypes.number.isRequired
-  }
-
   constructor (props, context) {
     super(props, context)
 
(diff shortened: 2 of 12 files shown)


bvaughn:
Hey @mbrevda!

Stripping `propTypes` in production is a great idea- but I think it belongs in application land rather than in react-virtualized. (If `propTypes` are stripped from RV, then users won't get helpful validation messages when developing their applications using RV.)

Fortunately this can be accomplished after the fact (and easier!) using the [babel-plugin-transform-react-remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types) plug-in.

Out of curiosity, I ran it over `Grid` and the CommonJS version shrinks from 40k to 35k. So it's definitely worth doing (as an application build step)!

mbrevda:
If I may make a (possible erroneous) counter argument: running `npm run
build` didn't seem to strip the proptypes from the built dist files. This
means that devs would still be able to take advantage of them AND have them
auto stripped if they follow typical react good practices - all without
having to add extra globally affecting plugins.

bvaughn:
> running `npm run build` didn't seem to strip the proptypes from the built dist files.

What do you mean? Within react-virtualized? That's because I don't use that plugin to strip them (because I think they're useful to leave in place for application developers). But they can _later_ be stripped when the application is built using a plug-in like the one I mentioned above.

I'm afraid I don't understand the counter-argument?

mbrevda:
> That's because I don't use that plugin to strip them

Good, then they will be there during development.

> when the application is built using a plug-in

That's assuming a plugin is, indeed, used. It's also assuming that babel
will operate on `node_modules` (which it usually doesn't).

Wouldn't it be simpler - and just as useful - to surround proptypes with a
NODE_ENV check?

bvaughn:
>>  when the application is built using a plug-in

> That's assuming a plugin is, indeed, used. It's also assuming that babel
will operate on `node_modules` (which it usually doesn't).

Ah! I understand the point you're making now. Okay, let me re-open this PR and think about it. I'm not a fan of moving the props outside of the class (just a subjective thing).

bvaughn:
This discussion thread seems relevant: https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types/issues/38

bvaughn:
I reached out to Olivier (author of the Babel plugin I mentioned above) asking for his recommendations. (He's also a contributor to Material UI so I figured he's thought about this from the library angle a lot already.)

He's going to get back to me with a more detailed answer tomorrow, but in passing he said:
> I believe that the wrap mode is used in the v0.16.0 version of Material-UI right now in production. Also I had to fix one issue, I'm not sure I have released it.

So I'd probably like to wait until I hear his thoughts too and then we can decide the best way to move forward. 😄

bvaughn:
Reached out for some input on Twitter ([here](https://twitter.com/brian_d_vaughn/status/787033766774476801)) and it looks like your suggested approach is right on the money.

bvaughn:
Did a practical experiment to see how much space would be saved by a typical react-virtualized usage.

```js
// Using create-react-app and the following App.js code
class App extends Component {
  render() {
    return (
      <Grid />
    );
  }
}

// Importing Grid this way ~> 252 kb
import { Grid } from 'react-virtualized';

// Importing Grid this way ~> 186 kb
import Grid from 'react-virtualized/dist/commonjs/Grid/Grid';

// propTypes wrapped with NODE_ENV ~> 185 kb
import Grid from 'react-virtualized/dist/commonjs/Grid/Grid';
```

Doing the same as above but using `AutoSizer` + `List` + `Grid` (a more typical sample) the sizes are: 252 kb, 197kb, and 195 kb.

So the size savings aren't huge but they're definitely significant enough to warrant the change.

source/Table/Column.js
```diff
@@ -77,3 +67,14 @@ export default class Column extends Component {
     width: PropTypes.number.isRequired
   }
 }
+
+if (process.env.NODE_ENV !== 'production') {
```
bvaughn: This should be wrapping `propTypes` not `defaultProps`. Those should stay in the class.

bvaughn:
Looks like `Table` and `SortIndicator` got skipped and `Column` got shuffled slightly (`defaultProps` got wrapped instead of `propTypes`).

bvaughn:
I'll go ahead and address those minor issues

bvaughn:
This change just went out as 8.1.1. Gave you thanks in the CHANGELOG. Cheers!

mbrevda:
Belated response:
> let me re-open this PR

Thanks!

> So the size savings aren't huge

Wasn't expecting much, but if every package reduced just a few bytes...

> This should be wrapping propTypes not defaultProps

Oops, sorry

> I'll go ahead and address those minor issues

Thanks!

> This change just went out

First PR! 🎊🎉 Awesome!

Issue #441, opened 2016-10-20
Add support for passing id to components


mnquintana:
* **What's your use case?** Currently, you can pass a custom `className` prop to attach to the root nodes of non-HOC components – it'd be awesome to be able to also pass a custom `id` prop to be able to do the same thing. We're in the midst of refactoring some components that rely on `id`s heavily in their style rules, and unfortunately we can't refactor the CSS selectors for awhile, so we need some way to assign an `id` to virtualized components so they can pick up the necessary styling.
* **What interface do you have in mind?** Just an optional `id` prop on `List`, `Grid`, `Table`, and `Collection`.

I'd be happy to make a PR for this if this sounds 👍  to you.

bvaughn:
Hi @mnquintana,

I understand the motivation behind this feature. Hate to see more pass-through properties added but I understand the situation you're in.

If you're willing to put together a PR, please do.

Note that you would need to update Collection, Table, and Grid. List will already pass thru props to Grid so it will be okay as is.

PR #442, opened 2016-10-20
Add support for id prop to components

mnquintana:
Closes https://github.com/bvaughn/react-virtualized/issues/441

This PR adds a new optional prop, `id`, to `List`, `Grid`, `Table`, and `Collection`, which lets users pass a CSS `id` down to the root nodes of those components for styling purposes. `className`s definitely preferred, but sometimes you have to use the nuclear option and pass an `id` for styling purposes. 😝


bvaughn:
Thanks for contributing, @mnquintana!

8.2.0 has this feature.

PR #443, opened 2016-10-21
Fix minor typo in CellSizeCache usage example

clauderic:
:)


clauderic:
On a side-note, I'm curious about CellSizeCache. Currently, it stores and does lookups based on row/cell indexes. In a scenario where you may have elements that shift indexes but always remain the same size, what are your thoughts on a CellSizeCache implementation that would do lookup based on a cell's `key` property instead? (and, if no key is provided fallback on the index)

bvaughn:
Ha! Look at that. Thanks buddy 😄

PR #446, opened 2016-10-27
Add visibility

mbrevda:
First draft adding a visibility object showing the percentage of the cell that is visible. Not that comfortable yet with the code base 😊, so feel free to point out anything I missed. Thinking of moving `overlap()` somewhere, but not sure where. It's own package? Just another file?

Feedback welcome!

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -665,7 +665,6 @@ export default class Grid extends Component {
         columnSizeAndPositionManager: this._columnSizeAndPositionManager,
         columnStartIndex: this._columnStartIndex,
         columnStopIndex: this._columnStopIndex,
-        height,
         horizontalOffsetAdjustment,
         isScrolling,
         rowSizeAndPositionManager: this._rowSizeAndPositionManager,
@@ -675,8 +674,7 @@ export default class Grid extends Component {
         scrollTop,
         verticalOffsetAdjustment,
         visibleColumnIndices,
-        visibleRowIndices,
-        width
+        visibleRowIndices
       })
     }
   }
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -674,6 +674,8 @@ export default class Grid extends Component {
         scrollLeft,
         scrollTop,
         verticalOffsetAdjustment,
+        visibleColumnIndices,
+        visibleRowIndices,
         width
       })
     }
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -665,6 +665,7 @@ export default class Grid extends Component {
         columnSizeAndPositionManager: this._columnSizeAndPositionManager,
         columnStartIndex: this._columnStartIndex,
         columnStopIndex: this._columnStopIndex,
+        height,
         horizontalOffsetAdjustment,
         isScrolling,
         rowSizeAndPositionManager: this._rowSizeAndPositionManager,
@@ -672,7 +673,8 @@ export default class Grid extends Component {
         rowStopIndex: this._rowStopIndex,
         scrollLeft,
         scrollTop,
-        verticalOffsetAdjustment
+        verticalOffsetAdjustment,
+        width
       })
     }
   }
--- a/source/Grid/defaultCellRangeRenderer.js
+++ b/source/Grid/defaultCellRangeRenderer.js
@@ -1,13 +1,4 @@
 /** @flow */
-const isVisible = (grid, cell) => {
-  return !(
-    cell.left > grid.scrollLeft + grid.width ||
-    cell.top > grid.scrollTop + grid.height ||
-    grid.scrollLeft > cell.left + cell.height ||
-    grid.scrollTop > cell.top + grid.height
-  )
-}
-
 /**
  * Default implementation of cellRangeRenderer used by Grid.
  * This renderer supports cell-caching while the user is scrolling.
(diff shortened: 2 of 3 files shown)


source/Grid/Grid.js
```diff
@@ -672,7 +672,9 @@ export default class Grid extends Component {
         rowStopIndex: this._rowStopIndex,
         scrollLeft,
         scrollTop,
-        verticalOffsetAdjustment
+        verticalOffsetAdjustment,
+        width,
+        height
```
bvaughn: nit: Props should be alpha-sorted.

source/Grid/defaultCellRangeRenderer.js
```diff
@@ -40,7 +97,8 @@ export default function defaultCellRangeRenderer ({
         isScrolling,
         key,
         rowIndex,
-        style
+        style,
+        visibility: findOverlap(gridDimensions, style)
```
bvaughn: IMO this calculation is unnecessarily expensive. I don't think it's important to know _how_ visible a cell is, only whether it's visible (at all) or not. Scroll events fire quickly and often. This is probably the single area where perf matters the most in react-virtualized.

mbrevda: > This is probably the single area where perf matters

Perf is an extremely part of this lib, and I dont want to mess that up!

I think there are two questions here:

1. What does visible mean?
Does it mean the item is fully displayed? Partly displayed? In the use case where a dev may want to render just the "cheaper" parts of the cell, they need to know how much of the cell will be visible, as they will render things differently if 0.03 is visible vs if 0.98 is visible.

2. When is visibility important?
The purpose of visibility is to display a cheaper version of the cell _when it is visibile_. Arguably, this also means _when the cell is at rest_, as if it were scrolling the dev would use `isScrolling` and render an even cheaper version of the cell! Hence, the simplest solution might be not to test or return visibility while a cell is being scrolled:
`visibility: isScrolling ? {x: null, y: null} : findOverlap(gridDimensions, style)`

mbrevda: Thinking about this more, we could even deprecated (but not remove, for backwards compatibility) `isScrolling` in favor of `visibility` and set the state to `scrolling` during scrolling, so like:

```js
visibility: isScrolling ? 'scrolling' : findOverlap(gridDimensions, style)
```

This also removes the ambiguity of the values that would otherwise need to be set for `x` and `y` during scrolling.

bvaughn: > 1. What does visible mean?
Does it mean the item is fully displayed? Partly displayed? In the use case where a dev may want to render just the "cheaper" parts of the cell, they need to know how much of the cell will be visible, as they will render things differently if 0.03 is visible vs if 0.98 is visible.

I disagree that that information is useful. Let's say you have a cell with an image in the top/left corner and text that wraps around it. If the top/left is 0.05 visible (using the code in this PR) you would need to load the image. If the bottom/left, top/right, or bottom/right corners were 0.05 visible you would not need to load it. In other words, the information is more specific than is useful.

I feel pretty strongly that- if we are to add this feature- it should be a simple boolean. `isVisible` would be `true` if any part of the cell shows at all. If it's an over-scanned row then it is, by definition/previous-implementation, not visible. This would make the calculation very performant.

> Thinking about this more, we could even deprecated (but not remove, for backwards compatibility) isScrolling in favor of visibility and set the state to scrolling during scrolling, so like

Perhaps for a version 9, I could get behind a `visibility` enum [visible, overscan, scrolling] but v9 is (hopefully) a ways off. 😉

mbrevda:
> the information is too-specific without adding value

I changed the code to simply detect if the two rectangles intersect. I'm not sure this is  a less expensive job, but I do see what your saying about the old way being unnecessarily elaborate.

Let me know what you think! Presumably this needs to be done for List too, right?

bvaughn:
This is better/simpler :smile: but it could be simpler still.

[`Grid` already has the index information](https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/Grid.js#L640-L677). We just need to pass a little more info into `cellRangeRenderer` (in a backwards-compatible way) so that it can do a range-check:

```js
const isVisible = !(
  columnIndex < visibleColumnStartIndex ||
  columnIndex > visibleColumnStopIndex ||
  rowIndex < visibleRowStartIndex ||
  rowIndex > visibleRowStopIndex
)
```

mbrevda:
Why not just use `visibleColumnIndices` & `visibleRowIndices` and check the cell's index?

bvaughn:
> Do you prefer flatting the objects flattened out to variables? I can do that if you want. (if yes, I'm curious as to why)

I don't understand. Re-word or give an example?

mbrevda:
I passed in objects, you passed only variables. Any reason?

bvaughn:
Oh. My code snippet was just indicated to be an inline expression. I didn't see a real compelling reason to add a function call just to do a quick check like that. 😄 I was assuming we'd be passing in the 4 new visible*index named params to `cellRangeRenderer` and we'd just use them in the inner `for` loop.

mbrevda:
Thanks! Any eta on when this will be published?

bvaughn:
Patience, patience 😉 I noticed that `List` wasn't forwarding the prop to its `rowRenderer` so I'm adding that first.

mbrevda:
Was going to ask about List. Yup, your right. I'll wait patiently  😆

bvaughn:
8.3.0 has this property~ thanks to you in the changelog / tag message. 😄

PR #449, opened 2016-10-31
scrollToAlignment should be pull from props and not from state

coluccini:
ISSUE:  if you set a value to scrollToCell in a Collection component, when you scroll up or down the previous/next cells are not being rendered. This is happening in the Collection example (https://bvaughn.github.io/react-virtualized/#/components/Collection) as well.

ORIGIN: on `componentDidUpdate()` from CollectionView.js you were trying to pull scrollToAlignment value from `this.state` instead from `this.props` so it was always `undefined` making `scrollToAlignment !== prevProps.scrollToAlignment` being always true and `this._updateScrollPositionForScrollToCell()` always triggered.

### DIFF
--- a/source/Collection/CollectionView.js
+++ b/source/Collection/CollectionView.js
@@ -205,8 +205,8 @@ export default class CollectionView extends Component {
   }
 
   componentDidUpdate (prevProps, prevState) {
-    const { height, scrollToCell, width } = this.props
-    const { scrollLeft, scrollPositionChangeReason, scrollToAlignment, scrollTop } = this.state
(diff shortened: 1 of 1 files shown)


bvaughn:
Oh my! What a silly oversight on my part. Nice fix. 😄

bvaughn:
Fix looks great! Thanks!

bvaughn:
Fix will roll as 8.3.1 in a moment or two :)

coluccini:
Once again: it's a real pleasure to use this component. Not only because is an amazing project, but cause you're like one of the most reachable and open developer ever ;) Thanks a lot!

PR #451, opened 2016-11-02
Added getColumnSizeAndPositionManager & getRowSizeAndPositionManager …

olivierandriessen:
There are situations I want to look at size and position information without the standard render callback. This allows for (public) access instead of accessing the underscored (protected) property's.

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -246,6 +246,24 @@ export default class Grid extends Component {
   }
 
   /**
+   * Get the instance of ScalingCellSizeAndPositionManager
+   * for columns to access size and position information
+   * outside the render callback
+   */
+  getColumnSizeAndPositionManager() {
+    return this._columnSizeAndPositionManager
+  }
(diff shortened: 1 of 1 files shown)


bvaughn:
Hm...these helpers are an implementation detail of `Grid` that I feel a bit uncertain about exposing. Adding these getter methods basically implies that I'm willing to commit to not changing the method names and functionality of those inner classes. I'm not sure about that. Need to think it over.

olivierandriessen:
I understand, bit of an edge case perhaps. I'll go for extending the Grid class for now. Thanks ;)

bvaughn:
No problem.

I actually have plans (that I just haven't found the time to try out yet) that would heavily refactor the implementation of the size-and-position-managing logic.

bvaughn:
Going to close this for now. If my idea doesn't pan out and it looks like the helper utils will stay around longer in their current form, I'll consider re-opening it. :)

PR #456, opened 2016-11-08
Update size to height in basic Collection Example

kennethtruong:
height was missing in the basic collection example


bvaughn:
Oh, thank you :)

Issue #459, opened 2016-11-08
Disabled table header shortens table height


Jakehp:
**Example of current functionality / 'Issue'**
https://plnkr.co/sMsCl49eLCjXyxQAv6qn

**Proposition**
When the table header is disabled, make the available row height use the table height and do not reduce it by the header height.

**Result**
When a table's header is disabled, the table height will be what was specified and not shortened by header height.

**Use Case**
The table's height dimension is proportional or equal to parent element, and when header is disabled the table looks short of the expected height.

https://github.com/bvaughn/react-virtualized/pull/460

bvaughn:
If only all issue reports came with repro cases and PRs. Thanks again!

bvaughn:
Version 8.5.1 is being published now with this fix.

PR #460, opened 2016-11-08
When table header is disabled, rows get all available height

Jakehp:
see https://github.com/bvaughn/react-virtualized/issues/459


bvaughn:
Fan-freakin-tastic 😁 Thanks @Jakehp!

PR #462, opened 2016-11-10
Add basename to fix demo links

cesarandreu:
react-router links in the demo are broken when you try opening em in a new tab with cmd+click, middle-mouse button, or right-click. You can reproduce it by opening the demo and trying to open the wizard in a separate tab.

I'm not 100% sure this change fixes it, but it seems to work locally.

To test it out I copied the html entry point (`build/index.html` => `build/react-virtualized/index.html`) and updated the script tag's source (`static/demo.js` => `../static/demo.js`).  Opening links in a new tab seems to work for both of the following urls:

* `http://localhost:3001/#/wizard`
* `http://localhost:3001/react-virtualized/#/wizard`

### DIFF
--- a/source/demo/Application.js
+++ b/source/demo/Application.js
@@ -53,7 +53,7 @@ export default class Application extends Component {
 
   render () {
     return (
-      <HashRouter>
+      <HashRouter basename=''>
         <div className={styles.demo}>
           <div className={styles.headerRow}>
             <div className={styles.logoRow}>


bvaughn:
Hey Cesar. Thanks for the PR. Let's give it a shot and see. :)

bvaughn:
Hm...looking at the [react-router source](https://github.com/ReactTraining/react-router/blob/be62ac357e910e3c9f48dd66c14cb8cdf69891d1/modules/StaticRouter.js#L45-L57), it doesn't seem like this change would have any affect:

```jsx
  // StaticRouter.js
  createHref = (to) => {
    let path = createRouterPath(to, this.props.stringifyQuery)

    if (this.props.basename)
      if (path === '/')
        path = this.props.basename
      else if (path.length >= 2 && path.charAt(0) === '/' && path.charAt(1) === '?')
        path = this.props.basename + path.substring(1)
      else
        path = this.props.basename + path

    return this.props.createHref(path)
  }
```

I'm not able to verify the fix using the steps you mentioned above either. Thoughts?

cesarandreu:
If I had to take a wild guess, is it possible you have an outdated local dependency? I just tried it out on my work laptop and it's working correctly. I used yarn to install the dependencies. Here's a gist to my resulting [`yarn.lock`](https://gist.github.com/cesarandreu/708c071abf6e5bd505d600f341a1466f) file.  Could you try using this `yarn.lock` to install dependencies?

Since there have been many so many changes between the package.json react-router version (`4.0.0-2`) and the latest beta version (`4.0.0-alpha.5`), it wouldn't surprise me if they fixed it already.

If that's the case, this PR wasn't needed, sorry. :P I guess bumping deps and rebuilding the demo site should be enough.

bvaughn:
Yup, alpha v5 fixes this 😄 Thanks

PR #464, opened 2016-11-14
Update CellMeasurer.md

jpollard-cs:
per documents `List` should be passed a prop named `rowRenderer` not `cellRenderer`


bvaughn:
Nice catch, @jpollard-cs. Thanks!

PR #465, opened 2016-11-15
Fix getComputedStyle can return null

neoziro:
getComputedStyle can return null if the element is not displayed, it must be considered.

### DIFF
--- a/source/AutoSizer/AutoSizer.js
+++ b/source/AutoSizer/AutoSizer.js
@@ -105,7 +105,7 @@ export default class AutoSizer extends Component {
     const height = boundingRect.height || 0
     const width = boundingRect.width || 0
 
-    const style = window.getComputedStyle(this._parentNode)
+    const style = window.getComputedStyle(this._parentNode) || {}
     const paddingLeft = parseInt(style.paddingLeft, 10) || 0
     const paddingRight = parseInt(style.paddingRight, 10) || 0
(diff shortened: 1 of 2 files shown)


oliviertassinari:
We see this errors on Firefox. It could be linked to https://bugzilla.mozilla.org/show_bug.cgi?id=548397.

bvaughn:
Hey @neoziro,

Thanks for the PR! What you're describing is a bit of a surprise given that the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) make no mention of a possible `null` return value.

I also made a quick test harness to try it myself- and Chrome, Firefox, and Safari are all returning non-null values for hidden elements.

Can you provide a repro case?

bvaughn:
@oliviertassinari Do you only observe the bug within the context of an `iframe` ?

bvaughn:
Regardless, the fix is trivial so I'm not opposed to it. 😄

bvaughn:
Patch fix available in 8.5.2

oliviertassinari:
@bvaughn Thanks for this quick release. One less js error in production 🎉 
capture d ecran 2016-11-15 a 17 43 56

bvaughn:
👍 Happy to help

neoziro:
Thanks!

PR #466, opened 2016-11-15
Add yarn.lock file

neoziro:


mbrevda:
[It would appear](https://yarnpkg.com/en/docs/managing-dependencies) that using yarn requires it to be manually updated & kept in sync with `package.json`. Meaning if a dependency is updated in `package.json`, it will also have to be updated in yarn.

Is that correct @neoziro? Is that something this project is interested in at this time @bvaughn?

bvaughn:
Right, we would need to keep the 2 in sync.

Adding this file is probably a good idea... I just haven't taken the time to wipe out `node_modules`, do a fresh install (using the lock), and verify that everything works as expected. (Been busy.)

mbrevda:
Is that something that should be run by CI? Having a quick look at the docs, it seems that something like this might work (not sure thought, never used Circle):

```
test:
  post:
    - curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
    - echo "deb http://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
    - rm -rf node_modules
    - yarn install
    - npm test
```

neoziro:
@mbrevda yes it's correct, you have to use yarn to update, add or remove dependencies and it will keep yarn.lock and package.json up to date.

bvaughn:
Well it took me a while but I merged this after testing it 😄 Thanks!

Not sure why Github doesn't show the PR merged. Probably because I rebased it first. Oh well!

Issue #476, opened 2016-11-20
overscanRowCount with a cell cache


arbesfeld:
Hi! Thanks for your work on this great library.

I am trying to use a cell cache along with overscan row count, and seeing extra renders from some children components. Here is an example: https://plnkr.co/edit/ebIVXCgOjM4XAddErRGi?p=preview

If you scroll very small amounts, notice that `rerendering` is logged 30 times on each change. This is far less than the number of times that `rowRenderer` is called.

What I would expect is that the child component here is only re-rendered if necessary.

If I set `overscaneRowCount` to 0 I don't notice this behavior any more.

bvaughn:
I don't really understand what you're saying here.

arbesfeld:
Sorry about that- it's possible that I didn't save the updated example. 

I am using the technique as described here: https://github.com/bvaughn/react-virtualized/issues/448 to cache cells with the goal of avoiding unnecessary re-renders of rows.

However, this cache does not seem to prevent re-renders for "overscanned" rows. In the example above, the render method of `MyClass` is repeatedly called for overscanned rows.

bvaughn:
>  and seeing extra renders from some children components

What do you consider to be an "extra render"?

> If you scroll very small amounts, notice that `rerendering` is logged 30 times on each change. This is far less than the number of times that `rowRenderer` is called.

That's because you're caching, so you avoid re-rendering once a row has already been rendered, no?

> What I would expect is that the child component here is only re-rendered if necessary.

What do you define as "necessary"? By default, react-virtualized does not permanently cache cells. This is because (I think) cells are often dynamic and that would be a confusing default. It does reduce the number of cells that are rendered though, and it temporarily caches during scrolling, all of which means...overall performance is good enough that the lack of a permanent-cache isn't usually a problem.

> If I set `overscaneRowCount` to 0 I don't notice this behavior any more.

I don't think you should ever set `overscanRowCount` to 30 like you're doing in this case. Usually 3-10 is a sufficient amount.

bvaughn:
> However, this cache does not seem to prevent re-renders for "overscanned" rows. In the example above, the render method of MyClass is repeatedly called for overscanned rows.

The Plnkr you linked to is preventing a cell from being rendered more than once. For example, check out this fork (that actually logs the index, so you can tell what's going on): https://plnkr.co/edit/H4btaSacFA5Fa6ft76Tu?p=preview

bvaughn:
[...]

PR #478, opened 2016-11-21
Changed predictive overscan cell shifting to improve perf

bvaughn:
## How does overscan work now?

Background information: [This](https://bvaughn.github.io/connect-tech-2016/#/22/9) is how react-virtualized overscanning works.

For simplicity I'll mention rows only. Columns work the same though. It works like this:

1) Overscanned rows are initially rendered above and below your visible rows (eg overscan of 5 means 5 rows above, 5 rows below)
2) When you start scrolling, overscanned rows are stacked in the direction you're scrolling, to help cut down on visible lag (eg overscan of 5 means 0 rows above, 10 rows below)
3) Once you stop scrolling, react-virtualized returns to its default behavior (eg overscan of 5 means 5 rows above, 5 rows below)

## How is overscan changing?

This PR changes that behavior so that:

1) Overscanned rows are initially rendered below your visible rows (eg overscan of 5 means 5 rows below)
2) If you scroll down, nothing changes. If you scroll up, overscanned rows are shifted (eg 5 above, 0 below).
3) When you stop scrolling, nothing changes- until/unless you change scroll direction.

## Why the change?

React reinstantiates components that have been unmounted and remounted. If components are expensive this leads to a perforamnce hit. RV changing overscan on scroll-stop caused half of the overscanned rows to be reinstantiated due to this (since half of the cells were previously unmounted- _even if their elements_ were cached).

This changeset modifies the shifting behavior. Predictive shifting will remain until/unless scroll direction is reversed, in which case it will be reversed as well. Hopefully this should not have any noticeable impact on UX regarding 'scroll lag' around the edges of the list being scrolled.

It may further help performance as RV was previously doubling down on the number of rows/cols being rendered in the direction being scrolled. (I think this was the wrong initial implementation, but...live and learn.) Now it will render only the exact number specified by the overscan props.

## Test the build

I'd love for you to give this change a test before I release it. You can install the beta release:

```
npm install react-virtualized@8.5.3-beta
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Changelog
 ------------
 
+##### 8.5.3
+Changed overscan rows/cols behavior as described [here](https://github.com/bvaughn/react-virtualized/pull/478).
+This change targets performance improvements only and should have no other noticeable impact.
+
 ##### 8.5.2
 Added guard against potential `null` return value from `getComputedStyle` for hidden elements (see PR #465).
 
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "description": "React components for efficiently rendering large, scrollable lists and tabular data",
   "author": "Brian Vaughn <brian.david.vaughn@gmail.com>",
   "user": "bvaughn",
-  "version": "8.5.3",
+  "version": "8.5.3-beta",
   "homepage": "https://github.com/bvaughn/react-virtualized",
   "main": "dist/commonjs/index.js",
   "module": "dist/es/index.js",
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "description": "React components for efficiently rendering large, scrollable lists and tabular data",
   "author": "Brian Vaughn <brian.david.vaughn@gmail.com>",
   "user": "bvaughn",
-  "version": "8.5.3-beta",
+  "version": "8.5.3",
   "homepage": "https://github.com/bvaughn/react-virtualized",
   "main": "dist/commonjs/index.js",
   "module": "dist/es/index.js",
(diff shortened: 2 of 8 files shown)


benthehenten:
Hey @bvaughn, I'm working with @arbesfeld on the project where #476 came up.

I just tested out this fix in our app and it works as expected. 

Thanks so much for working so quickly to help out here!

bvaughn:
Thanks for the quick feedback! 😁 I'll leave this PR up a bit longer to hopefully give the Housing Eng team and some others a chance to test it out if they're interested. Stay tuned~

ritz078:
how about releasing this as a beta ?

bvaughn:
Great suggestion. Done! 
```
npm install react-virtualized@8.5.3-beta
```

ritz078:
its working as expected 👍

jpollard-cs:
is this still in beta / need help testing?

bvaughn:
I released it a few moments ago as 8.5.3 based on the positive feedback I'd gotten so far (and based on my own testing). But feedback is always welcome!

jpollard-cs:
👍

mlynek14:
Thx for library :) , but... with this change (pull) my component stop working. I am writing list carousel with scrolling based on buttons (scrollToColumn) and animating it with margin left on element with class ReactVirtualized__Grid__innerScrollContainer. Is it possible to put some props, which always keep overflow left and right with the same number? I have realized that this change don't work with scrollToColumn (Grid) param - I put number i.e. 2, 4, 6 (scrolling right) but have no overflow on right side...

bvaughn:
This change should not rbeak any pre-existing components. It _only_ affects overscanning.

If you are using overscanning to display items (sounds like you are) then I think you're misusing it. It isn't intended to work that way.

Rather than making your carousel the size of a single image and relying on overscanning to show the next/previous images- make your carousel wider, to include the next/previous images, and use `scrollToAlignment="center"` (or similar) to manage the position of the "focused" image.

PR #480, opened 2016-11-22
Update CellMeasurer.md

jpollard-cs:
While it may be tempting to share the `List` props with the `CellMeasurer` I think it is best to explicitly call them out rather than spreading the `List` props over the `CellMeasurer` as this could lead to potential conflict by setting a prop you may not have intended to set (e.g. `ref`, `width`, or `height`)

I know this is just an example, but many (most?) people will copy and paste this and it could lead to rather insidious bugs.

### DIFF
--- a/docs/CellMeasurer.md
+++ b/docs/CellMeasurer.md
@@ -174,14 +174,15 @@ For example:
 ```jsx
 import { CellMeasurer, List } from 'react-virtualized';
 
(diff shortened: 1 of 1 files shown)


bvaughn:
Good point. I never _actually_ share props this way. I only did it in the docs because I thought the brevity helped to not obscure the important details. You make a good point though!

Issue #474, opened 2016-11-18
Allow WindowScroller to be used with scroll containers besides window?


andrewbranch:
Wondering if you would consider a PR adding a `scrollElement` (name up for debate) prop to WindowScroller such that it could be used with containers that are neither window nor constrained only to the boundaries of a fully on-screen list. (As an example, imagine an app layout with multiple independently scrolling side-by-side panels.)

At a super quick glance, it looks like instances of `window` in WindowScroller could fairly easily be replaced with a custom element, but I haven't yet tried it, and I wanted to see if it had already been attempted, or if you knew of a reason why it couldn't be done. Or if you would be terribly opposed to the idea since it could make "WindowScroller" a misnomer. 😛

I'm probably going to give it a shot either way, but if it's something you're interested in incorporating, I can let you know how it goes and eventually open a PR.

bvaughn:
> As an example, imagine an app layout with multiple independently scrolling side-by-side panels.

`WindowScroller` is unique because it allows the top-level browser scrollbar to control the position of windowed content. What you're describing sounds to me like a couple of vanilla, side-by-side `List`s. (They can be independently scrolled.) I imagine you wouldn't want the browser's scrollbar to control one, so the only useful function of `WindowScroller` is not applicable. 😁

Maybe I'm misunderstanding you?

bvaughn:
In other words, why not this: https://plnkr.co/edit/A2zDoCWQTvzwW1C53bhS?p=preview

andrewbranch:
Yeah, that's not _quite_ it. What I meant to describe was [a layout like this](https://plnkr.co/edit/ogVK1Sa4UemJ5U8BNAOU?p=preview).

bvaughn:
Oh interesting. I see.

That layout _could_ be accomplished by providing your own `cellRangeRenderer`. (Doing that really gives you a _ton_ of extra control over the innards of the list/grid/whatever.) But it would be more straight forward to do with something like `WindowScroller`. Hm... sure, share a PR. 😄 Happy to review it.

andrewbranch:
Cool! Appreciate the feedback so far. I’ll let you know what happens!

bvaughn:
Released in [8.10.0](https://github.com/bvaughn/react-virtualized/blob/master/CHANGELOG.md#8100)

PR #481, opened 2016-11-22
Add custom element option to WindowScroller

andrewbranch:
In reference to #474 

Adds an optional prop `scrollElement` to WindowScroller which, if provided, attaches scroll handlers to the given element instead of to `window`.

@bvaughn I “tested” this via adding an option to the demo (although maybe you’d like to change the way I did it—added a couple things to Application’s `context`—I’m not familiar with HashRouter, but maybe it can be done through props instead?) and it seems to work nicely. I’m happy to help add tests and docs, but wanted to let you take a look at it as soon as possible, and also get your feedback on the best way to add some tests for this. It almost seems like a big group of the tests could just be run twice, once for `window` and once for some custom element.

There are a couple things I wasn’t sure about, so I’ll start adding a few inline comments to ask about specific things.

### DIFF
--- a/source/WindowScroller/WindowScroller.example.js
+++ b/source/WindowScroller/WindowScroller.example.js
@@ -17,11 +17,17 @@ export default class AutoSizerExample extends Component {
   constructor (props) {
     super(props)
 
+    this.state = {
+      showHeaderText: true
+    }
+
+    this._hideHeader = this._hideHeader.bind(this)
     this._rowRenderer = this._rowRenderer.bind(this)
   }
 
   render () {
     const { list } = this.context
+    const { showHeaderText } = this.state
 
     return (
       <ContentBox>
@@ -31,13 +37,27 @@ export default class AutoSizerExample extends Component {
           docsLink='https://github.com/bvaughn/react-virtualized/blob/master/docs/WindowScroller.md'
         />
 
-        <ContentBoxParagraph>
-          This component decorates <code>List</code>, <code>Table</code>, or any other component
-          and manages the window scroll to scroll through the list
-        </ContentBoxParagraph>
+        {showHeaderText && (
+          <ContentBoxParagraph>
+            This component decorates <code>List</code>, <code>Table</code>, or any other component
+            and manages the window scroll to scroll through the list
+          </ContentBoxParagraph>
+        )}
+
+        {showHeaderText && (
+          <ContentBoxParagraph>
+            <button onClick={this._hideHeader}>
+              Hide header text
+            </button>
+          </ContentBoxParagraph>
+        )}
 
         <div className={styles.WindowScrollerWrapper}>
-          <WindowScroller>
+          <WindowScroller
+            ref={(ref) => {
+              this._windowScroller = ref
+            }}
+          >
             {({ height, isScrolling, scrollTop }) => (
               <AutoSizer disableHeight>
                 {({ width }) => (
@@ -45,9 +65,10 @@ export default class AutoSizerExample extends Component {
                     autoHeight
                     className={styles.List}
                     height={height}
+                    overscanRowCount={2}
                     rowCount={list.size}
                     rowHeight={30}
-                    rowRenderer={({ index, key, style }) => this._rowRenderer({ index, isScrolling, key, style })}
+                    rowRenderer={({ index, isVisible, key, style }) => this._rowRenderer({ index, isScrolling, isVisible, key, style })}
                     scrollTop={scrollTop}
                     width={width}
                   />
@@ -64,11 +85,22 @@ export default class AutoSizerExample extends Component {
     return shallowCompare(this, nextProps, nextState)
   }
 
-  _rowRenderer ({ index, isScrolling, key, style }) {
+  _hideHeader () {
+    const { showHeaderText } = this.state
+
+    this.setState({
+      showHeaderText: !showHeaderText
+    }, () => {
+      this._windowScroller.updatePosition()
+    })
+  }
+
+  _rowRenderer ({ index, isScrolling, isVisible, key, style }) {
     const { list } = this.context
     const row = list.get(index)
     const className = cn(styles.row, {
-      [styles.rowScrolling]: isScrolling
+      [styles.rowScrolling]: isScrolling,
+      isVisible: isVisible
     })
 
     return (
--- a/source/WindowScroller/WindowScroller.example.js
+++ b/source/WindowScroller/WindowScroller.example.js
@@ -9,9 +9,12 @@ import AutoSizer from '../AutoSizer'
 import shallowCompare from 'react-addons-shallow-compare'
 import styles from './WindowScroller.example.css'
 
-export default class AutoSizerExample extends Component {
+export default class WindowScrollerExample extends Component {
   static contextTypes = {
-    list: PropTypes.instanceOf(Immutable.List).isRequired
+    list: PropTypes.instanceOf(Immutable.List).isRequired,
+    customElement: PropTypes.any,
+    isScrollingCustomElement: PropTypes.bool.isRequired,
+    setScrollingCustomElement: PropTypes.func
   }
 
   constructor (props) {
@@ -23,10 +26,15 @@ export default class AutoSizerExample extends Component {
 
     this._hideHeader = this._hideHeader.bind(this)
     this._rowRenderer = this._rowRenderer.bind(this)
+    this.onChangeCustomElementCheckbox = this.onChangeCustomElementCheckbox.bind(this)
+  }
+
+  onChangeCustomElementCheckbox (event) {
+    this.context.setScrollingCustomElement(event.target.checked)
   }
 
   render () {
-    const { list } = this.context
+    const { list, isScrollingCustomElement, customElement } = this.context
     const { showHeaderText } = this.state
 
     return (
@@ -52,11 +60,25 @@ export default class AutoSizerExample extends Component {
           </ContentBoxParagraph>
         )}
 
+        <ContentBoxParagraph>
+          <label className={styles.checkboxLabel}>
+            <input
+              aria-label='Use custom element for scrolling'
+              className={styles.checkbox}
+              type='checkbox'
+              checked={isScrollingCustomElement}
+              onChange={this.onChangeCustomElementCheckbox}
+            />
+            Use custom element for scrolling
+          </label>
+        </ContentBoxParagraph>
+
         <div className={styles.WindowScrollerWrapper}>
           <WindowScroller
             ref={(ref) => {
               this._windowScroller = ref
             }}
+            scrollElement={isScrollingCustomElement ? customElement : null}
           >
             {({ height, isScrolling, scrollTop }) => (
               <AutoSizer disableHeight>
--- a/source/WindowScroller/WindowScroller.js
+++ b/source/WindowScroller/WindowScroller.js
@@ -43,14 +43,18 @@ export default class WindowScroller extends Component {
     this._enablePointerEventsAfterDelayCallback = this._enablePointerEventsAfterDelayCallback.bind(this)
   }
 
-  componentDidMount () {
-    const { height } = this.state
-
+  updatePosition () {
     // Subtract documentElement top to handle edge-case where a user is navigating back (history) from an already-scrolled bage.
     // In this case the body's top position will be a negative number and this element's top will be increased (by that amount).
     this._positionFromTop =
       ReactDOM.findDOMNode(this).getBoundingClientRect().top -
       document.documentElement.getBoundingClientRect().top
+  }
+
+  componentDidMount () {
+    const { height } = this.state
+
+    this.updatePosition()
 
     if (height !== window.innerHeight) {
       this.setState({
@@ -59,6 +63,7 @@ export default class WindowScroller extends Component {
     }
 
     registerScrollListener(this)
+
     window.addEventListener('resize', this._onResizeWindow, false)
   }
 
@@ -91,6 +96,8 @@ export default class WindowScroller extends Component {
 
   _onResizeWindow (event) {
     const { onResize } = this.props
+
+    this.updatePosition()
 
     const height = window.innerHeight || 0
 
--- a/source/WindowScroller/WindowScroller.js
+++ b/source/WindowScroller/WindowScroller.js
@@ -3,6 +3,7 @@ import { Component, PropTypes } from 'react'
 import ReactDOM from 'react-dom'
 import shallowCompare from 'react-addons-shallow-compare'
 import { registerScrollListener, unregisterScrollListener } from './utils/onScroll'
+import { getVerticalScroll, getPositionFromTop, getHeight } from './utils/dimensions'
 
 export default class WindowScroller extends Component {
   static propTypes = {
@@ -12,6 +13,9 @@ export default class WindowScroller extends Component {
      * ({ height, scrollTop }) => PropTypes.element
      */
     children: PropTypes.func.isRequired,
+
+    /** Element to attach scroll event listeners. Defaults to window. */
+    scrollElement: PropTypes.any,
 
     /** Callback to be invoked on-resize: ({ height }) */
     onResize: PropTypes.func.isRequired,
@@ -28,13 +32,13 @@ export default class WindowScroller extends Component {
   constructor (props) {
     super(props)
 
-    const height = typeof window !== 'undefined'
-      ? window.innerHeight
+    const height = typeof this.scrollElement !== 'undefined'
+      ? getHeight(this.scrollElement)
       : 0
 
     this.state = {
-      isScrolling: false,
       height,
+      isScrolling: false,
       scrollTop: 0
     }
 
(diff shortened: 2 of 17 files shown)


source/WindowScroller/WindowScroller.css
```diff
@@ -0,0 +1,3 @@
+.disabled > * {
+  pointer-events: none !important;
```
andrewbranch: I’m using this instead of inline styles because a custom scroll container may have many children that need `pointer-events: none`, whereas scrolling the window only needs to apply that style to `body`. So, I wasn’t sure whether to add `!important` or not. It ought to work pretty similarly to the way it used to (replacing inline styles), and it is still overrideable (higher-specificity selectors with `!important`, or inline styles with `!important`). On the other hand, I hate it when libraries use `!important`.

source/WindowScroller/WindowScroller.example.js
```diff
@@ -9,19 +9,27 @@ import AutoSizer from '../AutoSizer'
 import shallowCompare from 'react-addons-shallow-compare'
 import styles from './WindowScroller.example.css'
 
-export default class AutoSizerExample extends Component {
+export default class WindowScrollerExample extends Component {
   static contextTypes = {
-    list: PropTypes.instanceOf(Immutable.List).isRequired
+    list: PropTypes.instanceOf(Immutable.List).isRequired,
+    customElement: PropTypes.object,
```
andrewbranch: In an earlier commit, I had `PropTypes.instanceOf(Element)`, but the linter complained about that, and I realized that it’s probably for good reason: `Element` won’t be defined on servers, and this is a static property. So, not sure if there’s something better than `object` here? I guess we could make a custom validator function which, when invoked, looks at `window.Element` but gracefully does nothing on server-side render?

source/WindowScroller/utils/dimensions.js
```diff
@@ -0,0 +1,30 @@
...
+    : element.getBoundingClientRect().height
+}
+
+/**
+ * Gets the vertical position of an element within its scroll container.
+ * Elements that have been “scrolled past” return negative values.
+ * Handles edge-case where a user is navigating back (history) from an already-scrolled page.
+ * In this case the body’s top position will be a negative number and this element’s top will be increased (by that amount).
+ */
+export function getPositionFromTop (element, container) {
```
andrewbranch: I was having trouble wrapping my mind around this (previously [here](https://github.com/bvaughn/react-virtualized/pull/481/files#diff-668ee30b9b9d3323692d842bec7a5f55L106)), so extra care in making sure this satisfies the original intention is appreciated.

source/demo/ContentBox.css
```diff
@@ -1,5 +1,5 @@
 .ContentBox {
-  flex: 1;
+  flex: 1 0 auto;
```
andrewbranch: This is obviously not directly related to the source changes, but something to this effect was required to make my addition to the demo work (a ContentBox was shrinking and scrolling, rather than `.Body`). I spot checked the other demo changes and didn't see anything obviously broken. Removing `overflow: auto` has the same effect, if that is preferable to you.

source/WindowScroller/WindowScroller.css
bvaughn: Do you actually mean `.disabled *`?

Hm... either way, `> *` is not a very performant selector so I'm pretty reluctant to use it, and definitely not with such a generic class name as `disabled`. RV classes are named so as to (hopefully) not conflict with any other global CSS names.

Also related- RV has also been shifting away from any CSS-based styles (with the exception of `Table`). Basically anything that's functional is defined via an inline style; anything that's presentational (only `Table`) is defined via CSS to make it easier to override in a variety of ways (without requiring use of `!important`).

source/WindowScroller/WindowScroller.example.js
bvaughn: `any` type is probably okay here. It's not great, but even [the official Flow types use `any` for `ref` type](https://github.com/facebook/flow/blob/7aa23dd3899e02c8d3cd7691890b268db913bd6b/lib/react.js#L125).

source/WindowScroller/utils/dimensions.js
bvaughn: Okay. When I do a real review, I'll read this one carefully.

source/demo/ContentBox.css
bvaughn: No strong feelings here. That's fine. 😄

source/WindowScroller/WindowScroller.css
andrewbranch: If I understand correctly, the intention is to disable pointer events on the contents of whatever element is scrolling, without disabling them on the scrolling element itself. So if the contents can now be _N_ number of nodes, either we use CSS to select all children, or we iterate over all _N_ nodes to add/remove inline styles.

To the initial question, `.disabled > *` and `.disabled *` should have the same effect, and I didn't expect the latter to be more efficient, but that’s fine.

And as far as the name goes, we can definitely change it—since the demo was using CSS Modules, and I'm using CSS Modules in my own stuff, I totally didn't think about it. Another option would be to dynamically generate/insert a stylesheet, but that would be a lot more magic; harder for developers to track down if they have the need to customize behavior.

andrewbranch:
Tests were failing due to missing `autoprefixer-loader`, which is referenced in webpack.config.umd.js, but is deprecated, and you have the correct postcss-loader with autoprefixer config in other webpack config files, so I'm guessing you intend to update the UMD config soon, but I went ahead and added the simplest possible fix so we can see test results.

andrewbranch:
Back to the CSS issue, I see we definitely have to do something different. I totally wasn't thinking and did `import styles from '../WindowScroller.css'` in that util, which doesn’t work in the built output, because the CSS file isn't copied, and not everyone is using Webpack / something that can import CSS. So, not sure what the best way forward is.

source/WindowScroller/utils/onScroll.js
```diff
@@ -1,5 +1,5 @@
+import styles from '../WindowScroller.css'
```
bvaughn: Not using CSS modules in RV proper

bvaughn:
> Tests were failing due to missing autoprefixer-loader, which is referenced in webpack.config.umd.js, but is deprecated, and you have the correct postcss-loader with autoprefixer config in other webpack config files. I'm guessing you intend to update the UMD config soon, but I went ahead and added the simplest possible fix so we can see test results.

> Back to the CSS issue, I see we definitely have to do something different. I totally wasn't thinking and did import styles from '../WindowScroller.css' in that util, which doesn’t work in the built output, because the CSS file isn't copied, and not everyone is using Webpack / something that can import CSS. So, not sure what the best way forward is.

I'm intentionally not using CSS modules within react-virtualized because I don't want to impose that requirement on lib users. Things started failing b'c you are importing a stylesheet to use it this way and the build isn't setup to. (Static style names are used.)

source/WindowScroller/WindowScroller.css
andrewbranch: > The first disables pointer events only for immediate children. The latter disables for all descendants.

Well, yeah, but [`pointer-events` is inherited](http://codepen.io/anon/pen/pNWejN), which I guess is what I was counting on there—analogous to WindowScroller in master counting on elements to inherit it from `document.body`.

>I wonder if it would be more appropriate to approach this by passing a prop (eg disablePointerEvents) to Grid telling it to disable pointer events.

That works, but it _is_ different than the existing effect. WindowScroller in current `master` disables pointer-events on `document.body`, e.g. all the stuff inside the scroll container. My CSS (while obviously not being feasible) accomplishes exactly that for arbitrary scroll containers. What you’re suggesting would narrow the scope of what gets `pointer-events: none` down to just the grid/list, not the whole scroll container. That seems like a reasonable trade-off to me, but I don’t have the same prior knowledge as you about how `pointer-events` effects performance of this whole system. If you’re ok with that trade-off, I can move forward with implementing it.

andrewbranch:
@bvaughn don't want to be a bother, but any thoughts on [my last comments](https://github.com/bvaughn/react-virtualized/pull/481#discussion_r89879941)?

source/WindowScroller/WindowScroller.css
bvaughn: `pointer-evetns` is an Inherited style, yes, but it can be overridden by children. And `Grid` would override, since it sets the property on the `ReactVirtualized__Grid__innerScrollContainer`, which is not the direct child of `WindowScroller` but- at best- a child of its child.

Also using `!important` is kind of a last resort, so I like to avoid it.

bvaughn:
Yup. Just replied.

andrewbranch:
@bvaughn ok so, at least for the moment, I reverted the contentious stuff and only do anything with pointer-events in the default, original case where `window` itself is scrolling. Do you think other changes to the way pointer events works (e.g. passing down as a prop to Grid) is outside the scope of this PR, or does it need to be addressed here?

bvaughn:
> Do you think other changes to the way pointer events works (e.g. passing down as a prop to Grid) is outside the scope of this PR, or does it need to be addressed here?

I don't think it needs to be addressed here.

Unfortunately I noticed a bug while testing this change locally. Steps to reproduce it:

1. Open the `WindowScroller` demo
2. Click the "Use custom element for scrolling" checkbox
3. Scroll down a bit
4. Open another demo (any one)
5. Return to the `WindowScroller` demo

At this point the list will be cut off in the middle, like:
screen shot 2016-12-01 at 8 56 45 pm

Unfortunately it's been a long, rough day here and I don't have the mental energy to dig into this. I don't feel comfortable moving forward with the PR though without it being addressed.

andrewbranch:
No worries, I'll look into it tomorrow! Definitely need to get that fixed. Thanks for reviewing!

bvaughn:
❤️ thank you for understanding.

andrewbranch:
@bvaughn fixed the bug you were seeing—good catch. When WindowScroller’s `componentDidMount` runs, the list has not yet been rendered, so the scroll container, at the time its height was being calculated and cached, had shrunk down to fit its contents. A moment later, the list finished rendering and the scroll container grew to take up all of its available space (the bottom of the window).

Above, I fixed the issue by making sure that scroll container always grows to fill the window, regardless of its contents. However, do you think this is a good case for using `detectElementResize` on the custom scroll containers?

source/WindowScroller/WindowScroller.js
```diff
@@ -43,27 +47,31 @@ export default class WindowScroller extends Component {
...
-      })
+  componentWillReceiveProps (nextProps) {
+    if (nextProps.scrollElement && nextProps.scrollElement !== this.scrollElement) {
+      this._updateDimensions(nextProps.scrollElement)
+      unregisterScrollListener(this, this.scrollElement)
+      registerScrollListener(this, nextProps.scrollElement)
+    } else if (!nextProps.scrollElement && this.scrollElement !== window) {
+      this._updateDimensions(window)
+      unregisterScrollListener(this, this.scrollElement)
+      registerScrollListener(this, window)
```
bvaughn: Am I reading this wrong, or could we reduce the complexity of this a bit?

```js
if (nextProps.scrollElement !== this.scrollElement) {
  this._updateDimensions(nextProps.scrollElement || window)
  unregisterScrollListener(this, this.scrollElement)
  registerScrollListener(this, nextProps.scrollElement || window)
}
```

source/WindowScroller/WindowScroller.js
```diff
@@ -92,7 +113,7 @@ export default class WindowScroller extends Component {
   _onResizeWindow (event) {
     const { onResize } = this.props
 
-    const height = window.innerHeight || 0
+    const height = getHeight(this.scrollElement)
```
bvaughn: We should probably call `this._updateDimensions()` here instead? That will handle cases where a resize affects top position (due to things like text wrapping) and it will also only call `setState` if the `height` has actually changed (eg not just a horizontal resize).

The `onResize` call should also be moved into this helper method for similar reasons (and to avoid duplication).

```js
_onResizeWindow (event) {
  this._updateDimensions()
}
```

source/WindowScroller/WindowScroller.example.js
```diff
@@ -9,19 +9,27 @@ import AutoSizer from '../AutoSizer'
...
+    this.onChangeCustomElementCheckbox = this.onChangeCustomElementCheckbox.bind(this)
+  }
+
+  onChangeCustomElementCheckbox (event) {
+    this.context.setScrollingCustomElement(event.target.checked)
   }
 
   render () {
-    const { list } = this.context
+    const { list, isScrollingCustomElement, customElement } = this.context
```
bvaughn: I know this is probably necessary to test the change within the demo site, but I'm not a fan of the complexity this adds to the `WindowScroller.example` or `Application` components. I wonder if it wouldn't be better to leave this particular functionality _out_ of the demo and just cover it via unit tests instead?

Speaking of which 😁 this change set should be accompanied by a couple of new unit tests so others (including myself) don't break it in the future.

source/WindowScroller/WindowScroller.js
```diff
@@ -83,6 +91,19 @@ export default class WindowScroller extends Component {
...
+  _updateDimensions (scrollElement = this.scrollElement) {
+    const { height } = this.state
+
+    this._positionFromTop = getPositionFromTop(ReactDOM.findDOMNode(this), scrollElement)
+
+    const newHeight = getHeight(scrollElement)
+    if (height !== newHeight) {
+      this.setState({
+        height: newHeight
+      })
```
bvaughn: This is probably the place to call the `onResize` prop too.

bvaughn:
Nice work on fixing that bug.

I resumed the code review and left some more thoughts. I also merged in `master` and there were a few conflicts which I've fixed.

To save you a little time, I've pushed the merged result here: https://github.com/bvaughn/react-virtualized/tree/andrewbranch-window-scroller-custom-element

source/WindowScroller/WindowScroller.example.js
andrewbranch: Totally. I was hoping to get your thoughts on what/how to test. Was wondering if some of the existing WindowScroller tests could be extracted into a function and be called once for each "mode" (scrolling `window` and scrolling `scrollElement`). Does that make sense or should each new test be written out distinctly? Do you want to duplicate some of the basic assertions to make sure WindowScroller as a whole doesn’t mess up with `scrollElement`, or should the new tests only address the specific added functionality?

source/WindowScroller/utils/dimensions.js
```diff
@@ -26,5 +26,5 @@ export function getHeight (element) {
  */
 export function getPositionFromTop (element, container) {
   const containerElement = container === window ? document.documentElement : container
-  return element.getBoundingClientRect().top - containerElement.getBoundingClientRect().top
+  return element.getBoundingClientRect().top + getVerticalScroll(container) - containerElement.getBoundingClientRect().top
```
andrewbranch: This fixes a bug which can be reproduced in the demo, before this commit, by

1. Check the "use custom element for scrolling" box
2. Press Shift + Tab twice so that the "Hide header text" button is focused
3. Scroll down a ways
4. Press the space bar so as to press "Hide header text"
5. Scroll again

I also made sure that the back-button navigation edge case still works.

andrewbranch: @bvaughn Can you check this out, because it ostensibly fixes bugs, but breaks some tests 😭

andrewbranch: @bvaughn Any chance you could look at this sometime soonish? Too long in PRgatory makes me anxious 😅

bvaughn:
I'm so sorry. Promise I'll try to make time to look this weekend. Please ping me again if you haven't heard back from my by Saturday.

bvaughn:
PS Thanks for keeping on top of this!

andrewbranch:
No worries! I didn't want to bother you until after the holidays. Maintaining popular open source things sounds hard 😁

bvaughn:
That's very understanding of you 😄  Let's get this done this weekend. We'll talk more soon!

bvaughn:
Looks like this branch has diverged from master a bit, unfortunately. There are changes showing up in the diffs that aren't related, and some things (see below) committed to master haven't made it into this branch either. My fault for taking so long to review it but it is what it is.

If I checkout this branch as-is and merge in master, 4 tests fail:
* WindowScroller onScroll should trigger callback when window scrolls
* WindowScroller onScroll should update :scrollTop when window is scrolled
* WindowScroller updatePosition should recalculate the offset from the top when the window resizes
* WindowScroller updatePosition should recalculate the offset from the top if called externally

These tests were added in 2e91ec7 which was made after your branch and did not merged in. I'll take a stab at reconciling them this morning because I know we'd both really like this PR to get merged. 😄

bvaughn:
Merged in 5cc2ccb, after conflicts resolved.

Thanks!

bvaughn:
Okay @andrewbranch. This was just released in [8.10.0](https://github.com/bvaughn/react-virtualized/blob/master/CHANGELOG.md#8100). Thanks again for your work!

If you'd be willing, I'd still like to get [some test coverage for this new feature](https://github.com/bvaughn/react-virtualized/pull/481/files#r90760446) maybe as a follow-up PR. If you can't get to it I'll find some time to add them.

Cheers!

andrewbranch:
🎉 Thanks for all the help! I can probably add some tests in the near-ish future. Cheers!

bvaughn:
That would be great. Thanks! :)

PR #483, opened 2016-11-28
Reduce unnecessary render() calls by ensuring ref equality with style objects

MalucoMarinero:
Hey there, this a pull request following up what I was talking about on Twitter the other day.

Thankfully because the cellRangeRenderer is an optional property we can still use this behaviour in our project even if you don't think it's suitable to be merged in, however I think there's little downside to using it. Reusing the style objects through a light object cache means that referential equality is preserved, so 'pure' components that use `shallowCompare` will only execute on creation of elements and transition to/from `isScrolling` while scrolling, rather than every scroll tick.

The test that exercises this functionality can be made to fail by changing the defaultCellRangeRenderer to not use the styleCache:

```
      if (false) {
        style = styleCache[styleKey]
      } else {
        style = {height, width, left, top, position: 'absolute'}
        styleCache[styleKey] = style
      }
```

...which reverts to current behaviour where the style object is regenerated every tick.

### DIFF
--- a/docs/Grid.md
+++ b/docs/Grid.md
@@ -98,6 +98,8 @@ This function accepts the following named parameters:
 ```js
 function cellRangeRenderer ({
   cellCache,                    // Temporary cell cache for use while scrolling
+  styleCache,                   // Cache of style objects sent to cells, to
+                                // ensure referential equality
   cellRenderer,                 // Cell renderer prop supplied to Grid
   columnSizeAndPositionManager, // @see CellSizeAndPositionManager,
   columnStartIndex,             // Index of first column (inclusive) to render
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -208,6 +208,7 @@ export default class Grid extends Component {
   constructor (props, context) {
     super(props, context)
 
+    this._styleCache = {}
     this.state = {
       isScrolling: false,
       scrollDirectionHorizontal: SCROLL_DIRECTION_FORWARD,
@@ -663,6 +664,7 @@ export default class Grid extends Component {
       this._childrenToDisplay = cellRangeRenderer({
         cellCache: this._cellCache,
         cellRenderer,
+        styleCache: this._styleCache,
         columnSizeAndPositionManager: this._columnSizeAndPositionManager,
         columnStartIndex: this._columnStartIndex,
         columnStopIndex: this._columnStopIndex,
--- a/source/Grid/Grid.test.js
+++ b/source/Grid/Grid.test.js
@@ -3,6 +3,7 @@ import React from 'react'
 import { findDOMNode } from 'react-dom'
 import { Simulate } from 'react-addons-test-utils'
 import { render } from '../TestUtils'
+import shallowCompare from 'react-addons-shallow-compare'
 import Grid, { DEFAULT_SCROLLING_RESET_TIME_INTERVAL } from './Grid'
 import { SCROLL_DIRECTION_BACKWARD, SCROLL_DIRECTION_FORWARD } from './utils/getOverscanIndices'
 
@@ -1340,6 +1341,55 @@ describe('Grid', () => {
       cellRendererCalled = false
       render(markup)
       expect(cellRendererCalled).toEqual(false)
+    })
+
+    it('should not re-render grid components if they shallowCompare style', () => {
+      let componentUpdates = 0
+
+      class GridComponent extends React.Component {
+        shouldComponentUpdate (nextProps, nextState) {
+          return shallowCompare(this, nextProps, nextState)
+        }
+
+        componentDidUpdate () {
+          componentUpdates++
+        }
+
+        render () {
+          const { columnIndex, rowIndex, style } = this.props
+          return (
+            <div
+              className='gridItem'
+              style={style}
+            >
+              {`row:${rowIndex}, column:${columnIndex}`}
+            </div>
+          )
+        }
+      }
+
+      function cellRenderer ({ columnIndex, key, rowIndex, style }) {
+        return <GridComponent
+          key={key}
+          columnIndex={columnIndex}
+          rowIndex={rowIndex}
+          style={style}
+        />
+      }
+
+      const props = {
+        cellRenderer,
+        columnWidth: 100,
+        height: 40,
+        rowHeight: 20,
+        scrollTop: 0,
+        width: 100
+      }
+
+      const grid = render(getMarkup(props))
+      simulateScroll({ grid, scrollTop: 1 })
+
+      expect(componentUpdates).toEqual(0)
     })
   })
 })
--- a/source/Grid/defaultCellRangeRenderer.js
+++ b/source/Grid/defaultCellRangeRenderer.js
@@ -5,6 +5,7 @@
  */
 export default function defaultCellRangeRenderer ({
   cellCache,
+  styleCache,
   cellRenderer,
   columnSizeAndPositionManager,
   columnStartIndex,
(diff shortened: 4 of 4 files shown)


bvaughn:
Hey @MalucoMarinero. Thanks for the PR! This looks solid. I was thinking of trying something along these lines after our Twitter chat but I'm glad you beat me to it. 😁

I'd still love to be able to see first-hand the slow renderers you guys were referring to though, so I could see how much of an improvement this makes. Is there _any_ way you could share a repro with me? I'd be happy to sign an NDA if that's required.

MalucoMarinero:
I'm a contractor, not internal so I don't have the rights to do that, but I'll run it by the team. react-virtualized is pretty important to what our web app does -- having moved from Angular, to a React home built thing, to fixed-data-table, and now react-virtualized, all with performance improvements each step of the way -- so we're keen to help/learn anything that'll help us squeeze more performance out of it.

7rulnik:
Hey guys.
I'm not sure that it's about same problem, but we are using react-virtualized and having some perf issues in fifrefox/ie/edge. https://www.kupibilet.ru/search/Y10014DECMOWLED22DEC

clauderic:
Hey @MalucoMarinero, this looks great.

I'll pull it locally to give it a spin, but I believe this is good to be merged 👍

MalucoMarinero:
Cool @clauderic , that said though, I do wonder there is ever a time you'd want to flush the styleCache before unmount (where it happens naturally due to the drop of the reference).

clauderic:
Hmmm, good question. I don't suppose there'd be any harm in manually flushing the styleCache when the Grid unmounts 🤔

MalucoMarinero:
Shouldn't that automatically happen unless you hold onto the reference for the unmounted component I believe? It's not like it's event handlers or anything with additional book keeping required.

I was more thinking with regards to say a long lived Grid with thousands of cells streaming in... needing to trim the cache every now and again, but then maybe that just needs a protection against an upper limit, like when scrolling stops just do a quick check of the key length of styleCache to ensure it's not over an upper limit (and flush it if it goes over)

bvaughn:
Sweet. Thanks for the extra data point @7rulnik! And thanks for giving it a second look as well @clauderic!

> Hmmm, good question. I don't suppose there'd be any harm in manually flushing the styleCache when the Grid unmounts 🤔

I think we should flush/reset the `styleCache` at the same points (in `Grid`) when we flush/reset the `cellCache` to avoid problems. Otherwise a user may change `rowHeight` or `columnWidth` and the style would be stale and cause layout problems.

bvaughn:
To elaborate- we flush the `cellCache` any time `recomputeGridSize` is called (which is necessary). We definitely want to flush the `styleCache` here. We can _probably_ avoid flushing it in `_enablePointerEventsAfterDelayCallback` though I think.

bvaughn:
> I was more thinking with regards to say a long lived Grid with thousands of cells streaming in... needing to trim the cache every now and again, but then maybe that just needs a protection against an upper limit, like when scrolling stops just do a quick check of the key length of styleCache to ensure it's not over an upper limit (and flush it if it goes over)

Flushing the `styleCache` in `_enablePointerEventsAfterDelayCallback` would guard against this.

bvaughn:
> I was more thinking with regards to say a long lived Grid with thousands of cells streaming in... needing to trim the cache every now and again, but then maybe that just needs a protection against an upper limit, like when scrolling stops just do a quick check of the key length of styleCache to ensure it's not over an upper limit (and flush it if it goes over)

Checking the size of (and potentially flushing) the `styleCache` in `_enablePointerEventsAfterDelayCallback` would guard against this.

clauderic:
@bvaughn That was my initial reflex as well, but since the sizeCache key isn't based on index but rather on both position and dimensions, I don't think that would be a concern. The only concern there would potentially be the fact that we might be holding on to style references that are no longer used (at least that's my understanding)

bvaughn:
> ...since the sizeCache key isn't based on index but rather on both position and dimensions, I don't think that would be a concern.

True! My first thought was, "index would be more efficient than all of that string concatenation" but then I realized that's being ridiculous. 😆 You're right, that does avoid the need to flush when recomputed!

clauderic:
> "index would be more efficient than all of that string concatenation" but then I realized that's being ridiculous
We're very, very serious about performance 😂 

I do however wonder whether it would be best to rely on index or position-dimension for the key 🤔

I can see arguments for both. The position-dimension key does have the advantage of not needing to be flushed when cellCache is flushed, but on the flip side, we'll potentially be holding on to a sizeable amount of unused style objects over time. Whereas using the index as `key` would ensure we don't hold on to more style objects then we have cells, _but_ could potentially cause unnecessary repaints (I think)

MalucoMarinero:
Yeah, the style cache key incorporates all the things that could possibly change for that reason, so it shouldn't cause any bugs of incorrect styles being returned. It's not like the optimisation is really in this code, it runs a tiny bit slower here to ensure referential equality down the line, which is where the gains are made back.

That said though, as soon as a row is discarded technically there's no benefit to keeping it's old style object around. Referential equality only gets a performance gain on updates. If there is elsewhere in the code that tracks when cells are popped off the stack, it could potentially clear the style cache then.

bvaughn:
> If there is elsewhere in the code that tracks when cells are popped off the stack, it could potentially clear the style cache then.

If we used index based caching, we could roll through and clear older ranges once scrolling stops- but this (iterating and deleting keys) would have a cost too that would probably outweigh any gains.

Maybe worth looking at a hybrid approach to what `cellCache` does- which is to say reset the `styleCache` after the scroll timeout- but retain styles for the visible rows. If we used the index to cache, instead of the complex key, then we could do something like...

```js
// Inside of _enablePointerEventsAfterDelayCallback
const styleCache = this._styleCache

// Reset the style cache
this._styleCache = {}

// But copy over the visible cells
for (let rowIndex = this._rowStartIndex; rowIndex < this._rowStopIndex; rowIndex++) {
  for (let columnIndex = this._columnStartIndex; columnIndex < this._columnStopIndex; columnIndex++) {
    this._styleCache = styleCache[key]
  }
}
```

Maybe this would be a reasonable compromise?

MalucoMarinero:
Possibly, I mean, reuse of styles matters most when scrolling right, if the component's regular props change there's gonna be a rerender anyway. As long as the flush does retain the visible stuff, and the key is reliable and properly flushes when style objects are invalidated by positioning/sizing changes.

MalucoMarinero:
Anyway, the reason I opted for the string concatenation is because I could be confident nothing would break in the improvement, we'd probably need to try and break whatever we come up with if we try for something a bit cleverer.

bvaughn:
I think we could probably get reasonably solid test coverage of the caching+flushing with 3-4 unit tests. I think that style caching is important...

1) Most importantly when scrolling- to avoid rendering a given row/col more than once. We'd get this with either of the approaches discussed here.
2) Secondarily important (for users who aren't using the `isScrolling` param) when scrolling stops- to avoid a re-render of unchanged cells. We could achieve this with either approach as well.

So then the remaining question is...how much do we care about an ever-growing cache? Since this library is intended to facility display of huge chunks of data, I tend to learn towards caring about this more than not-caring, so...this nudges me towards the pruned-index approach.

bvaughn:
Hey @MalucoMarinero. Friendly ping. Are you still interested in seeing this PR merged? If you've lost interest or more important things have come up, that's totally okay. I don't like long-running branches though so- if that's the case- I just want to know so I can close the PR. 😄

thomasdavis:
I don't believe that the cleaning up of style cache is of the utmost importance but it wouldn't hurt to add it for those who have 10k+ rows etc

Does this PR need a cache clean up system and a few unit tests to be merged in?

bvaughn:
I'm wary of an unbound cache for a library that's performance-focused like this. As it stands, I feel a little uneasy about moving forward with the PR.

I'll pick it up and finish the index/caching-reset changes suggested.

bvaughn:
I don't have permission to push to this branch so I have moved the PR to #506.

I'd love a second set of eyes on it before I merge it. It's late and I'm tired. Any takers? 😄

bvaughn:

PR #484, opened 2016-11-29
Added parameter alias in CellMeasurer to ease List integration

romulof:
This allows to use CellMeasurer with Lists without mocking rowRenderer.

### DIFF
--- a/docs/CellMeasurer.md
+++ b/docs/CellMeasurer.md
@@ -12,7 +12,7 @@ This is an advanced component and has some limitations and performance considera
 ### Prop Types
 | Property | Type | Required? | Description |
 |:---|:---|:---:|:---|
-| cellRenderer | Function | ✓ | Renders a cell given its indices. `({ columnIndex: number, rowIndex: number }): PropTypes.node` |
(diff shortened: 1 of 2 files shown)


bvaughn:
Hm... I have considered doing this in the past but decided not to because I'm afraid this pattern opens the door to potential confusion. For example `columnCount` is still a required prop for `CellMeasurer` even though it doesn't make sense in the context of `List`. If users forget to include that though, things may fail in a less-obvious way.

I don't know. Maybe this PR has more "pro" than "con" to it though. I guess I don't really object. 😄

bvaughn:
Thanks for the contribution. I fixed the tests and merged it. This will go out with the 8.6 release, in the next day or so.

romulof:
I was thinking about creating a mock component called RowMeasurer.

bvaughn:
That would also have been a reasonable route to take. If you like that route better (I do actually) you could submit a follow up PR and I'll be happy to review it

bvaughn:
Disregard. We'll stick with this approach. It seems reasonable.

bvaughn:
8.6.0 has just been released with this change.

PR #495, opened 2016-12-06
UNMET PEER DEPENDENCY error

oceanpad:

│├── UNMET PEER DEPENDENCY react-addons-shallow-compare@^0.14.0 || ^15.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,8 +147,8 @@
     "dom-helpers": "^2.4.0"
   },
   "peerDependencies": {
-    "react": "^0.14.0 || ^15.0.0",
-    "react-addons-shallow-compare": "^0.14.0 || ^15.0.0",
-    "react-dom": "^0.14.0 || ^15.0.0"
+    "react": "^0.14.0 || ^15.3.2",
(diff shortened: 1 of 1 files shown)


mbrevda:
Would it be safe to assume any major version will be compatible? `"react": "^0.14.0 || ^15"`

bvaughn:
Hmm...this change set has nothing to do with the unmet peer dependency issue (mentioned [here](https://github.com/bvaughn/react-virtualized#dependencies)). It just narrows down the range of supported 15.x versions of React, something I don't want to do without a very strong reason.

In case there's confusion, the current peer dependency version range _will_ match against react version 15.3.2 (if that's the one you prefer to use).

For a visual explanation of what this change would do, go [here](https://semver.npmjs.com/), select "react" as the package, and then compare `^0.14.0 || ^15.0.0` to `^0.14.0 || ^15.3.2`. You'll see how much more permissive 15.0.0 is than this change.

Anyway, thanks for the contribution but I can't merge it without causing problems for others who may be using versions in between 15.0.0 and 15.3.2.

PR #499, opened 2016-12-06
MultiGrid feedback

bvaughn:
### What is `MultiGrid`?
1. It's an attempt to stitch together the scroll position of multiple `Grid` components in such a way as to reduce the amount of lag between `Grid`s when scrolling.
1. It's also an attempt reduce the amount of code/complexity required to setup a fixed header/column UI.

Still wondering what the heck am I talking about? ~~Check out the demo [here](https://bvaughn.github.io/react-virtualized/#/components/MultiGrid).~~ **Edit**: The online demo has been overridden by a recent release. It will become available again at some point but in the meanwhile you will have to checkout the branch and run it locally.

### How can I test it?
To test this build, install react-virtualized _next_:
```bash
```

### Known Issues
The following are the only known issues for this PR. (If you have ideas about how to address them- please chip in! **I need your help!**)

- [ ] When scrollbars are visible, the horizontal scrollbar for bottom-right `Grid` is hidden unless you scroll to the very bottom. If I'm not able to find a workaround for this feature I could rewrite `MultiGrid` to use `ScrollSync` internally. This would still provide a simpler API to users wanting basic fixed columns/rows. However it would lose some of the super-smooth vertical scrolling currently offered by the component.

### DIFF
--- a/source/MultiGrid/MultiGrid.js
+++ b/source/MultiGrid/MultiGrid.js
@@ -3,9 +3,6 @@ import React, { Component, PropTypes } from 'react'
 import shallowCompare from 'react-addons-shallow-compare'
 import Grid from '../Grid'
 
-// TODO: Handle misplaced horizontal scrollbar for bottom/right grid
-// TODO: Handle :scrollLeft prop conflict with onScroll :scrollLeft param
-
 /**
  * Renders 1, 2, or 4 Grids depending on configuration.
  * A main (body) Grid will always be rendered.
--- a/source/MultiGrid/MultiGrid.js
+++ b/source/MultiGrid/MultiGrid.js
@@ -3,6 +3,9 @@ import React, { Component, PropTypes } from 'react'
 import shallowCompare from 'react-addons-shallow-compare'
 import Grid from '../Grid'
 
+// TODO: Handle misplaced horizontal scrollbar for bottom/right grid
+// TODO: Handle :scrollLeft prop conflict with onScroll :scrollLeft param
+
 /**
  * Renders 1, 2, or 4 Grids depending on configuration.
  * A main (body) Grid will always be rendered.
@@ -67,8 +70,23 @@ export default class MultiGrid extends Component {
   }
 
   render () {
-    const { height, style, width } = this.props
-    const { scrollLeft, scrollTop } = this.state
+    const {
+      height,
+      onScroll,
+      onSectionRendered,
+      scrollLeft: scrollLeftProp,
+      scrollToColumn,
+      style,
+      width,
+      ...rest
+    } = this.props
+
+    // scrollTop and scrollToRow props are explicitly filtered out and ignored
+
+    const {
+      scrollLeft,
+      scrollTop
+    } = this.state
 
     return (
       <div
@@ -85,8 +103,11 @@ export default class MultiGrid extends Component {
             width
           }}
         >
-          {this._renderTopLeftGrid()}
-          {this._renderTopRightGrid({ scrollLeft })}
+          {this._renderTopLeftGrid(rest)}
+          {this._renderTopRightGrid({
+            ...rest,
+            scrollLeft
+          })}
         </div>
         <div
           onScroll={this._onScrollVertical}
@@ -96,12 +117,23 @@ export default class MultiGrid extends Component {
           style={{
             height: height - this._getTopGridHeight(),
             overflowY: 'auto',
+            overflowX: 'hidden',
             position: 'relative',
             width
           }}
         >
-          {this._renderBottomLeftGrid({ scrollTop })}
-          {this._renderBottomRightGrid({ scrollTop })}
+          {this._renderBottomLeftGrid({
+            ...rest,
+            scrollTop
+          })}
+          {this._renderBottomRightGrid({
+            ...rest,
+            onScroll,
+            onSectionRendered,
+            scrollLeft: scrollLeftProp,
+            scrollToColumn,
+            scrollTop
+          })}
         </div>
       </div>
     )
@@ -121,21 +153,21 @@ export default class MultiGrid extends Component {
   }
 
   _cellRendererBottomRightGrid ({ columnIndex, rowIndex, ...rest }) {
-    const { cellRenderer, fixedRowCount } = this.props
+    const { cellRenderer, fixedColumnCount, fixedRowCount } = this.props
 
     return cellRenderer({
       ...rest,
-      columnIndex: columnIndex + columnIndex,
+      columnIndex: columnIndex + fixedColumnCount,
       rowIndex: rowIndex + fixedRowCount
     })
   }
 
   _cellRendererTopRightGrid ({ columnIndex, ...rest }) {
-    const { cellRenderer } = this.props
+    const { cellRenderer, fixedColumnCount } = this.props
 
     return cellRenderer({
       ...rest,
-      columnIndex: columnIndex + columnIndex
+      columnIndex: columnIndex + fixedColumnCount
     })
   }
 
@@ -219,8 +251,14 @@ export default class MultiGrid extends Component {
     })
   }
 
-  _renderBottomLeftGrid ({ scrollTop }) {
-    const { fixedColumnCount, fixedRowCount, rowCount, styleBottomLeftGrid } = this.props
+  _renderBottomLeftGrid (props) {
+    const {
+      fixedColumnCount,
+      fixedRowCount,
+      rowCount,
+      scrollTop,
+      styleBottomLeftGrid
+    } = props
 
     if (!fixedColumnCount) {
       return null
@@ -228,7 +266,7 @@ export default class MultiGrid extends Component {
 
     return (
       <Grid
-        {...this.props}
+        {...props}
         autoHeight
         cellRenderer={this._cellRendererBottomLeftGrid}
         columnCount={fixedColumnCount}
@@ -240,6 +278,7 @@ export default class MultiGrid extends Component {
           height: 'auto',
           left: 0,
           outline: 0,
+          overflow: 'hidden',
           position: 'absolute',
           ...styleBottomLeftGrid
         }}
@@ -248,12 +287,20 @@ export default class MultiGrid extends Component {
     )
   }
 
-  _renderBottomRightGrid ({ scrollTop }) {
-    const { columnCount, fixedColumnCount, fixedRowCount, rowCount, styleBottomRightGrid } = this.props
+  _renderBottomRightGrid (props) {
+    const {
+      columnCount,
+      fixedColumnCount,
+      fixedRowCount,
+      rowCount,
+      scrollToColumn,
+      scrollTop,
+      styleBottomRightGrid
+    } = props
 
     return (
       <Grid
-        {...this.props}
+        {...props}
         autoHeight
         cellRenderer={this._cellRendererBottomRightGrid}
         columnCount={columnCount - fixedColumnCount}
@@ -262,6 +309,7 @@ export default class MultiGrid extends Component {
         onScroll={this._onScrollHorizontal}
         rowCount={rowCount - fixedRowCount}
         rowHeight={this._rowHeightBottomGrid}
+        scrollToColumn={scrollToColumn - fixedColumnCount}
         scrollTop={scrollTop}
         style={{
           height: 'auto',
@@ -275,8 +323,12 @@ export default class MultiGrid extends Component {
     )
   }
 
-  _renderTopLeftGrid () {
-    const { fixedColumnCount, fixedRowCount, styleTopLeftGrid } = this.props
+  _renderTopLeftGrid (props) {
+    const {
+      fixedColumnCount,
+      fixedRowCount,
+      styleTopLeftGrid
+    } = props
 
     if (!fixedColumnCount || !fixedRowCount) {
       return null
@@ -284,13 +336,14 @@ export default class MultiGrid extends Component {
 
     return (
       <Grid
-        {...this.props}
+        {...props}
         columnCount={fixedColumnCount}
         height={this._getTopGridHeight()}
         rowCount={fixedRowCount}
         style={{
           left: 0,
           outline: 0,
+          overflow: 'hidden',
           position: 'absolute',
           top: 0,
           ...styleTopLeftGrid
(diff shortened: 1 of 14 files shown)


source/MultiGrid/MultiGrid.js
```diff
@@ -0,0 +1,338 @@
...
+      rowIndex: rowIndex + fixedRowCount
+    })
+  }
+
+  _cellRendererBottomRightGrid ({ columnIndex, rowIndex, ...rest }) {
+    const { cellRenderer, fixedRowCount } = this.props
+
+    return cellRenderer({
+      ...rest,
+      columnIndex: columnIndex + columnIndex,
```
bvaughn: Guh, this is a typo. Supposed to be `fixedColumnCount + columnIndex`. I will add to the list above.

PhilippSpo:
When setting scrollbars to always visible (in the system preferences) the (fixed) header row is cut off:
image
We can set the fixed rows to `overflow: hidden` to hide the scrollbars while still being able to set the scroll position.

source/MultiGrid/MultiGrid.js
```diff
@@ -0,0 +1,338 @@
...
+
+    return this._leftGridWidth
+  }
+
+  _getRightGridWidth () {
+    const { width } = this.props
+
+    let leftGridWidth = this._getLeftGridWidth()
+
+    return width - leftGridWidth
```
PhilippSpo: This calculation does not consider the scrollbar width. Thus when you (or your browser) set scrollbars to be always visible the left and the right grid will overflow the total width of the table:
image
Adding scrollbar widths to this calculation is a bad idea, though, because scrollbars can have different dimensions depending on the browser.
We should probably not calculate this width manually, but rather make the right grid take up the remaining space automatically. We could probably use something like Flexbox to do this.

bvaughn: Good catch. Fortunately `dom-helpers` has a helper method for calculating scrollbar size based on the current browser/OS settings. I have to take this into account within `Grid` so it wouldn't be the worst thing to also have to take into account here. Should only affect the vertical scrollbar (the one on the right) I believe.

PhilippSpo: Ah right. I totally forgot about `dom-helpers`. Sounds good to me then 👍

PhilippSpo:
At the moment you don't see the horizontal scrollbars unless you scroll down to the very bottom.
Since you don't see the scrollbars you could think that there are no more columns to scroll to.
Right now I don't see a solution to this problem because decreasing the height of the Grid will break the vertical scroll. 😕

bvaughn:
> At the moment you don't see the horizontal scrollbars unless you scroll down to the very bottom.
Since you don't see the scrollbars you could think that there are no more columns to scroll to.

Oh, _yuck_. Really? ☹️  Not sure of the best way to fix this.

bvaughn:
Version `8.8.0-alpha.2` just published with fixes for all but the most recent horizontal scrollbar issue Philipp pointed out.

sontek:
Looks like `MultiGrid` doesn't work with `CellMeasurer`?  It should snag up the `getColumnWidth` function instead of using it as a value

bvaughn:
@sontek The 2 should work together so far as I know. `MultiGrid` checks if `columnWidth` is a function or a number and uses the right one accordingly. Neither `Grid` nor `MultiGrid` work with a `getColumnWidth` prop; both expect a `columnWidth` prop. You have to do the mapping yourself (see [here](https://github.com/bvaughn/react-virtualized/blob/368a17765f2c1450ea4a02f7004691f8d754e51c/source/CellMeasurer/CellMeasurer.example.js#L108) for an example).

sontek:
Yeah, I understand that.  Here is my code:

```javascript
const list = [
            { name: 'Brian Vaughn', description: 'Software engineer', address: "101 Whatever rd, clare, MI",
                age: 30 },
            { name: 'John Michael Anderson', description: 'Designer', address: "1012 Whatever st, mt.pleasant, MI",
                age: 17},
            { name: 'Joseph Lombrozo', description: 'Football Player', address: "201 No Dr, Salt Lake City, UT",
                age: 24},
            { name: 'Hoover Von Hacker', description: 'Software engineer', address: "301 Yes Dr, Salt Lake City, UT",
                age: 102}
            // And so on...
        ];

        let data = [
            {
                "name": "Name",
                "description": "Description",
                "address": "Address",
                "age": "Age"
            }
        ];

        const columns = [
            "name",
            "description",
            "address",
            "age"
        ];
        for (let i = 0; i < 200; i++) {
            data = [
                ...data,
                ...list
            ];
        }
        const cellRenderer = ({ columnIndex, key, rowIndex, style }) => {
            return (
                <div
                    className={styles.cell}
                    key={key}
                    style={style}
                >
                    {data[rowIndex][columns[columnIndex]]}
                </div>
            );
        };
        const ROW_HEIGHT = 25;


        const renderCellMeasurer = (height, width) => (
            <CellMeasurer
                cellRenderer={cellRenderer}
                columnCount={columns.length}
                height={ROW_HEIGHT}
                rowCount={data.length}
            >
                {({ getColumnWidth }) => (
                    <MultiGrid
                        fixedColumnCount={1}
                        fixedRowCount={1}
                        cellRenderer={cellRenderer}
                        columnWidth={getColumnWidth}
                        columnCount={columns.length}
                        height={height}
                        rowHeight={ROW_HEIGHT}
                        rowCount={data.length}
                        width={width}
                    />
                )}
            </CellMeasurer>
        );

        const renderAutoSizer = () => (
            <AutoSizer>
                {({height, width}) => (
                    renderCellMeasurer(height, width)
                )}
            </AutoSizer>
        );

        return (
            <div className={styles.tableContainer}>
                {renderAutoSizer()}
            </div>
        );
```

If I pass a static columnWidth of `300` it looks like this:

screenshot 2016-12-07 17 03 24

but if I used the code above with `getColumnWidth` I get:

screenshot 2016-12-07 17 05 34

bvaughn:
That doesn't necessarily imply that the function is being used as a value. The function could be returning an invalid size. Have you tried setting a breakpoint in the `getColumnWidth` function to see if it's being measured? There are a few things that can cause `CellMeasurer` to return a height or width of 0 which would also lead to a screenshot like you pasted above.

`MultiGrid` only uses `columnWidth` in 2 places, [here](https://github.com/bvaughn/react-virtualized/blob/368a17765f2c1450ea4a02f7004691f8d754e51c/source/MultiGrid/MultiGrid.js#L174-L180) and [here](https://github.com/bvaughn/react-virtualized/blob/368a17765f2c1450ea4a02f7004691f8d754e51c/source/MultiGrid/MultiGrid.js#L194-L204). In both places it checks if it's a function and uses it as such. Am I missing something?

sontek:
One thing I did notice is when I'm using that function:

```
warning.js:36Warning: _renderNewRootComponent(): Render methods should be a pure function of props and state; triggering nested component updates from render is not allowed. If necessary, trigger nested updates in componentDidUpdate. Check the render method of MultiGrid.
```

That doesn't happen when using static width.

sontek:
I just set a debugger in here:

https://github.com/bvaughn/react-virtualized/blob/368a17765f2c1450ea4a02f7004691f8d754e51c/source/MultiGrid/MultiGrid.js#L178

`index` is an object that looks like this:

```
{
    index: 0
}
```

so the math doesn't work out there.

It is set by this function:
https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/utils/CellSizeAndPositionManager.js#L58

sontek:
The other function has the opposite problem, it does `_ref.index` but we are passing an integer there:

https://github.com/bvaughn/react-virtualized/blob/368a17765f2c1450ea4a02f7004691f8d754e51c/source/MultiGrid/MultiGrid.js#L198

bvaughn:
D'oh. Ok. I'll add that to the list of things to fix before alpha.3

sontek:
I just you a PR to fix that.  My only other feedback right now is having the ability to control what renders after the cells.  Right now if I stretch my table and have a background it looks funky:

screenshot 2016-12-07 20 22 05

Since the extra parts don't run `cellRenderer` they don't get the backgrounds on the rest of the cell.   Ideally I could just autostretch the columns to fill the space have have `columnWidth` be `minWidth` instead.

sontek:
@bvaughn A little more feedback, looks like `ColumnSizer` only expects a child of a `Grid` but not a `MultiGrid`

bvaughn:
Version `8.8.0-alpha.3` was just released as `next`. Fixes the following issues:
* `columnWidth` / `rowHeight` functions are now called correctly by `MultiGrid` (aka `CellMeasurer` should now work correctly)
* `ColumnSize` supports `MultiGrid` as a child (aka `MultiGrid` now defines pass-thru `measureAllRows` and `recomputeGridSize` methods)

To test the new release `npm i react-virtualized@8.8.0-alpha.3`

### I'm also openly asking for help on the last remaining major issue
_When scrollbars are visible, the horizontal scrollbar for bottom-right Grid is hidden unless you scroll to the very bottom._

jquense:
I had some ideas about the scrollbar issue but haven't had any time to try yet :(

bvaughn:
Would love to hear them if you find time to share!

felixguertler:
Not related to the scrollbar but since this issue is called "MultiGrid feedback":

I've noticed that scrolling performance for the Multi Grid demo is not as good as the other demos. 

I'm testing in Chrome Version 55.0.2883.87.

It seems to be related to the columns shown at once. 

On a wide screen with 26 columns shown at once I get around 12-25 FPS. 
When making the window smaller to 12 columns, I get around 40-50 FPS.

At the same time, the scrollsync demo does 55-60 FPS even with 25 columns shown at once.

I'd be happy to do more testing!

bvaughn:
Hm. That's unexpected, @felixguertler. I would welcome more testing/ideas/input from you if you have any, yes!

jquense:
so unfortunately i don't have a lot of time to look into this right now :/ but i did wanna throw together a little codepen with an idea for dealing with the scrollbar issue. I have no idea if this is gonna scale, but here ya go: http://codepen.io/anon/pen/gLBmON I've seen this approach (simpler) before and it worked ok. The idea being that you fake the overflow which we can do since we know the scrollHeights for everything...there is a bit of reimplementing scrolling with the wheel event, but i don't know that it's too bad :P

timothyjlaurent:
I'm not seeing anything on the demo page.

bvaughn:
@timothyjlaurent Demo change is now reflecting the recent 8.8.0 release (with style caching). Unfortunately the `MultiGrid` demo is temporarily only available by checking out this branch and running it locally. Sorry!

timothyjlaurent:
NP. I'll try it out if I get a chance.

timothyjlaurent:
It feels really good, can't wait til it's ready to use. No time to do extensive testing but it looks great!!

bvaughn:
Thanks for the input, Tim!

vyorkin:
I've just noticed another issue if you first scroll horizontally and then vertically:
example

leonprou:
We would love to use the feature!  
Had myself some problems with overflows (they're can be really problematic). Can't think for any simple solution for this

SpencerCarstens:
Are we going to get a `MultiTable` version for the flexbox and sorting? 💯  Haha

bvaughn:
@SpencerCarstens: Since `Table` doesn't support horizontal scrolling, I don't think a `MultiTable` component makes sense. 😅  If you want no-scrolling, use `Table`. If you want scrolling and fixed columns, use `MultiGrid`. (At least once I resolve the horizontal scrollbar issue.)

SpencerCarstens:
@bvaughn Sorry, was a bad joke. :)

bvaughn:
No worries.  I can't tell sometimes. 🤣

leonprou:
@bvaughn When MultiGrid will be finished, do you think that will be a good idea to rewrite Table to use MultiGrid instead of Grid? I think that much of the code there like rendering header can be replaced with MultiGrid features.

bvaughn:
`Table` serves a different purpose, though they have some similarities. One of the important aspects of `Grid` and `MultiGrid` is that they support horizontal windowing (and scrolling) for when you need to render a lot of data on that axis. `Table` is about resizing columns to fit the available width. I think both are useful and which one you use will depend on the situation.

I'm currently thinking that `MultiGrid` is going to need to use `ScrollSync` internally, unfortunately. I can't think of a solution to the horizontal scroll bar issue mentioned in the PR description. ☹️

source/MultiGrid/MultiGrid.js
bvaughn: FYI I'm not actually able to reproduce the screenshot you've attached with scrollbars always visible. If I scroll all the way to the right it looks as I would expect:

screen shot 2017-01-02 at 1 49 39 pm

screen shot 2017-01-02 at 1 50 31 pm

agirton:
@bvaughn I tested out the multi-grid and one nice feature would be to allow for an user to use their own header renderer. In the application I'm working the header has some special data/styling so having control over the way the header is rendered would be nice.

bvaughn:
@agirton This should already be possible, no? The header `Grid` gets [all of the same props](https://github.com/bvaughn/react-virtualized/blob/multi-grid/source/MultiGrid/MultiGrid.js#L398-L434) passed to the main `Grid` including your `cellRenderer`. So you should be able to determine if you're in the header based on the `rowIndex` and then you can render however you'd like.

agirton:
@bvaughn that's a good point I could just use the `rowIndex`. It would be nice not have to check the `rowIndex` though. On the other hand it's possible it adds more complexity because once the `cellHeaderRenderer` is added, then another user may want the `cellLeftColumnRenderer` and the API complexity grows. I'll just use the `rowIndex`. Thanks!

bvaughn:
Agreed! Splitting props for all 4 `Grid`s would add a ton of complexity. Thanks for understanding. 😄 

I'm hoping to release a version of this, potentially this weekend, although it will likely have to be based on `ScrollSync` because I haven't been able to think of a solution to the horizontal scrollbar issue mentioned in the description. 😦 

If anyone has any last minute ideas I would welcome them.

bvaughn:
Okay! Perhaps we can improve the internal design in the future, if someone can help figure out that horizontal scrollbar issue- but I'm going to merge the `ScrollSync` implementation so people can at least take advantage of the reduced boilerplate for fixed cols/rows. 😁  Thanks all for the feedback and input!

NathanBWaters:
@bvaughn just want to say how much `MultiGrid` rocks.  Thanks for all the great work!

bvaughn:
Usually I get bug reports or requests. How nice of you to take the time to say this. Thanks @NathanBWaters 😄

PR #500, opened 2016-12-07
Arrowkeystepper align

emilebres:
Hey,

First thanks for building and maintaining this great library.

I have recently begun using ArrowKeyStepper to add keyboard navigation to a pivot grid component I am developing using react-virtualized.
However, I did not found a combination of `ArrowKeyStepper` `mode` and `Grid` `scrollToAlignement` that gives the behaviour I want.

### Behaviour
- on key press, the grid moves to the next column/row in the direction
- the borders of the top left cell are aligned with the borders of the grid

### Existing
- `mode:  cells` and `scrollToAlignment: start`
Works almost as wished but `ArrowKeyStepper` does not follow the scroll.
- `mode:  edges` and `scrollToAlignment: start`
The grid jumps to the next cell not yet visible.
- `mode:  edges` and `scrollToAlignment: auto`
Follows the scroll but the align the borders of the bottom right cell instead. This causes problems when the cells are of unequal sizes.

### Proposition
I have been able to get the desired behaviour by setting `scrollToAlignment` to `start` and creating a new `mode` in `ArrowKeyStepper` called `align:top-left`. For completion, (and to help support right-to-left setups) I also added a `align:top-right` `mode` (to be used with `scrollToAlignment` `end`. You can see the commits for the changes but it's pretty straight-forward replacing some `columnStartIndex` with `columStopIndex` and vice-versa.

I will be happy to help if you have any questions/remarks.
Thanks

### DIFF
--- a/source/ArrowKeyStepper/ArrowKeyStepper.example.js
+++ b/source/ArrowKeyStepper/ArrowKeyStepper.example.js
@@ -74,7 +74,7 @@ export default class ArrowKeyStepperExample extends Component {
               onChange={event => event.target.checked && this.setState({ mode: 'align:top-left' })}
               value='align:top-left'
             />
-            align:top-left
+            align:top-left (<code>Grid</code> prop <code>SetToAlignment</code> is set to <code>start</code>)
           </label>
           <label>
             <input
@@ -85,7 +85,7 @@ export default class ArrowKeyStepperExample extends Component {
               onChange={event => event.target.checked && this.setState({ mode: 'align:bottom-right' })}
               value='align:bottom-right'
             />
-            align:bottom-right
+            align:bottom-right (<code>Grid</code> prop <code>SetToAlignment</code> is set to <code>end</code>)
           </label>
 
         </ContentBoxParagraph>
--- a/source/ArrowKeyStepper/ArrowKeyStepper.example.js
+++ b/source/ArrowKeyStepper/ArrowKeyStepper.example.js
@@ -78,14 +78,14 @@ export default class ArrowKeyStepperExample extends Component {
           </label>
           <label>
             <input
-              aria-label='Set mode equal to "align:top-right"'
-              checked={mode === 'align:top-right'}
+              aria-label='Set mode equal to "align:bottom-right"'
+              checked={mode === 'align:bottom-right'}
               className={styles.Radio}
               type='radio'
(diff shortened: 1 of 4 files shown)


bvaughn:
Hey @emilebres! This looks great at a glance. I'll do the PR after work today. Just wanted to leave a quick "thanks"! :)

bvaughn:
Just getting around to reviewing this PR and I'm kind of confused by it.

It seems to be a little buggy, at least if I'm understanding your intentions correctly. Here's a video showing what I'm talking about: https://vid.me/Ij1s

The change to the CSS include also breaks the `Table` demo page. (It should be changed back to the previous path.)

emilebres:
So, I reverted the path for the css file. The problem was that I had not `npm build` the project before doing `npm start` so the `styles.css` file was not yet created under the `react-virtualized` folder. Sorry about that.

I agree that the demo looks pretty ugly with `scrollToAlignment` set to `auto`. Basically `top:align-left` only works properly when `scrollToAlignment` is set to `start`. So I corrected to demo to have this behaviour.

Furthermore, `align:top-right` did not work properly with `scrollToAlignment` set to `end`. Because `end` aligns to the bottom-right of the grid, not to the top-right. So I changed `align:top-right` to `align:bottom-right`. It works as wished but the scroll bars cover the bottom-right cell partially.

As for the reason for introducing these changes, the `align:top-left` `mode` with `scrollToAlignment` to `start` allows to navigate the grid with both the mouse and the keyboard while keeping the top-left cell borders aligned with the borders of the grid which seemed the most natural to me. I am aware that this is becoming pretty complicated for something not directly linked to the core of `react-virtualized`. If you feel, it's better, I have no problem with you closing this pull request. I would just maintain my own version of `ArrowKeyStepper` :-)

Thanks for the review.

bvaughn:
No problem at all WRT the CSS file. The current setup isn't super intuitive (since you have to just _know_ to build the CSS file ahead of time). That's my fault. I should clean it up.

And no problem reviewing the PR either. I really appreciate the contribution! 😁

I think this may be a bit too specific to merge into the base library though. It's just a gut feeling. Your code looks fine! I'm just not sure the use-case is generically useful and as such I'm wary of committing to support and maintain it going forward.

Since `ArrowKeyStepper` is pretty tiny, I hope you won't mind just maintaining this feature in a forked component. 😄

PR #503, opened 2016-12-08
Pass correct datastructure to getColumnWidth

sontek:
This makes it so `MultiGrid` doesn't error out and renders appropriately when using a `CellMeasurer`.

I still see this error:

```
Warning: _renderNewRootComponent(): Render methods should be a pure function of props and state; triggering nested component updates from render is not allowed. If necessary, trigger nested updates in componentDidUpdate. Check the render method of MultiGrid.
```

but I'm not sure whats causing that.  It only happens when using CellMeasurer though.

### DIFF
--- a/source/MultiGrid/MultiGrid.js
+++ b/source/MultiGrid/MultiGrid.js
@@ -175,7 +175,7 @@ export default class MultiGrid extends Component {
     const { fixedColumnCount, columnWidth } = this.props
 
     return columnWidth instanceof Function
-      ? columnWidth(index + fixedColumnCount)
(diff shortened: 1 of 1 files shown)


sontek:
@bvaughn Here is the fix for that `getColumnWidth` issue we were discussing.

bvaughn:
Thanks a ton for the PR!

We need to make the same change for `rowHeight` too so I'm just going to patch them both locally (and updated the `MultiGrid.example` to make sure this functionality is tested). Will be doing alpha-3 release this morning _hopefully_.

PR #505, opened 2016-12-09
Fix a bug where the columnWidth callback was being invoked with an in…

rcoedo:
Hello Brian

I'm testing the alpha version and I've just encountered this bug.

I've been developing something quite close to this component but using ScrollSync instead. This solution is way more performant, good job!

### DIFF
--- a/source/MultiGrid/MultiGrid.js
+++ b/source/MultiGrid/MultiGrid.js
@@ -171,11 +171,11 @@ export default class MultiGrid extends Component {
     })
   }
 
-  _columnWidthRightGrid (index) {
+  _columnWidthRightGrid ({ index }) {
     const { fixedColumnCount, columnWidth } = this.props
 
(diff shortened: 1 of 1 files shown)


bvaughn:
Thanks @rcoedo. This PR looks like a dup of #503. Which is fine, both are good. I just haven't had the time to merge either and make an alpha 3. I will this weekend.

bvaughn:
Thanks a ton for the PR!

We need to make the same change for `rowHeight` too so I'm just going to patch them both locally (and updated the `MultiGrid.example` to make sure this functionality is tested). Will be doing alpha-3 release this morning _hopefully_.

rcoedo:
Oops, I didn't notice about #503!

I'll be testing the alpha-3 this week, I'll let you know if I find something else. Thanks! 🍻

PR #506, opened 2016-12-13
Reduce unnecessary render() calls by ensuring ref equality with style objects

bvaughn:
Continuation of PR #483.

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -208,7 +208,6 @@ export default class Grid extends Component {
   constructor (props, context) {
     super(props, context)
 
-    this._styleCache = {}
     this.state = {
       isScrolling: false,
       scrollDirectionHorizontal: SCROLL_DIRECTION_FORWARD,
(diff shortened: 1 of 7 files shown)


source/Grid/Grid.js
```diff
@@ -701,8 +704,22 @@ export default class Grid extends Component {
   _enablePointerEventsAfterDelayCallback () {
```
zackargyle: Might be worth renaming this function now.

zackargyle:
Looks awesome. Curious if using an LRU cache would perform better. You wouldn't have to worry about clearing the cache on scroll end and you'd always have the viewable cell styles in cache. Either way, solid optimization!

MalucoMarinero:
Sorry, haven't lost interest, more like having a week off looking after the family, but either way you guys are probably better situated to make your decisions on how to move this forward, you don't need/want any input from me yeah?

bvaughn:
Input is valuable, definitely. I'd welcome any you have! But family stuff is more important than coding so no worries at all.

source/Grid/Grid.js
bvaughn: Agreed!

bvaughn:

romulof:
I think this has broken my code.

I have List with variable-height elements. At start it renders a small "loading" element, then it renders the actual element. Even manually controlling CellMeasurer cache, it still renders the actual element with the same small height as the "loading" element.

bvaughn:
Size info has always been cached. If size changes/invalidates, you need to inform Grid via the public API method telling it that a given cell's size is invalid.

I can't diagnose a brief description of a bug like this though. You should post to Stack Overflow.

PR #507, opened 2016-12-13
Avoid duplicated style declarations for detectElementResize

edulan:
## Issue
When using multiple instances of AutoSizer (or mounting and unmounting multiple times) a new style tag is being added to the document head. This is due to `detectElementResize` vendor library that forces creation  every time it's called.

### Solution 1
Move `stylesCreated` variable outside of function scope. This way this variable would be set just once independently of how many times function is called.

### Solution 2
Identify style tag with an `id` attribute so we don't duplicate it if it already exists.

### Solution 3
Extract style declaration to RV `styles.css` file.

## Current approach
I've opted for using an id as it ensures style tag is created even if it has been previously removed somehow.

## Screenshot
Here you can see it in action:

detectElementResize

### DIFF
--- a/source/vendor/detectElementResize.js
+++ b/source/vendor/detectElementResize.js
@@ -21,7 +21,6 @@ export default function createDetectElementResize () {
   }
 
   var attachEvent = typeof document !== 'undefined' && document.attachEvent;
-  var stylesCreated = false;
 
   if (!attachEvent) {
     var requestFrame = (function(){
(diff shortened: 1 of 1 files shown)


bvaughn:
Fantastic pull request. So much detail. 😁  Thanks!

edulan:
You are very welcome :)

bvaughn:

edulan:
Awesome @bvaughn :clap: :clap:

PR #515, opened 2016-12-26
New windowing algorithm for Grid

bvaughn:
This branch explores a totally new layout approach for windowing. Goals include:

* Avoid measuring cells until absolutely necessary (for better performance with dynamic content)
* Better support for reverse lists and/or RTL lists
* Enable async measurements (eg accordions, late-loaded images)
* More natural scrolling for super-large (compressed) lists

Secondary goals and considerations:

* Controlled vs uncontrolled scroll offset props
* An approach that might better fit with some of the unique constraints of React Native


akraines:
Has there been any progress on this? 
I'm looking forward for this update very much as I'm using a grid to render dynamic content where the rows are noticeably expensive to render, and depending on the user-selected options the rows content is transformed in different ways dynamically - meaning that our use case of RV will benefit greatly from this update.
Currently I'm using multiple CellMeasuring Caches (depending on the selected options) to maximise the rendering efficiency to the extent that I can, but ultimatley not having to measure every row until the target row that the user jumped to will be a game changer for my usage.
Thanks a lot for your great work on this project.

bvaughn:
Hi @akraines,

No. This has stalled. Version 9 (PR #577) may bring you performance improvements for `CellMeasurer` but will not substantially change the way `Grid` works.

akraines:
Thanks, I just got around to updating and it worked like a charm! Keep up the good work!

TrySound:
@bvaughn Is this still a thing? I may dig in on holidays.

bvaughn:
Yeah, it's still something I have interest in but haven't had the time to finish making it work correctly. I'd be thrilled if you forked it and spent some time digging in.

dcolens:
I'm interested in the work being done on this branch, it looks like grid.refactor.js is not backward compatible with grid.js, is the plan to add the missing functionalities and make it backward compatible or modify HOCs ? Is there some sort of plan/roadmap/architecture ?

bvaughn:
There is no active plan, although help would be welcome finishing this effort out.

When I started it, I was trying to get the basics working first- at which point I would decide which additional features were needed.

PR #519, opened 2016-12-30
Add loose-envify transform to package.json

chrisvasz:
We are building our application using Browserify. When including react-virtualized through Browserify, the `process.env.NODE_ENV` checks are not processed, causing Browserify to include a `process` shim in the bundle. This bloats the bundle size and reduces opportunity for minification since `"development" !== "production"` checks are no longer removable.

This PR tells Browserify to apply the `loose-envify` transform when react-virtualized is included in a Browserify bundle. After applying this change, `process.env.NODE_ENV` is correctly replaced during our build process.

Other libraries, such as Redux, use this method to provide `process.env` replacement too -- see [here](https://github.com/reactjs/redux/commit/a9d7c0ef09b1d7970e774ab09a55aff44048cdb5) for the relevant commit on Redux.

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,8 +144,7 @@
   "dependencies": {
     "babel-runtime": "^6.11.6",
     "classnames": "^2.2.3",
-    "dom-helpers": "^2.4.0",
-    "loose-envify": "^1.1.0"
+    "dom-helpers": "^2.4.0"
   },
   "peerDependencies": {
     "react": "^0.14.0 || ^15.0.0",
--- a/package.json
+++ b/package.json
@@ -144,11 +144,17 @@
   "dependencies": {
     "babel-runtime": "^6.11.6",
     "classnames": "^2.2.3",
-    "dom-helpers": "^2.4.0"
+    "dom-helpers": "^2.4.0",
+    "loose-envify": "^1.1.0"
   },
   "peerDependencies": {
     "react": "^0.14.0 || ^15.0.0",
     "react-addons-shallow-compare": "^0.14.0 || ^15.0.0",
     "react-dom": "^0.14.0 || ^15.0.0"
+  },
+  "browserify": {
+    "transform": [
+      "loose-envify"
+    ]
   }
 }


bvaughn:
Hi @chrisvasz,

Thanks for the PR. 😄

I don't think `loose-envify` should not be added as a dependency for react-virtualized since (a) it isn't a runtime dependency and (b) not everyone uses Browserify. I _think_ this `package.json` entry is just for projects that use Browserify to bundle JS, no?

Disclaimer: I'm not very familiar with Browserify.

bvaughn:
I'm going to commit the new `browserify` block by itself. If anything, it seems like `loose-envify` should be a `devDependencies` in an _external_ project that actually uses Browserify. (`react-virtualized` uses Babel+Webpack so it doesn't seem appropriate to add the `devDependencies` here.) Please let me know if I've misunderstood anything. 😄

chrisvasz:
Thanks! This will save us about 7kb in production because of the propTypes removal and because Browserify will no longer include a `process` shim. Woohoo!

I do think `loose-envify` needs to be a dependency though. In my case, without the `dependencies` entry, this *happens* to work because I use other modules that require `loose-envify` and Browserify can find it in `node_modules`. If I remove the other projects that depend on `loose-envify`, my build breaks because Browserify can't find the `loose-envify` module.

From what I can tell, this is actually the reason Redux specifies `loose-envify` as a dependency (not a devDependency) also. Check out [this conversation](https://github.com/reactjs/redux/issues/1304) and [this followup PR](https://github.com/reactjs/redux/pull/1306).

bvaughn:
Good call @chrisvasz. If you'd be willing to make a PR similar to reactjs/redux/pull/1306 I'll accept it. Thanks again!

chrisvasz:
All over it. Thanks!

bvaughn:
Circling back to this because I found myself, again, wondering why we had to list `loose-envify` as a dependency instead of a dev dependency.

PR #522, opened 2017-01-02
Update dom-helpers

danez:
Currently uses the older version, which leads to dom-helpers being included twice, when bundling react-virtualized with webpack and having dom-helpers 2.x and 3.x used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,7 +144,7 @@
   "dependencies": {
     "babel-runtime": "^6.11.6",
     "classnames": "^2.2.3",
-    "dom-helpers": "^2.4.0"
+    "dom-helpers": "^3.2.0"
   },
   "peerDependencies": {
     "react": "^0.14.0 || ^15.0.0",
--- a/package.json
+++ b/package.json
@@ -144,7 +144,7 @@
   "dependencies": {
     "babel-runtime": "^6.11.6",
     "classnames": "^2.2.3",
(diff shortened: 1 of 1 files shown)


bvaughn:
Hi @danez,

Thank you for creating this PR.

Jumping a dependency from 2.x to 3.x is a questionable change without a major release.

I don't see a changelog on the `dom-helpers` lib unfortunately so I'm not sure if it would be safe enough to just relax the dependency to allow either 2.x or 3.x (eg `"^2.4.0 || ^3.0.0"`)

@jquense would you have any suggestion by chance?

jquense:
should be safe to upgrade. the "breaking" change in 2 -> 3 was around the import structure when importing all of `dom-helpers`, since RV just cherry-picks scrollbarSize , it shouldn't affect ya'll

bvaughn:
Excellent. Thanks @jquense!

@danez let's go with a pattern that supports either version (eg `"^2.4.0 || ^3.0.0"`) and leave it up to library users which they use. I don't want to yank the rug out from under anyone who may be relying on the 2.x import structure. It sounds like RV is safe to work with either though. 😁

danez:
Okay updated :)

bvaughn:
Fantastic. Thanks to both of your for your help with this. 😄

I'll try to get a release out sometime within the next day or two.

PR #523, opened 2017-01-03
Add loose-envify dep for Browserify users

chrisvasz:
As per the discussion in #519.


bvaughn:
Thanks again @chrisvasz 😄

PR #527, opened 2017-01-06
Fix bad styles being cached in massive lists

nickclaw:
In lists large enough to trigger scaling, the cached styles can be invalid. Causing weird banding effects and overlapping as you scroll.

bands

### DIFF
--- a/source/Grid/Grid.test.js
+++ b/source/Grid/Grid.test.js
@@ -6,6 +6,7 @@ import { render } from '../TestUtils'
 import shallowCompare from 'react-addons-shallow-compare'
 import Grid, { DEFAULT_SCROLLING_RESET_TIME_INTERVAL } from './Grid'
 import { SCROLL_DIRECTION_BACKWARD, SCROLL_DIRECTION_FORWARD } from './utils/getOverscanIndices'
+import { DEFAULT_MAX_SCROLL_SIZE } from './utils/ScalingCellSizeAndPositionManager'
 
 const DEFAULT_COLUMN_WIDTH = 50
 const DEFAULT_HEIGHT = 100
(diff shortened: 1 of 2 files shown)


source/Grid/Grid.test.js
```diff
@@ -1507,4 +1508,33 @@ describe('Grid', () => {
...
+      cellRendererCalls.push(props)
+    }
+
+    const grid = render(getMarkup({
+      cellRenderer,
+      width: 100,
+      height: 100,
+      rowHeight: 100,
+      columnWidth: 100,
+      rowCount: DEFAULT_MAX_SCROLL_SIZE * 2 / 100, // lots of offset
```
bvaughn: This seems like an arbitrary/odd `rowCount` value. What's the logic?

nickclaw: Well, the height of the inner container needs to be over DEFAULT_MAX_SCROLL_SIZE to trigger this. Doing `DEFAULT_MAX_SCROLL_SIZE / rowHeight` will get us exactly enough rows to fill  the container -- so I figured I might as well double it to be sure :P

I'm happy to add comments explaining it, or if you have a better suggestion for the test use that.

bvaughn: Ah, I see. I think the order of operations threw me a bit. (It's late here.) It's fine as is. :)

nickclaw:
I can tell haha, writing the ScalingCellSizeAndPositionManager alone looked like enough to make one go crazy

bvaughn:
I have hopes that the new layout algorithm in PR #515 might totally change the way scaling is done- making things like offset-adjustments and `ScalingCellSizeAndPositionManager` unnecessary. But it's still kind of an experiment. It might not pan out.

PR #530, opened 2017-01-06
Avoid double forceUpdate

nickclaw:
I realized that recomputeGridSize and recomputeRowHeights already call force update, causing this nice little performance hit.

screen shot 2017-01-06 at 2 52 50 pm

The diff on the tests looks a little weird, but basically I deleted the `forceUpdate` and `forceUpdateGrid` tests and added one to fallback to `forceUpdate` only if neither `recomputeGridSize` or `recomputeRowHeights` exist.

Here's what the new profile looks like

screen shot 2017-01-06 at 2 56 20 pm


bvaughn:
Hey! Nice little optimization. 😁  Good catch.

PR #533, opened 2017-01-10
changed InfiniteLoader.md

leondraganmdsol:
Changed text from 'high order component' to 'higher-order component' to match FB documentation.
See: https://facebook.github.io/react/docs/higher-order-components.html


bvaughn:
👍

PR #536, opened 2017-01-11
multigrid support onScroll

Pana:
multigrid support onScroll

### DIFF
--- a/source/MultiGrid/MultiGrid.js
+++ b/source/MultiGrid/MultiGrid.js
@@ -423,7 +423,8 @@ export default class MultiGrid extends Component {
       fixedRowCount,
       rowCount,
       scrollToColumn,
-      scrollToRow
+      scrollToRow,
+      onScroll
     } = props
 
     return (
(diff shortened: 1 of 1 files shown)


source/MultiGrid/MultiGrid.js
```diff
@@ -433,7 +434,7 @@ export default class MultiGrid extends Component {
         columnCount={columnCount - fixedColumnCount}
         columnWidth={this._columnWidthRightGrid}
         height={this._getBottomGridHeight(props)}
-        onScroll={this._onScroll}
+        onScroll={(info) => {this._onScroll(info);onScroll && onScroll(info)}}
```
bvaughn: How about moving this into the `_onScroll` method? That way the code's a bit easier to read _and_ we avoid creating a new wrapper function on each render. :)

Pana: Oh right, i thought the _onScroll has been used not only in bottomRight grid, also in other grid. And the passed onScroll only should use in one Grid.
I will update it.

bvaughn:
Thanks 😄  This will go out in a point release soon.

PR #537, opened 2017-01-11
update Grid.md

leondraganmdsol:
In the section explaining cellRenderer function, provide a little bit more information regarding the purpose of the 'style' object and list the properties that are set by default when the grid is created.


bvaughn:
Great addition. Thanks!

Issue #453, opened 2016-11-04
will-change: transform seems to create painting issues in Chrome?


mcyze:
I am using the List component (8.4.1) for a very long list (2000+) of fairly complicated elements.  I have started seeing painting issues but I haven't been able to pin-point the problem exactly.  If I flick the scroll wheel on my mouse to zip down the list, then flick the opposite direction - *sometimes* - when I get to the top of the list, the top cells don't paint.  The elements are in the dom and laid out on screen, I can click on the elements and inspect them -  but scrolling no longer works and my list component is blank.  Resizing the browser, or changing the data in the list seems to trigger a paint and everything is good.

I have tried adding a "refresh" property to my List, and setting it to today's Date at the end of the scroll to trigger a paint... this doesn't seem to have any effect.  What DOES seem to have an effect (but adversely affects scrolling performance) is to override the css property will-change to "auto".

Sorry, I don't have code that reliably re-creates the problem.

Any ideas??

bvaughn:
Sounds like a browser bug to me.

Don't think there's anything I can do to help with this in its current state. I've never seen it and there's no repro steps.

mcyze:
It does seem like a browser bug; I was mostly interested in knowing if you had seen it before.  Thanks for the quick reply.  Setting "will-change" to auto definitely fixes things for me, so I'm good.  Thanks for your quick reply.

bvaughn:
Can you provide a few more details about your browser version, operating system, etc? I can try to pass the word along.

mcyze:
Version 54.0.2840.71 (64-bit)
macOS 10.12

Several of my coworkers saw the same issue on various other versions of macOS and Chrome.

bvaughn:
I'm running the same version of Chrome as you but on OSX 10.11.6. Can't reproduce the behavior you've described by doing what I think you're saying you're doing.

It would help if you could record a video of it happening. I don't know if the above report is enough detail to get any traction from the Chrome team.

bvaughn:
@mcyze: Try this build? (Rename .zip to .tgz ~ Github won't allow me to upload a file with .tgz extension for some reason.)

[react-virtualized-8.4.2.zip](https://github.com/bvaughn/react-virtualized/files/574053/react-virtualized-8.4.2.zip)

maxdeviant:
@bvaughn After looking at this, I don't think this is the same issue that I'm seeing.

[...]

Issue #291, opened 2016-06-22
Faster onScroll with Scrollsync


martinpesout:
I'm using your example https://bvaughn.github.io/react-virtualized/ to create big datatable with fixed table header. Problem is when I start scrolling to the sides. You can see it on your example (tested in MacOS Chrome with magic mouse). If you scroll to the side you can see that header is updated with some delay. Is there any chance how to improve this speed. I know that it's Javascript solution that calculations take some time, but if you check different solution from Facebook http://facebook.github.io/fixed-data-table/example-object-data.html it's also achieved by Javascript and scrolling is fluid.

Thanks for your help.

bvaughn:
Hey @martinpesout,

Unfortunately fixed-data-table is kind of an apples-to-oranges comparison. `ScrollSync` is synchronizing the scroll offsets between multiple, individually-scrollable items. Because scrolling is managed by a separate thread, and JavaScript is only periodically notified of the updated position, there's some latency issues with this. Unfortunately they can't be avoided (at least not to my knowledge, having spent a great deal of time trying to optimize things like this).

That being said, I'm always trying to reduce latency and increase FPS. (I spent most of this past weekend working on FPS improvements for `FlexTable`.) And I'd welcome any PRs or contributions that you or others would like to make to help out in this regard.

But I'm going to close this issue for now. There isn't any direct action I can take to resolve it- at least none that I'm aware of. If you find some time to experiment and you have suggestions though- let's talk! :)

bvaughn:
(We can keep talking here too. I'll see and respond to comments on this issue.)

vincent-io:
Out of curiosity: why doesn't this happen when using the scrollbar? (It moves in sync then)

bvaughn:
I _think_ scrollbar dragging is done in the UI thread. The reason scroll-wheel scrolling is in a separate thread is so they can do the smooth animation.

vojtech-dobes:
Hi, if I understand it correctly, `ScrollSync` is relying on native scrolling and synchronization of the scroll position is happening ex post? Therefore to achieve sooner synchronization one would have to sacrifice benefits of native scrolling - is that the problem?

vincent-io:
So, as a work-around, can we actually artificially drag the scrollbar when scrolling?

bvaughn:
You are both correct.

[...]

PR #541, opened 2017-01-17
Fix mouse wheel tear in scroll synced grids

implausible:
Addresses tearing across scroll synced Grids in Chrome: https://github.com/bvaughn/react-virtualized/issues/453.

This is accomplished by forcing Grids that receive an onScroll prop to use the onWheel event. The onWheel event occurs before any scrolling occurs in the browser. We grab that event, modify the scrollTop/scrollLeft ourselves, and then toss that into the scrollCallback and observe a scroll change.

Tearing in the browser is occurring, because onScroll events actually occur _post_ scroll in windows/linux on chrome. Interestingly, the tearing behavior is only associated with mouse wheel / track pad events. Grabbing the scroll bar and dragging it works as expected where the onScroll event occurs before the browser actually scrolls the contents, so this bug is not present under scrollbar dragging conditions.

Further, as this only happens in scroll sync situations and the current scroll behavior is acceptable for singular grids, we only apply this change when provided an onScroll prop in Grid.

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,6 +121,7 @@
     "karma-phantomjs-launcher": "^1.0.0",
     "karma-sourcemap-loader": "^0.3.6",
     "karma-webpack": "^1.7.0",
+    "lodash.throttle": "^4.1.1",
     "phantomjs-prebuilt": "^2.1.14",
     "postcss": "^5.0.14",
     "postcss-cli": "^2.3.3",
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -7,6 +7,7 @@ import createCallbackMemoizer from '../utils/createCallbackMemoizer'
 import getOverscanIndices, { SCROLL_DIRECTION_BACKWARD, SCROLL_DIRECTION_FORWARD } from './utils/getOverscanIndices'
 import getScrollbarSize from 'dom-helpers/util/scrollbarSize'
 import shallowCompare from 'react-addons-shallow-compare'
+import throttle from 'lodash.throttle'
 import updateScrollIndexHelper from './utils/updateScrollIndexHelper'
 import defaultCellRangeRenderer from './defaultCellRangeRenderer'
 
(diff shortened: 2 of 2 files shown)


bvaughn:
Hi @implausible,

Thanks for putting together this PR! It's an interesting approach to the scroll lag problem.

The essence of this problem (as you probably know) is that modern browsers handle scroll input in a separate thread in order to prevent slow JavaScript from causing the scrolling interaction to feel unresponsive. Unfortunately this results in the scroll lag issue that you're seeing- _but_ it is an important optimization for keeping under-powered devices from feeling really unresponsive to touch input.

There are things developers can do though (eg add "touch" or "wheel" event listeners) that force the browser to behave differently- blocking until JavaScript completes (see [here](https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md#the-problem)). These can "fix" the scroll lag issue but at a potentially large performance cost. For example, your approach here seems to perform well in Chrome and Safari but totally destroys performance in Firefox. I've considered forcing `ScrollSync` to operate this way (synchronously) in the past but have decided against it for this reason.

Unfortunately I don't feel comfortable merging this PR for the same reason.

Perhaps you could modify this approach to enable your `onWheel` handler to be injected in? eg what if `Grid` exposed a new `onWheel: PropTypes.func`  (optinoal) property that enabled you to inject your own throttled wheel handler? This way if it works well for your use-case you could continue to use it without potentially causing regressions for existing apps.

> Further, as this only happens in scroll sync situations and the current scroll behavior is acceptable for singular grids, we only apply this change when provided an onScroll prop in Grid.

Scroll-syncing is not the only use-case for using an `onScroll` prop. Some apps react to changes in scroll position/delta in other, less-direct ways (eg load more data, save scrolled position for later, etc.)

PS. For related background discussion on this check out issue #291.

bvaughn:
Closing for now due to inactivity. We can pick this discussion back up in the future perhaps.

PR #548, opened 2017-01-19
Fix positionFromTop when scrolling window after window resize

### LINKED ISSUE #544: WindowScroller setting incorrect scrollTop after resizing window height
I updated the version of WindowScroller for one of my projects today from `8.8.1` to the current version (`8.11.0`) and noticed this bug. More details about the use cases and the specifics of the bug below.

### To reproduce

To reproduce the bug: 
- Set up `WindowScroller` without specifying the new `scrollElement` prop (i.e. you want it to use `window`). My use case uses a `Grid` inside `WindowScroller`, but I think a similar issue would occur with other components.
- Scroll the page. `scrollTop` is set as expected at this point and everything works well.
- Scroll the page down far enough that the top of the container (i.e. element for your `Grid` or other component inside the `WindowScroller`) is off the top of the viewport.
- Resize the height of your browser window.
- Scroll the page. `scrollTop` is incorrectly set and the `Grid` behaves weirdly because it has the wrong scroll value.

I took a quick look at the project's [`WindowScroller` demo](https://bvaughn.github.io/react-virtualized/#/components/WindowScroller) and see similar "weird" behavior when I follow these steps, so hopefully that will be an easy example to work from.

### Suspected cause

After looking at the changelog and some of the related source code, I suspect this bug came in with the support for a custom target element in v8.10.0. I'm guessing the problem lies in [getPositionFromTop](https://github.com/bvaughn/react-virtualized/blob/master/source/WindowScroller/utils/dimensions.js#L17) and how it is handling the calculations for the `window` use case.

### Additional details:

I rolled back to my previous version (`8.8.1`) and verified this bug does not exist there.

Browser: Chrome 55 on OSX 10.11

Use case in my code: `Grid` inside a `WindowScroller` that only uses WindowScroller for vertical scrolling per the documentation that says it should not be used for horizontal scrolling.

andrewbranch:
Fixes #544 😱 

I am unable to explain why this fixes the issue, but I can verify that it does. The `getScrollTop(container)` term here is necessary when scrolling a custom element. Apparently it breaks things when scrolling `window`. This keeps the calculation for `window` scrolling the same as it was pre-8.10.0, while adding the necessary offset to the calculation for scrolling a custom `scrollElement`.

### DIFF
--- a/source/WindowScroller/utils/dimensions.js
+++ b/source/WindowScroller/utils/dimensions.js
@@ -15,12 +15,13 @@ export function getHeight (element) {
  * In this case the body’s top position will be a negative number and this element’s top will be increased (by that amount).
  */
 export function getPositionFromTop (element, container) {
+  const offset = container === window ? 0 : getScrollTop(container)
   const containerElement = container === window
     ? document.documentElement
     : container
   return (
     element.getBoundingClientRect().top +
-    getScrollTop(container) -
+    offset -
     containerElement.getBoundingClientRect().top
   )
 }


bvaughn:
Thanks buddy!

PR #552, opened 2017-01-24
MultiGrid (with rowHeight and columnWidth functions) resizing

codingbull:
For a MultiGrid (wrapped by an AutoSizer) with rowHeight and columnWidth functions, grid sizes are not recalculated after window resizing. I attached an onResize function to the AutoSizer to explicitly force it via calling MultiGrid.recomputeGridSize(). In the course of debugging that, I encountered these 3 bugs.

Even after calling recomputeGridSize with these fixes, I still needed to call _maybeCalculateCachedStyles, but I wan't sure if that should always occur in the recomputeGridSize call or not.

### DIFF
--- a/source/MultiGrid/MultiGrid.js
+++ b/source/MultiGrid/MultiGrid.js
@@ -99,7 +99,7 @@ export default class MultiGrid extends Component {
     const { columnWidth, fixedColumnCount, fixedRowCount, rowHeight } = this.props
 
     if (
-      columnWidth !== nextProps.columnWidth ||
+      columnWidth instanceof Function || columnWidth !== nextProps.columnWidth ||
       fixedColumnCount !== nextProps.fixedColumnCount
     ) {
       this._leftGridWidth = null
@@ -107,7 +107,7 @@ export default class MultiGrid extends Component {
 
     if (
       fixedRowCount !== nextProps.fixedRowCount ||
-      rowHeight !== nextProps.rowHeight
+      rowHeight instanceof Function || rowHeight !== nextProps.rowHeight
     ) {
       this._topGridHeight = null
     }
--- a/source/MultiGrid/MultiGrid.js
+++ b/source/MultiGrid/MultiGrid.js
@@ -90,6 +90,10 @@ export default class MultiGrid extends Component {
       columnIndex: col,
       rowIndex: rowIndex
     })
+
+    this._leftGridWidth = null
+    this._topGridHeight = null
+    this._maybeCalculateCachedStyles(null, this.props)
   }
 
   componentWillMount () {
@@ -100,7 +104,7 @@ export default class MultiGrid extends Component {
     const { columnWidth, fixedColumnCount, fixedRowCount, rowHeight } = this.props
 
     if (
-      columnWidth instanceof Function || columnWidth !== nextProps.columnWidth ||
+      columnWidth !== nextProps.columnWidth ||
       fixedColumnCount !== nextProps.fixedColumnCount
     ) {
       this._leftGridWidth = null
@@ -108,7 +112,7 @@ export default class MultiGrid extends Component {
 
     if (
       fixedRowCount !== nextProps.fixedRowCount ||
-      rowHeight instanceof Function || rowHeight !== nextProps.rowHeight
+      rowHeight !== nextProps.rowHeight
     ) {
       this._topGridHeight = null
     }
@@ -340,7 +344,6 @@ export default class MultiGrid extends Component {
       this._bottomLeftGridStyle = {
         left: 0,
         outline: 0,
-        overflow: 'hidden',
         overflowX: 'hidden',
         overflowY: 'hidden',
         position: 'absolute',
@@ -368,7 +371,6 @@ export default class MultiGrid extends Component {
       this._topLeftGridStyle = {
         left: 0,
         outline: 0,
-        overflow: 'hidden',
         overflowX: 'hidden',
         overflowY: 'hidden',
         position: 'absolute',
@@ -385,7 +387,6 @@ export default class MultiGrid extends Component {
       this._topRightGridStyle = {
         left: this._getLeftGridWidth(props),
         outline: 0,
-        overflow: 'hidden',
         overflowX: 'hidden',
         overflowY: 'hidden',
         position: 'absolute',
--- a/source/MultiGrid/MultiGrid.js
+++ b/source/MultiGrid/MultiGrid.js
@@ -340,6 +340,8 @@ export default class MultiGrid extends Component {
         left: 0,
         outline: 0,
         overflow: 'hidden',
+        overflowX: 'hidden',
+        overflowY: 'hidden',
         position: 'absolute',
         ...styleBottomLeftGrid
       }
@@ -366,6 +368,8 @@ export default class MultiGrid extends Component {
         left: 0,
         outline: 0,
         overflow: 'hidden',
+        overflowX: 'hidden',
+        overflowY: 'hidden',
         position: 'absolute',
         top: 0,
         ...styleTopLeftGrid
@@ -381,6 +385,8 @@ export default class MultiGrid extends Component {
         left: this._getLeftGridWidth(props),
         outline: 0,
         overflow: 'hidden',
+        overflowX: 'hidden',
+        overflowY: 'hidden',
         position: 'absolute',
         top: 0,
         ...styleTopRightGrid
--- a/source/MultiGrid/MultiGrid.js
+++ b/source/MultiGrid/MultiGrid.js
@@ -72,22 +72,23 @@ export default class MultiGrid extends Component {
     rowIndex = 0
   } = {}) {
     const { fixedColumnCount, fixedRowCount } = this.props
-
-    this._bottomLeftGrid && this._bottomLeftGrid.measureAllCells({
-      columnIndex,
-      rowIndex: rowIndex - fixedRowCount
+    const col = (columnIndex >= fixedColumnCount ? columnIndex - fixedColumnCount : columnIndex)
+    const row = (rowIndex >= fixedRowCount ? rowIndex - fixedRowCount : rowIndex)
+    this._bottomLeftGrid && this._bottomLeftGrid.recomputeGridSize({
+      columnIndex: columnIndex,
+      rowIndex: row
     })
-    this._bottomRightGrid && this._bottomRightGrid.measureAllCells({
-      columnIndex: columnIndex - fixedColumnCount,
-      rowIndex: rowIndex - fixedRowCount
+    this._bottomRightGrid && this._bottomRightGrid.recomputeGridSize({
+      columnIndex: col,
+      rowIndex: row
     })
-    this._topLeftGrid && this._topLeftGrid.measureAllCells({
-      columnIndex,
-      rowIndex
+    this._topLeftGrid && this._topLeftGrid.recomputeGridSize({
+      columnIndex: columnIndex,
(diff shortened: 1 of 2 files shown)


source/MultiGrid/MultiGrid.js
```diff
@@ -72,22 +72,23 @@ export default class MultiGrid extends Component {
...
-    this._bottomLeftGrid && this._bottomLeftGrid.measureAllCells({
-      columnIndex,
-      rowIndex: rowIndex - fixedRowCount
+    const col = (columnIndex >= fixedColumnCount ? columnIndex - fixedColumnCount : columnIndex)
+    const row = (rowIndex >= fixedRowCount ? rowIndex - fixedRowCount : rowIndex)
+    this._bottomLeftGrid && this._bottomLeftGrid.recomputeGridSize({
+      columnIndex: columnIndex,
+      rowIndex: row
     })
-    this._bottomRightGrid && this._bottomRightGrid.measureAllCells({
```
bvaughn: Duh, this was a stupid typo on my part. Apologies.

source/MultiGrid/MultiGrid.js
```diff
@@ -99,15 +100,15 @@ export default class MultiGrid extends Component {
...
+      columnWidth instanceof Function || columnWidth !== nextProps.columnWidth ||
       fixedColumnCount !== nextProps.fixedColumnCount
     ) {
       this._leftGridWidth = null
     }
 
     if (
       fixedRowCount !== nextProps.fixedRowCount ||
-      rowHeight !== nextProps.rowHeight
+      rowHeight instanceof Function || rowHeight !== nextProps.rowHeight
```
bvaughn: The previous behavior here was intentional. If a function prop is passed, `Grid` (and so `MultiGrid`) don't assume they can detect when it has changed. That's because (a) things outside of the function may invalidate the size without the function prop itself changing and (b) people often define inline props like so: `rowHeight={(index) => blah}` which will recreate on each render. Invalidating cache is expensive so this is an intentional optimization. If a user uses a function prop, he should manually clear the cache.

source/MultiGrid/MultiGrid.js
```diff
@@ -366,6 +369,8 @@ export default class MultiGrid extends Component {
         left: 0,
         outline: 0,
         overflow: 'hidden',
+        overflowX: 'hidden',
+        overflowY: 'hidden',
```
bvaughn: Why are the long-form props needed?

source/MultiGrid/MultiGrid.test.js
```diff
@@ -88,6 +113,19 @@ describe('MultiGrid', () => {
...
+      let multiGrid
+      findDOMNode(render(getMarkup({
+        fixedColumnCount: 1,
+        fixedRowCount: 1,
+        ref: (ref) => {
+          multiGrid = ref
+        }
+      })))
+
+      multiGrid.recomputeGridSize()
```
bvaughn: I don't understand the purpose of this test. Could you explain what it's testing?

bvaughn:
> For a MultiGrid (wrapped by an AutoSizer) with rowHeight and columnWidth functions, grid sizes are not recalculated after window resizing

Function props like `rowHeight` and `columnWidth` don't change when the `Grid`'s outer size changes so I'm not sure why they are mentioned together like this. The size of a cell and the size of the `Grid` are different. `AutoSizer` only drives the size of the `Grid`

source/MultiGrid/MultiGrid.js
codingbull: That makes sense on the optimization. By what method should the cache be cleared?
Perhaps a new public method like:
```
clearCachedStyles() {
  this._topGridHeight = null;
  this._leftGridWidth = null;
  this._maybeCalculateCachedStyles(null, this.props)
}
```

source/MultiGrid/MultiGrid.js
codingbull: I think they are needed to counteract gridStyle.overflowX and gridStyle.overflowY being set in
https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/Grid.js#L566

source/MultiGrid/MultiGrid.test.js
codingbull: When I first switched the measureAllCells calls to recomputeGridSize, negative values were getting passed as column and row indexes, which would throw an error. This was my (poor) failing test prior to adding the comparison against the fixed row/col sizes.

codingbull:
Sorry, I didn't explain the setup well. Nothing wrong with AutoSizer, I just wanted to explain my usage since I didn't have a stand-alone test case to show. 
I've got a rowHeight method that is returning different sizes for my fixed row based on the window size. I realize that having non-constant cell sizes is a performance hit, but I'm still getting phenomenal render performance even on mobile (so much thanks and kudos to you for that).
After a window resize, the cell was getting new position styles, but the top-left and top-right grids still have the cached styles from prior to the resize.

```
isSmallScreen() {
    // (fyi, IncludeMedia is just exposing my CSS media query breakpoints into the JS).
    return IncludeMedia.lessThan('md');
}
_rowHeight({rowIndex}) {
    if (rowIndex == 0) { return (this.isSmallScreen() ? 150 : 100); }
    return 40;
}
```

source/MultiGrid/MultiGrid.test.js
codingbull: I'm not sure what else to explicitly assert, but at least I can add expect(...).not.toThrowError(...);

source/MultiGrid/MultiGrid.js
bvaughn: Users should call `recomputeGridSize` if something has happened to invalid one or more columns/rows. `Grid` then [resets its own cache when `recomputeGridSize` is called](https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/Grid.js#L273-L276).

source/MultiGrid/MultiGrid.js
bvaughn: Roger that. We should go ahead and get rid of the short-hand form (`overflow`) then. It's totally redundant.

source/MultiGrid/MultiGrid.test.js
bvaughn: Hm. I see.

`expect(...).not.toThrowError(...)` is kind of the default behavior of any testing framework though so probably not worth explicitly adding.

Perhaps I was just confused by the name of this test. What makes it different than [this one](https://github.com/codingbull/react-virtualized/blob/9de4f2b786a61cfa3ad17dc2eb86358440aa6b65/source/MultiGrid/MultiGrid.test.js#L42-L62) for instance. You could update it to say something about `recomputeGridSize` specifically, since that's what it's testing, but I wonder if we couldn't do 1 better than that and assert that it _actually_ recomputes the inner `Grid` sizes (and resets `_leftGridWidth` and `_topGridHeight` like we chatted about above)?

bvaughn:
> After a window resize, the cell was getting new position styles, but the top-left and top-right grids still have the cached styles from prior to the resize.

Thank you for clarifying! I understand now. 😁

I _think_ this is actually caused by a combination of the typo you fixed [here](https://github.com/bvaughn/react-virtualized/pull/552/files#r97701281) and what I mentioned in [this comment](https://github.com/bvaughn/react-virtualized/pull/552#discussion_r97809453).

Thanks for taking the time to submit this PR too! Couple of minor changes and I think it will be good to merge!

source/MultiGrid/MultiGrid.test.js
codingbull: I was thinking the same thing this morning, I'll see what I can whip up.

codingbull:
@bvaughn I think I've got a much better test now.

source/MultiGrid/MultiGrid.js
```diff
@@ -72,23 +72,28 @@ export default class MultiGrid extends Component {
     rowIndex = 0
   } = {}) {
     const { fixedColumnCount, fixedRowCount } = this.props
-
-    this._bottomLeftGrid && this._bottomLeftGrid.measureAllCells({
-      columnIndex,
-      rowIndex: rowIndex - fixedRowCount
+    const col = (columnIndex >= fixedColumnCount ? columnIndex - fixedColumnCount : columnIndex)
+    const row = (rowIndex >= fixedRowCount ? rowIndex - fixedRowCount : rowIndex)
```
bvaughn: Sorry, just noticed these 2 lines. This change doesn't look right. I believe these 2 lines should be reverted to [the way it was before](https://github.com/bvaughn/react-virtualized/blob/080572f2e74af72416ba2f542ab6a6ea6aa3476a/source/MultiGrid/MultiGrid.js#L76-L91) but with your correction of my `recomputeGridSize` typo 😄

source/MultiGrid/MultiGrid.js
```diff
@@ -380,7 +387,8 @@ export default class MultiGrid extends Component {
       this._topRightGridStyle = {
         left: this._getLeftGridWidth(props),
         outline: 0,
-        overflow: 'hidden',
+        overflowX: 'hidden',
+        overflowY: 'hidden',
```
bvaughn: Good fix 👍

source/MultiGrid/MultiGrid.js
codingbull: Without that guard,  I get:

```
TypeError: Cannot read property 'offset' of undefined
    at CellSizeAndPositionManager.getVisibleCellRange (CellSizeAndPositionManager.js:184)
    at ScalingCellSizeAndPositionManager.getVisibleCellRange (ScalingCellSizeAndPositionManager.js:150)
    at Grid._calculateChildrenToRender (Grid.js:544)
    at Grid.componentWillUpdate (Grid.js:428)
    at ReactCompositeComponent.js:711
    at measureLifeCyclePerf (ReactCompositeComponent.js:75)
    at ReactCompositeComponentWrapper._performComponentUpdate (ReactCompositeComponent.js:710)
    at ReactCompositeComponentWrapper.updateComponent (ReactCompositeComponent.js:645)
    at ReactCompositeComponentWrapper.receiveComponent (ReactCompositeComponent.js:547)
    at Object.receiveComponent (ReactReconciler.js:125)
```

codingbull: Instead of dealing with it in MultiGrid, perhaps a change to [CellSizeAndPositionManager#resetCell](https://github.com/bvaughn/react-virtualized/blob/080572f2e74af72416ba2f542ab6a6ea6aa3476a/source/Grid/utils/CellSizeAndPositionManager.js#L180) instead?

```
resetCell (index: number): void {
    let guarded = Math.max(index, 0)
    this._lastMeasuredIndex = Math.min(this._lastMeasuredIndex, guarded - 1)
  }
```

bvaughn: Ah, fair enough. I think the correct fix is to add a `Math.max(0, maybeNegativeIndex)` in `MultiGrid` so that it doesn't pass negative values. As it is (with this change) we may skip cells in the bottom and/or right grids if indices > 0 and < fixed-count are passed.

bvaughn:
I've merged your PR locally, after fixing that one small issue. Thanks so much, again!

bvaughn:
Not sure why Github still thinks the PR is open. You can see that it's been merged [here](https://github.com/bvaughn/react-virtualized/commits/master). Oh well. 😁

bvaughn:
Released with 8.11.3

Gave you credit for this fix [in the changelog](https://github.com/bvaughn/react-virtualized/blob/master/CHANGELOG.md)

PR #566, opened 2017-02-08
iOS scroll bounce fix - skip event if scrollTop < 0

### LINKED ISSUE #532: visual bug: flickering when over scrolling on ios
using ios 9.2 device, go to https://bvaughn.github.io/react-virtualized/
pull down on the List so the first row goes lower (ie. this is an "over scroll" which can be done on ios)
i will see intermittent flicker/jumping of the list rendering for brief moment at the top of the overscroll. it should not render at the top of an overscroll.

this bug is not always apparent, if you do not see it first time, try overscrolling again or overscrolling faster.

i have tried adjusting overscan setting up and down, but I have not noticed any visible improvement.

toddtarsi:
If we're at a scrollTop above 0, just avoid triggering the scroll event entirely, or maybe I should pass the event through.
This is a McDonalds fix, feel free to critique this as needed. Thanks!
Resolves #532

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -900,6 +900,12 @@ export default class Grid extends Component {
       return
     }
 
+    // On iOS, we can arrive at negative offsets by swiping past the start
+    // To prevent flicker here, we make playing in the negative offset zone cause nothing to happen.
+    if (event.target.scrollTop < 0) {
+      return
+    }
+
     // Prevent pointer events from interrupting a smooth scroll
     this._debounceScrollEnded()
 


bvaughn:
Hey @toddtarsi,

This fix seems reasonably safe- but I'd really like to be able to reproduce the problem and verify that this change _fixes it_ before merging.

Any chance I could get a more detailed description of how to reproduce the behavior? Or a video demo?

bvaughn:
PS. Thanks for the PR 😁

toddtarsi:
Absolutely, I am filming device repro on my local iOS using the demo server. :)
Will have shortly.

toddtarsi:
Okay, I had some issues with the demo server, so I ended up using my team's application to demonstrate the fix:

PFA using the standard virtualized list (technically infinite loader, but irrelevant here when everything is loaded):

issue

PFA using the virtualized list after this fix:

resolution

bvaughn:
Thanks! This is great. Very helpful.

maxsalven:
Hi @bvaughn @toddtarsi,

This fix solved issues when scrolling upwards, however I still get visual flickers when over scrolling downwards past the bottom of a list.

Let me know if you are unable to reproduce, I can try capture a video somehow if so. Many thanks.

toddtarsi:
@maxsalven -

Nah. That issue makes sense. In the same way this avoids recalculating the render when scrolling above the beginning, I figured at some point we'd need to resolve this issue when scrolling past the end as well. If you could help me out by demoing if the fix (#616) resolves your issue, I would be appreciative. 

@bvaughn - Sorry I didn't include this in the original fix, but I wrote a small patch for it here: #616

Issue #576, opened 2017-02-13
Rendering problems for Grid with big amount of rows


martinpesout:
Hi, in the past there was a problem with rendering in Grid with 40k+ rows (https://github.com/bvaughn/react-virtualized/issues/524). I've found that there is still some problem with this Grid.

Steps to reproduce:

* Problem is visible on Grid with more rows. It works for table with 10000 rows. It doesn't work for table with 40000 rows.
* Problem is visible in Chrome (reproduce on Mac devices and on Linux, even when using magic mouse and also on normal mouse with scrollwheel). Than we will have to do many times quick scrolling down and up.

You can reproduce it on your example https://bvaughn.github.io/react-virtualized/#/components/Grid with 40000 rows.

You will get this problem: 

![snimek obrazovky 2017-02-13 v 13 32 31](https://cloud.githubusercontent.com/assets/239468/22883885/88a7c6c8-f1f2-11e6-9056-5f33cc81c301.png)

![uno-13-2017 13-35-03](https://cloud.githubusercontent.com/assets/239468/22883894/8d66b9b2-f1f2-11e6-8e83-1bcc962c4825.gif)

benjaminjs:
I'm having the same issue exactly. Using a table and the infinite load. I have a million row count, but 10,000 works fine.

<img width="1043" alt="screen shot 2017-02-13 at 10 41 34 am" src="https://cloud.githubusercontent.com/assets/2501984/22892768/4a8e75a8-f1d9-11e6-818b-06b9f909ba3a.png">

bvaughn:
Hm. I'm unable to reproduce this myself (using Chrome 56.0.2924.87, macOS 10.12.3) on the demo site. I just cranked up the demo to 400,000 rows at 30px height each (12M pixels total) and mouse and scrollbar scrolling seem fine.

johnhaley81:
@martinpesout does it flicker if you use the scroll bar to scroll?

bvaughn:
To be clear- the above video I attached shows trackpad flinging, scrollbar dragging, and arrow key scrolling as well. All 3 seem fine to me.

benjaminjs:
It only happens at the very top.

bvaughn:
> It only happens at the very top.

Ah! That's helpful. I see it now.

bvaughn:
I believe I have a fix 😄

bvaughn:
Fixed by commit e78036a. Will go out with version 9 (see PR #577)

[...]

Issue #568, opened 2017-02-09
Documentation for Collection could be corrected and/or clearer w/r/t `recomputeCellSizesAndPositions`


jasongonzales23:
I needed a hack/workaround to get my collection to update even though the length of it hadn't changed.

On the home page of the repo it states you can use `forceUpdate` as a public method on the Collection component. I couldn't get this to work. Does it still work or is that part of the documentation out of date?

On the docs for the Collection component it states you can force an update by calling `recomputeCellSizesAndPositions` as a public method. That said, it wasn't obvious to me how to do this (as this notion of a public method isn't terribly idiomatic to React Components). After digging for quite a while I learned that you do it like so:

```
<Collection
    ref={ref => this._collection = ref}
   ...
/>
```

And then wherever you need to force a render:
```
componentWillReceiveProps() {
    if (this._collection) {
        this._collection.recomputeCellSizesAndPositions()
    }
}
```
I'll be happy to make this change in a PR as soon as I have a moment if you like.

Cheers!

bvaughn:
I think refs are explained reasonably clearly in the [official React docs](https://facebook.github.io/react/docs/refs-and-the-dom.html). Their use is probably not super common, but they are common enough for things like focus management, animations, 3rd party libraries (like react-virtualized). I'm wary of duplicating the official docs too much here because I won't be able to keep them in sync.

```jsx
componentWillReceiveProps(nextProps) {
    if (this._shouldCellsRemeasure(nextProps)) {
      // Your data has changed in a way that affects the size or position of cells.
      // This is an expensive operation; don't do it unless you need to.
      // recomputeCellSizesAndPositions() itself is defined on the Collection component.
      this._collection.recomputeCellSizesAndPositions()
    } else if (this._shouldCellsRedraw(nextProps)) {
      // Your data has changed in a way that requires one or more cells to re-draw.
      // Cell sizes and positions are still valid.
      // This is a relatively cheap operation.
      // forceUpdate() is a built-in method available to all React components.
      this._collection.forceUpdate()
    }
}

render () {
  return (
    <Collection
      ref={ref => this._collection = ref}
      {...this.props}
    />
  )
}
```

[...]

PR #561, opened 2017-02-02
Per item caching

Mathieuu:
Add support to cache item size in addition to the existing row / column caching.

See #562 for details

The PR is not ready:
- Expose the interface to clearItemHeights, clearItemHeight, clearItemWidth, clearItemWidths.
- Replace item by cell, I found it confusing since you called the Row/Column cache, cellCache, when actually it doesn't cache anything about the cell. So I used item in the meantime but it is probably a poor choice.
- Write tests

Issue #559, opened 2017-01-30
missing MultiGrid.forceUpdateGrid()


oallouch:
Hi,
I just used this lib for the first time. I needed a MultiGrid...then came love. Great work !
Anyway, a forceUpdateGrid method is missing.
The obvious workaround was to use _bottomLeftGrid, _bottomRightGrid (etc) properties.

Thx in advance,
Olivier from Paris

bvaughn:
This would be a great first contributor issue. You interested? :)

bvaughn:
Fixed via 8482ad8; will release with version 9 (see PR #577). `MultiGrid` now has a method `forceUpdateGrids`

PR #577, opened 2017-02-13
Episode 9.0 (The One About React Fiber)

bvaughn:
To test version 9 yourself- `npm i react-virtualized@next`

You can also give it a spin [on the demo site](https://bvaughn.github.io/react-virtualized/) which is now powered by React fiber too!

# Remaining TODO items

- [x] ~~`CellMeasurerCache` should cache and return an average cell size as a sub for an unknown cell's `width`/`height` (instead of `minWidth`/`minHeight`). This will make it more accurate over time when estimating layout of previously unmeasured cells.~~ (deferred)
- [x] Add a few remaining unit tests for new functionality
- [x] Update docs and write a better upgrade guide

# Non-breaking changes

The following backwards-compatible issues will be addressed with version 9:

#### Bug fixes:
  * #532: iOS scroll bounce fix
  * #559: Added public method `forceUpdateGrids` to `MultiGrid` that calls `forceUpdate` on all child `Grid`s.
  * #561: Per item caching
  * #568: `Collection.forceUpdate()` now calls `forceUpdate` on inner `CollectionView`
  * #576: Rendering problems for `Grid` with huge number of rows
#### New features:
  * #554: Overscan direction can be customized using the new `overscanIndicesGetter` prop [for `Grid`](https://github.com/bvaughn/react-virtualized/blob/master/docs/Grid.md#prop-types).
  * #578: Presentational `Table` styles now support iOS 8.4
#### DEV mode warnings:
  * Use of `CellMeasurer` with an improperly configured `Grid` will now log a warning.
  * `Grid` cells rendered without a `style` prop now log a warning.

# Backwards-breaking changes

* **Supported React versions have changed.** Previously all React 14 and 15 releases were supported. Going forward only React 15.3 and newer will be supported. This was done to drop a peer dependency on `react-addons-shallow-compare` and replace it with `React.PureComponent` [as the docs suggest](https://facebook.github.io/react/docs/shallow-compare.html).
screen shot 2017-02-13 at 1 44 43 pm

* **`CellManager` has been completely rewritten to support fiber.** It's more flexible than before but uses a very different syntax. See below for more detailed instructions on how to upgrade.

### `CellMeasurer` changes

Previously, `CellMeasurer` wrapped a `Grid` and measured all rows and/or columns, like so:

```jsx
<CellMeasurer
  cellRenderer={cellRenderer}
  cellSizeCache={cellSizeCache}
  columnCount={columnCount}
  rowCount={rowCount}
>
  {({ getColumnWidth, getRowHeight }) => (
    <Grid
      columnCount={columnCount}
      columnWidth={getColumnWidth}
      cellRenderer={cellRenderer}
      rowCount={rowCount}
      rowHeight={getRowHeight}
      {...otherProps}
    />
  )}
</CellMeasurer>
```

This had one major drawback: `CellMeasurer` could not be used to measure only specific rows/columns. A secondary, related drawback was that cell measurements were cached by index. Because of this insertions, deletions, and re-ordering invalidated measurements and require re-measuring.

The new `CellMeasurer` wraps individual cells, allowing selective measurement. By providing a `keyMapper`, measurements can also be cached by cell ID instead of index- making them reusable after insertions, deletions, and re-orders. The new syntax for using `CellMeasurer` is like so:

```jsx
// In this example, average cell width is assumed to be about 100px.
// This value will be used for the initial `Grid` layout.
// Cell measurements smaller than 75px should also be rounded up.
// Height is not dynamic.
const cache = new CellMeasurerCache({
  defaultWidth: 100,
  minWidth: 75,
  fixedHeight: true
});

function cellRenderer ({ columnIndex, key, parent, rowIndex, style }) {
  const content // Derive this from your data somehow

  return (
    <CellMeasurer
      cache={cache}
      columnIndex={columnIndex}
      key={key}
      parent={parent}
      rowIndex={rowIndex}
    >
      {content}
    </CellMeasurer>
  );
}

function renderGrid (props) {
  return (
    <Grid
      {...props}
      columnWidth={cache.columnWidth}
      deferredMeasurementCache={cache}
      cellRenderer={cellRenderer}
    />
  );
}
```


edulan:
Can't wait to play with the new CellMeasurer :smile:

Thanks a lot for all the work @bvaughn!!!

bvaughn:
D'oh! Sorry @edulan I missed your comment. Thank you!

I plan a final release soon- but until then you can play with it via `npm i react-virtualized@next` if you'd like to give it a spin early. I don't expect the API to change any. I'm mostly finishing up documentation and adding a few more automated tests.

romulof:
Isn't Fiber only being enabled by default in React 16?
If so, the new CellMeasurer would require React 16 as dependency.

bvaughn:
`CellMeasurer` doesn't require React 16 to work. React 16 won't work with react-virtualized <= 8 though because the subtree/portal render method `CellMeasurer` used to use- which used to be synchronous- is asynchronous in React 16. So I rewrote `CellMeasurer` to support async out of the box and it is now compatible with React 16 (and older).

I did bump the React peer dependency up to 15.3 though because I wanted to use `PureComponent` and stop requiring people to also instead `react-addons-shallow-compare` for the `shallowCompare` util.

romulof:
Great. I thought you were using some Fiber-specific feature.

I still need to invalidate/update cache if a cell changes size or this new CellMeasurer handles it automatically?

bvaughn:
Still need to manually invalidate. Otherwise I'd have to force layout frequently which would defeat the purpose.

romulof:
Using this new CellMeasurer still renders the cell twice (for measure and actual rendering) ?

bvaughn:
Yes. You could prevent the second render (if it's worth it) using `shouldComponentUpdate`. The only prop that should change is the incoming `style` (which will have actual instead of estimated measurements).

mquan:
@bvaughn Is the way to manually trigger cell re-measurement [here](http://stackoverflow.com/a/40659812) still valid?  If not, can you point me to documentation of the new way of doing it for CellMeasurer & Grid combo? thanks

bvaughn:
@mquan The way to clear measurements now is to reset your `CellMeasurerCache`:

```js
// Clear measurements so CellMeasurer will know to re-measure when it is next rendered
cellMeasurerCache.clearAll()

// Let Grid know to clear its cached position data and re-render
gridRef.recomputeGridSize()
```

kirtimukha:
Maybe it's very simple question, but please understand I'm very new on react. 
Can you let me know where do I put the cellMeasureCache.clearAll() function?
I'm having a proble lists are disappear when I scrolled up. 
I used "cache.clear()" on " isRowLoaded (){ here:where i put the cache.clear() }"function now. 


and this is my retunr   <article className="wrap-article">
          {/*
                    <div
                        className="article img _0 clearfix"
                    >
                        <div className="cont no-article clearfix">
                                                      
                        </div>
                    </div>
                    */}
          <>

              <AutoSizer >
                {({ width, height }) => (
                  <InfiniteLoader
                  isRowLoaded={this.isRowLoaded.bind(this)}
                  loadMoreRows={this.fetchFeed.bind(this)}
                  rowCount={5}
                >
                 {({ onRwsRendered, registerChild }) => (
                    <List
                      height={height}
                      autoHeight
                      width={width}
                      overscanRowCount={0}
                      rowCount={this.state.community_list.length}
                      rowHeight={this.cellMeasurerCache.rowHeight}
                      rowRenderer={this.renderRow.bind(this)}
                      onRowsRendered={onRwsRendered}
                      deferredMeasurementCache={this.cellMeasurerCache}
                      onScroll={this.scrollListener}
                      ref={registerChild}
                      className="ReactVirtualize_customClass"
                    />
                  )}
                </InfiniteLoader>
              )}
            </AutoSizer>
            </>
        </article>


Can you help me?

Issue #554, opened 2017-01-27
overscan direction = both


kachkaev:
Overscan (#64) helps UX a lot when cells contain slowly rendered graphics. However, when a user suddenly changes scroll direction, flickering still occurs. This happens because the cells are deleted immediately after going even by 1 pixel beyond the screen.

Some exploration tasks expect users to navigate around a virtual grid in unpredictable directions to perform visual comparison between objects or features. It would help a lot if developers could tell `react-virtualized` to keep cell cache in both directions when needed. This could be something like `overscanBothDirections={true}` or `reverseOverscanColumnCount={42}` + `reverseOverscanRowCount={43}`.

What do you think?

bvaughn:
The amount of flickering that happens when a user reverses scroll direction should be _extremely_ minimal since the overscan direction will be updated after the first "scroll" event. I'm leaning toward saying that adding API complexity in this area wouldn't be worth it.

If you'd like to put together a proof of concept (a fork) I'd be happy to look at it.

kachkaev:
Hi again @bvaughn,

Check out this MVE:  
https://kachkaev.github.io/react-virtualized-overscan-direction-issue/

Source:  
https://github.com/kachkaev/react-virtualized-overscan-direction-issue/blob/master/src/App.js

When you scroll in one direction, the cells almost always render outside the viewport, because `overscanRowCount=2`. However, when you suddenly decide to scroll back, you start seeing green or red artifacts while the tiles or SVGs are being prepared. `reverseOverscanRowCount={N}` could solve this negative visual effect.

The problem I was able to reproduce in the demo is not as severe as in the real app I've been making. I've got [WebGL maps with vector tiles](https://www.mapbox.com/mapbox-gl-js/api/) in there and the components require 0.5-1 sec to warm up. MapboxGL.js has some issues with webpack, so I could not fit it into `create-react-app` for the demo (it works OK with [`react-boilerplate`](https://github.com/react-boilerplate/react-boilerplate), but that would make the MWE too bulky).

Would be glad to know what you think! The visual effect that's being observed at the moment makes it harder to perform data exploration and certain cognitive comparison tasks and it'd be great to escape this problem.

bvaughn:
Sorry for the slow response time. I see what you're describing in the demo you've shared. _However_ I'm concerned that reverse over-scanning probably won't be sufficient to avoid the problem you're mentioning. I'll try to elaborate.

[...]

PR #579, opened 2017-02-15
Expose overscanIndices getter to make it user-extendable

clauderic:
This PR exposes a new `overscanIndicesGetter` prop to allow the default behavior to be extended.

### Primary goal
By default, React Virtualized will try to optimize the overscan indices based on the scroll direction. This is great for most use cases, but it is something that users may want to implement differently, for a variety of reasons (for example, #554), and it would be challenging to try to support them all out of the box.

I took inspiration from the `defaultCellRangeRenderer` approach, which I believe is a great approach to give advanced users a more fine-grained control while still providing a sensible default and keeping the API surface that needs to be maintained small.

I'm absolutely open to feedback on this and making changes as needed!


bvaughn:
This is a very pragmatic solution. Love it. Thanks @clauderic!

Issue #578, opened 2017-02-14
Columns didn't align properly on iOS 8.4


lyf1988:
I am developing a web app using react, and planning to support iOS version down to iOS8.
But I found the Columns did't display properly on iOS8, it seems aligned left instead of display like a table.
If iOS 8 safari is out of support list, please just close this. 
Thank you all for providing such good looking and easy to use components.

## This is the Screen shot from OSX simulator using safari on iPhone 6 Plus - iOS 8.4.
![simulator screen shot 14 feb 2017 03 23 12](https://cloud.githubusercontent.com/assets/11595729/22913799/e5adcaa8-f264-11e6-879a-beb4fc5e03ed.png)

## This is same page on iPhone 6 Plus - iOS 9.0
![simulator screen shot 14 feb 2017 03 30 59](https://cloud.githubusercontent.com/assets/11595729/22913946/d6e1429c-f265-11e6-92fa-d9fce2affda1.png)

clauderic:
### Issue:
Looks like this is because of the current [autoprefixer config](https://github.com/bvaughn/react-virtualized/blob/ea90d300739009dffc08513d0b150f49aac90177/webpack.config.umd.js#L42) being set to last 2 versions. 

I went ahead and tested in iOS 8.4 on the simulator and indeed replicated the same behavior on the demo site. 

Here's what the generated prefixed CSS currently looks like for `Table__row`:

<img width="1344" alt="screen shot 2017-02-15 at 12 18 01 pm" src="https://cloud.githubusercontent.com/assets/1416436/22986286/fd977b18-f378-11e6-9d7f-6f829181b2b8.png">

Notice it's falling back to `display: -webkit-box;`, which seems to behave differently on Safari iOS 8 and prior.

If I go ahead and set it to `display: -webkit-flex;` all is well:

<img width="1343" alt="screen shot 2017-02-15 at 12 18 35 pm" src="https://cloud.githubusercontent.com/assets/1416436/22986353/35e312ac-f379-11e6-9de2-afa6beaf80e6.png">

### Possible solutions
@bvaughn Happy to submit a small PR for this if you feel like updating the autoprefixer config is the right way to go here. 

Alternatively @lyf1988, if you're using a module bundler like webpack, you can always use your own Autoprefixer config in your webpack loader configuration for `css` files.

bvaughn:
Good catch @clauderic!

If it's as simple as nudging the autoprefixer config, I'd be happy to accept a PR 😄

[...]

PR #580, opened 2017-02-15
Update PostCSS & Autoprefixer configs

clauderic:
This PR adds a unified `postcss.config.js` file to avoid duplication of settings across different webpack.config files and is also shared by postcss-cli (as per the recommended [postcss-loader setup guide](https://github.com/postcss/postcss-loader))

As discussed in #578, I also added iOS8 to the list of browsers so Autoprefixer can add the right vendor-prefixed properties. 

You'll notice I had to explicitly add `&minimize=false` to `css-loader`. This is to safeguard against `webpack.optimize.UglifyJsPlugin` affecting our vendor-prefixed css (see https://github.com/webpack/webpack/issues/283 for more details).

Finally, I noticed some CSS configs in [webpack.config.umd.js](https://github.com/bvaughn/react-virtualized/blob/master/webpack.config.umd.js#L42) but I don't think these are still used (I could be wrong, I'm guessing this was later replaced by the [build:css](https://github.com/bvaughn/react-virtualized/blob/master/package.json#L16) script). Is it safe to assume this can be removed? Happy to update this PR if so.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "scripts": {
     "build": "npm run build:commonjs && npm run build:css && npm run build:es && npm run build:demo && npm run build:umd",
     "build:commonjs": "npm run clean:commonjs && cross-env NODE_ENV=production cross-env BABEL_ENV=commonjs babel source --out-dir dist/commonjs --ignore *.example.js,*.test.js,source/demo/,source/tests.js",
-    "build:css": "postcss --use autoprefixer source/styles.css > styles.css",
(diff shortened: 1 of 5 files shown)


bvaughn:
You're awesome, @clauderic. Thanks for taking the time to put this PR together!

> Finally, I noticed some CSS configs in webpack.config.umd.js but I don't think these are still used

You're right. This is probably just copy-paste detritus from when I split the webpack configs up. It can totally be removed.

PR #583, opened 2017-02-17
Document List passing props to internal Grid

ivanalejandro0:
This will help users to find all the props they can use on the `List` that are not listed on its docs but on `Grid` docs.


bvaughn:
Thanks!

PR #596, opened 2017-02-25
Use typeof function instead of instanceof Function

rickychien:
Using ```instanceof Function``` would be error prone and always return false if the ```Function``` instance comes from another iframe window,. Replace it by typeof check can be safer.


bvaughn:
Thanks @rickychien! This looks like a solid change. 😄

PR #597, opened 2017-02-26
Update WindowScroller propTypes and docs

amje:
- Make WindowScroller's propTypes **onScroll**, **onResize** optional.
- Add missing prop **scrollElement** in WindowScroller's docs.

### DIFF
--- a/docs/WindowScroller.md
+++ b/docs/WindowScroller.md
@@ -13,6 +13,7 @@ This may change with a future release but for the time being this HOC is should 
 | children | Function | ✓ | Function responsible for rendering children. This function should implement the following signature: `({ height: number, isScrolling: boolean, scrollTop: number }) => PropTypes.element` |
 | onResize | Function |  | Callback to be invoked on-resize; it is passed the following named parameters: `({ height: number })`. | 
 | onScroll | Function |  | Callback to be invoked on-scroll; it is passed the following named parameters: `({ scrollTop: number })`. | 
+| scrollElement | any |  | Element to attach scroll event listeners. Defaults to window. |
 
 ### Public Methods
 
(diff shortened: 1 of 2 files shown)


bvaughn:
Hi @amje,

Why the change to `propTypes`? It shouldn't be necessary. `WindowScroller` provides default, noop functions for those 2 props:

```js
  static defaultProps = {
    onResize: () => {},
    onScroll: () => {}
  };
```

This approach lets me invoke them without checking if they are set or not (see [here](https://github.com/amje/react-virtualized/blob/f8cee5bd1311076e2dda8437e1e46b2619c91b7c/source/WindowScroller/WindowScroller.js#L131-L133) and [here](https://github.com/amje/react-virtualized/blob/f8cee5bd1311076e2dda8437e1e46b2619c91b7c/source/WindowScroller/WindowScroller.js#L70-L72)).

Thanks for updating the docs though. I'll accept that change and revert the other. 😄

amje:
> WindowScroller provides default, noop functions for those 2 props

@bvaughn Thats why I decided to **remove** `.isRequired` on those `propTypes`. Is there a reason to keep them?

bvaughn:
@amje Ah, I think I see the misunderstanding now. Those props are still required- the component just provides values for them by default (if their value is `undefined`). The difference is, you aren't allowed to pass a `null` value.

For example:
```jsx
class Test extends React.Component {
  static propTypes = {
    foo: React.PropTypes.string.isRequired
  };
  static defaultProps = {
    foo: 'bar'
  };
  render () {
    return <div>{this.props.foo}</div>
  }
}

<Test />                // <div>bar</div>
<Test foo={undefined}/> // <div>bar</div>
<Test foo={null}/>      // Warning: Failed prop type:
                        // The prop `foo` is marked as required in `Test`,
                        // but its value is `null`.
```

But more than just the prop-type check, if you were to pass `null` to `onScroll` or `onResize`- `WindowScroller` would actually cause a runtime exception when it tried to call those properties as functions. I could avoid this by wrapping them in a type-check, but it's easier/safer to just mark them as required and provide default values for them.

amje:
@bvaughn Oh, I got it. Thanks for detailed explanation.

bvaughn:
No problem :)

PR #600, opened 2017-02-28
add possibility to inject header-row-renderer + typescript type declarations 

kaoDev:
This pull request enables the table to handle higher order components in the header row. It is needed to make the columns/headers sortable
Additionally I have added a declaration file for typescript to have better declarations than the @types/react-virtualized repo


bvaughn:
Hi @kaoDev,

Thank for submitting a PR. I'll be happy to review it as soon as I get the chance, though that might not be for a couple of days. I'm pretty swamped now.

I did want to make a quick note that I'm not really willing to take on the maintenance burden of TypeScript type declarations. I understand why they are useful- but I'm not a TypeScript user _and_ I have enough of a challenge maintaining this project on my own as it is. Unless they can be auto-generated as part of the build process- they'll need to live somewhere externally so they can be owned by the community. I hope you understand!

kaoDev:
Hey @bvaughn,
good to know you're on it. I will remove the TypeScript declarations and will commit it to the definitely typed repository. (But I can highly recommend to try out TypeScript 😀 )

bvaughn:
Much appreciated 😄

I'm a fan of TypeScript! I've used it in the past. I mostly stick with Flow these days though because it's easier to consolidate my tech stack with work and OSS projects.

source/Table/types.js
```diff
@@ -25,7 +33,7 @@ export type HeaderRendererParams = {
 
 export type RowRendererParams = {
   className: string,
-  columns: Array,
+  columns: any[],
```
bvaughn: Hm, why was this change required?

bvaughn:
Merged with the change I mentioned. Thanks for the contribution! This will be released sometime this morning.

source/Table/Table.js
```diff
@@ -296,18 +312,14 @@ export default class Table extends PureComponent {
...
-              overflow: 'hidden',
-              paddingRight: scrollbarWidth,
-              width: width
-            }}
-          >
-            {this._getRenderedHeaderRow()}
-          </div>
+          this.props.headerRowRenderer({
+            className: cn('ReactVirtualized__Table__headerRow', rowClass),
+            style: rowStyleObject,
```
kaoDev: ok good to know

source/Table/types.js
kaoDev: hmm I think it just slipped in when I implemented the new types, or perhaps I tried to find an easy to implement more specific type information for this array ;). 
This change on it's own, is more or less a matter of taste any[] and Array should be the same in flowType.

bvaughn: Agreed! I just wanted to ask in case there was a difference I was unaware of. 😄

PR #605, opened 2017-03-03
Support for passing event objects thru to various event handlers

paulbrom:
So in addition to the change we made to support horizontal scrolling, we also added support for passing event objects thru the various event handlers, particularly in defaultCellRenderer.  Yes, we can (and did) build our own cellRenderer to accomodate passing these event objects thru (which we needed to check the CTRL and ALT key state for some behaviors we were looking at add), but it seemed to me that it would be nice if the default renderer passed these thru.

### DIFF
--- a/source/Table/Table.js
+++ b/source/Table/Table.js
@@ -403,17 +403,17 @@ export default class Table extends PureComponent {
         ? SortDirection.ASC
         : SortDirection.DESC
 
-      const onClick = () => {
+      const onClick = (evt) => {
         sortEnabled && sort({
           sortBy: dataKey,
           sortDirection: newSortDirection
         })
-        onHeaderClick && onHeaderClick({ columnData, dataKey })
+        onHeaderClick && onHeaderClick({ evt, columnData, dataKey })
       }
 
       const onKeyDown = (event) => {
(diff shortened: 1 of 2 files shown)


bvaughn:
I don't object to passing through the event object as a new named parameter, but I do object to the name chosen. (I don't like abbreviated names. They're too arbitrary.) If you'll rename the parameter to `event` - we're good to go.

I'd also love it if you updated any relevant documentation to list the new parameter. 😄

bvaughn:
FYI I'm going to make my own edit and merge this. Thanks~

source/Table/Table.js
```diff
@@ -403,17 +403,17 @@ export default class Table extends PureComponent {
...
           sortDirection: newSortDirection
         })
-        onHeaderClick && onHeaderClick({ columnData, dataKey })
+        onHeaderClick && onHeaderClick({ evt, columnData, dataKey })
       }
 
       const onKeyDown = (event) => {
         if (event.key === 'Enter' || event.key === ' ') {
-          onClick()
+          onClick(event)
```
bvaughn: This is a bit weird- though I understand why you did it. I guess it's okay since people can ignore the event or check its `type`

bvaughn:
Merged!

PR #607, opened 2017-03-04
Fix aria roles for the table component

### LINKED ISSUE #606: Table accessibility issues
There are a few issues with the aria roles on the current table implementation:
- `role="grid"` should be an attribute on the table element instead of the inner grid
- all table rows (including the header row) should have `role="row"` even if they don't have event handlers like `onRowClick`

See [aria documentation](https://www.w3.org/TR/wai-aria/roles#grid) for more details. Elements with `role="grid"` require descendants with `role="row"`, and elements with `role="rowheader"` require an ancestor with `role="row"` which requires an ancestor with `role="grid"`.

jchen527:
Move role="grid" from the inner grid to the table and add
role="row" to all table rows. Fixes #606.


bvaughn:
This is great! Thanks for taking the time to fix this oversight. 😄

PR #616, opened 2017-03-16
Allow for rubber band overscrolling on iOS

toddtarsi:
Untested resolution for follow up by @maxsalven on #566

Needs testing, and if no one is able to, I may be able to this weekend.

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -1037,15 +1037,6 @@ export default class Grid extends PureComponent {
       return
     }
 
-    // On iOS, we can arrive at negative offsets by swiping past the start
-    // To prevent flicker here, we make playing in the negative offset zone cause nothing to happen.
-    if (event.target.scrollTop < 0) {
-      return
-    }
-
-    // Prevent pointer events from interrupting a smooth scroll
-    this._debounceScrollEnded()
-
     // When this component is shrunk drastically, React dispatches a series of back-to-back scroll events,
     // Gradually converging on a scrollTop that is within the bounds of the new, smaller height.
     // This causes a series of rapid renders that is slow for long lists.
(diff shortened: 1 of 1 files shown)


bvaughn:
Thanks a ton @toddtarsi! I'm going to tweak it slightly to also cover `scrollLeft` but I think this looks fine.

toddtarsi:
:D Glad I can help. Thanks again!

bvaughn:
9.4.0 was just released with this feature. Thank you for your contribution! I listed you in the [9.4.0 release notes](https://github.com/bvaughn/react-virtualized/blob/master/CHANGELOG.md#940).

bvaughn:
Unfortunately this seems to have caused a fairly serious regression that I didn't notice until earlier tonight. You can see it [here](https://bvaughn.github.io/react-virtualized/#/components/Grid), for example. Scroll to the bottom of the `Grid` and then scroll right. `Grid` won't update when you scroll to the right- because the `eventScrollTop` is slightly greater than `scrollTop`. The case I'm seeing is:
```js
this._scrollingContainer.scrollHeight - this._scrollingContainer.clientHeight // 1367
totalRowsHeight - height + scrollbarSize // 1365
```
For now, I'm going to revert this change and we can fix it forward if we find a way around this issue.

bvaughn:
(I realize that my "repro" steps may no longer work once I've released the fix, but you could repro them locally by running this branch locally)

toddtarsi:
Ah, smart. Sorry about causing that @bvaughn. We use list pretty
aggressively where I'm at so i missed that case.

I will try and take a look later, using that example to test.

Perhaps we can reduce the footprint of the constraint to skip rerenders if
we are outside the constraints on one axis and the other axis is unchanged
or outside constraints on both axes.

On Mar 31, 2017 12:49 AM, "Brian Vaughn" <notifications@github.com> wrote:

Unfortunately this seems to have caused a fairly serious regression that I
didn't notice until earlier tonight. You can see it here
<https://bvaughn.github.io/react-virtualized/#/components/Grid>, for
example. Scroll to the bottom of the Grid and then scroll right. Grid won't
update when you scroll to the right- because the eventScrollTop is slightly
greater than scrollTop. The case I'm seeing is:

this._scrollingContainer.scrollHeight -
this._scrollingContainer.clientHeight // 1367
totalRowsHeight - height + scrollbarSize // 1365

For now, I'm going to revert this change and we can fix it forward if we
find a way around this issue.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/bvaughn/react-virtualized/pull/616#issuecomment-290622341>,
or mute the thread

bvaughn:
No worries! I appreciate the work you've done to help on this. 😊

PR #624, opened 2017-03-21
Allow syncing of scrollLeft and scrollTop from props

julianwong94:
MultiGrid manages scrollLeft and scrollTop through state and ignored them through props. Now syncs state if given props.

Modifed componentDidMount and componentWillReceiveProps to sync state if scrollLeft or scrollTop props are passed.

### DIFF
--- a/source/MultiGrid/MultiGrid.jest.js
+++ b/source/MultiGrid/MultiGrid.jest.js
@@ -328,10 +328,7 @@ describe('MultiGrid', () => {
     })
 
     it('should adjust :scrollTop and :scrollLeft when scrollTop and scrollLeft change', () => {
-      render(getMarkup({
-        columnWidth: 50,
-        fixedColumnCount: 2
-      }))
+      render(getMarkup())
       const rendered = findDOMNode(render(getMarkup({
         scrollTop: 750,
         scrollLeft: 900
--- a/source/MultiGrid/MultiGrid.jest.js
+++ b/source/MultiGrid/MultiGrid.jest.js
@@ -300,4 +300,50 @@ describe('MultiGrid', () => {
       expect(bottomRightGrid.style.backgroundColor).toEqual('red')
     })
   })
+  describe('scrollTop and scrollLeft', () => {
+    it('should adjust :scrollLeft for top-right and main grids when scrollLeft is used', () => {
+      const rendered = findDOMNode(render(getMarkup({
+        columnWidth: 50,
+        fixedColumnCount: 2,
+        scrollLeft: 850
+      })))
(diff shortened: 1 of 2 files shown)


source/MultiGrid/MultiGrid.js
```diff
@@ -144,6 +159,23 @@ export default class MultiGrid extends PureComponent {
       this._topGridHeight = null
     }
 
+    if (
+      nextProps.scrollLeft !== this.state.scrollLeft ||
+      nextProps.scrollTop !== this.state.scrollTop
```
bvaughn: This should compare `nextProps` to `this.props` instead of `this.state` right? Otherwise, it will interfere with user scrolling.

This kind of half-controlled prop is troublesome. In hindsight, I should have probably used a different approach here. Maybe a public method you call on the ref instead, I dunno.

source/MultiGrid/MultiGrid.js
```diff
@@ -144,6 +159,23 @@ export default class MultiGrid extends PureComponent {
...
+    ) {
+      const newState = {}
+
+      if (nextProps.scrollLeft != null) {
```
bvaughn: This should probably be `nextProps.scrollLeft !== this.props.scrollLeft` ?

source/MultiGrid/MultiGrid.jest.js
```diff
@@ -300,4 +300,50 @@ describe('MultiGrid', () => {
...
+      const bottomLeftGrid = grids[2]
+      const bottomRightGrid = grids[3]
+      expect(bottomLeftGrid.scrollTop).toEqual(500)
+      expect(bottomRightGrid.scrollTop).toEqual(500)
+    })
+
+    it('should adjust :scrollTop and :scrollLeft when scrollTop and scrollLeft change', () => {
+      render(getMarkup({
+        columnWidth: 50,
+        fixedColumnCount: 2
```
bvaughn: These are default values so you can remove them if you want. Either that, or copy them below. Otherwise- if they weren't default values- then the second render would also change `columnWidth` and `fixedColumnCount` too (which is probably not what you want, since it potentially gives you a false positive)

julianwong94:
I've made the changes to compare nextProps with props instead of state, and fixed the unit test.

bvaughn:
Looks good to me. Thanks :)

bvaughn:
9.4.0 was just released with this feature. Thank you for your contribution! I listed you in the [9.4.0 release notes](https://github.com/bvaughn/react-virtualized/blob/master/CHANGELOG.md#940).

PR #628, opened 2017-03-24
Allow extra props to be passed transparently through the `CellMeasurer`

bpneal:
Firstly the CellMeasurer is a great tool!!, however I have a use-case where I need the props past into the CellMeasurer to be forwarded onto its Child. This is because the props are added via a higher order component, in the parent component's render function (out of my control). Having the ability to transparently pass the props though would solve this problem.

Thanks

✅  Tests Pass Locally
✅  Lint Pass Locally

### DIFF
--- a/source/CellMeasurer/CellMeasurer.js
+++ b/source/CellMeasurer/CellMeasurer.js
@@ -1,5 +1,5 @@
 /** @flow */
-import { PureComponent } from 'react'
+import { PureComponent, cloneElement } from 'react'
 import { findDOMNode } from 'react-dom'
 
 type Props = {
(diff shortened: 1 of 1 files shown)


bvaughn:
Hi @bpneal. I appreciate the pull request. I don't want to introduce additional destructuring and clone steps into an already performance-sensitive area though- particularly for something that seems a pretty unusual use-case.

> This is because the props are added via a higher order component, in the parent component's render function (out of my control).

I don't really understand how something out of your control is passing params to your `cellRenderer` function. But I think you could solve this in your application by adding an intermediary/wrapper component that just reads the added props and passes them through to your `cellRenderer` child.

PR #641, opened 2017-04-04
Add public methods for animation

imadha:
This PR adds the following public methods to `Grid`:
 - `getScrollTop`
 - `getScrollLeft`
 - `scrollToPosition`

The `List` component also gets, which passes thru to its own `Grid`:
 - `getScrollTop`
 - `scrollToPosition`

`Grid.getScrollTop` and `Grid.getScrollLeft` are to be used by components that need to know the calculated offsets in pixels for given row and/or column indexes. This is useful for animating. 

:link: See the comment in this [Plunker](https://plnkr.co/edit/1S9UT7) for a use case: 
```
// TODO This works for numeric rowHeights but it's hacky for function props.
// react-virtualized should expose this as a new List/Grid instance method.
```

`Grid.scrollToPosition` is intended to be used much like `Grid.scrollToRow`, so that a component can call it from its `Grid` instance and scroll to a specific offset. This is also useful for animation, so that a component does not have to reset its own internal state after animating, or `forceUpdate` in order to force a change in props.

Added tests to illustrate how they would work. Commits are atomic. ⚛️ 

💬 Original conversation in Slack where this came from: https://react-virtualized.slack.com/archives/C1XUJH7HB/p1490898846575686

### DIFF
--- a/source/Grid/Grid.jest.js
+++ b/source/Grid/Grid.jest.js
@@ -379,34 +379,20 @@ describe('Grid', () => {
       expect(grid.state.scrollTop).toEqual(50)
     })
 
-    it('should support getScrollLeft() method', () => {
+    it('should support getOffsetForCell() public method', () => {
       const grid = render(getMarkup())
-      const initialScrollLeft = grid.getScrollLeft()
-      const calculatedScrollLeft = grid.getScrollLeft({
-        ...grid.props,
-        scrollToColumn: 24
+      const { scrollLeft, scrollTop } = grid.getOffsetForCell({
+        columnIndex: 24,
+        rowIndex: 49
       })
-
-      expect(initialScrollLeft).toEqual(0)
       // 100 columns * 50 item width = 5,000 total item width
       // 4 columns can be visible at a time and :scrollLeft is initially 0,
       // So the minimum amount of scrolling leaves the 25th item at the right (just scrolled into view).
-      expect(calculatedScrollLeft).toEqual(1050)
-    })
-
-    it('should support getScrollTop() method', () => {
-      const grid = render(getMarkup())
-      const initialScrollTop = grid.getScrollTop()
-      const calculatedScrollTop = grid.getScrollTop({
-        ...grid.props,
-        scrollToRow: 49
-      })
-
-      expect(initialScrollTop).toEqual(0)
+      expect(scrollLeft).toEqual(1050)
       // 100 rows * 20 item height = 2,000 total item height
       // 5 rows can be visible at a time and :scrollTop is initially 0,
       // So the minimum amount of scrolling leaves the 50th item at the bottom (just scrolled into view).
-      expect(calculatedScrollTop).toEqual(900)
+      expect(scrollTop).toEqual(900)
     })
 
     // See issue #565
--- a/source/Grid/Grid.jest.js
+++ b/source/Grid/Grid.jest.js
@@ -366,6 +366,21 @@ describe('Grid', () => {
       expect(grid.state.scrollTop).toEqual(900)
     })
 
+    it('should support getScrollLeft() method', () => {
+      const grid = render(getMarkup())
+      const initialScrollLeft = grid.getScrollLeft()
+      const calculatedScrollLeft = grid.getScrollLeft({
+        ...grid.props,
+        scrollToColumn: 24
+      })
+
+      expect(initialScrollLeft).toEqual(0)
+      // 100 columns * 50 item width = 5,000 total item width
+      // 4 columns can be visible at a time and :scrollLeft is initially 0,
+      // So the minimum amount of scrolling leaves the 25th item at the right (just scrolled into view).
+      expect(calculatedScrollLeft).toEqual(1050)
+    })
+
     it('should support getScrollTop() method', () => {
       const grid = render(getMarkup())
       const initialScrollTop = grid.getScrollTop()
--- a/source/Grid/Grid.jest.js
+++ b/source/Grid/Grid.jest.js
@@ -366,6 +366,21 @@ describe('Grid', () => {
       expect(grid.state.scrollTop).toEqual(900)
     })
 
+    it('should support getScrollTop() method', () => {
+      const grid = render(getMarkup())
+      const initialScrollTop = grid.getScrollTop()
+      const calculatedScrollTop = grid.getScrollTop({
+        ...grid.props,
+        scrollToRow: 49
+      })
+
+      expect(initialScrollTop).toEqual(0)
+      // 100 rows * 20 item height = 2,000 total item height
+      // 5 rows can be visible at a time and :scrollTop is initially 0,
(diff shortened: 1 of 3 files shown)


source/Grid/Grid.js
```diff
@@ -264,8 +264,11 @@ export default class Grid extends PureComponent {
     this._invokeOnGridRenderedHelper = this._invokeOnGridRenderedHelper.bind(this)
     this._onScroll = this._onScroll.bind(this)
     this._setScrollingContainerRef = this._setScrollingContainerRef.bind(this)
+    this._getCalculatedScrollLeft = this._getCalculatedScrollLeft.bind(this)
     this._updateScrollLeftForScrollToColumn = this._updateScrollLeftForScrollToColumn.bind(this)
+    this._getCalculatedScrollTop = this._getCalculatedScrollTop.bind(this)
     this._updateScrollTopForScrollToRow = this._updateScrollTopForScrollToRow.bind(this)
+    this._setScrollPosition = this._setScrollPosition.bind(this)
```
bvaughn: I don't see why these 3 new functions are bound. They're internal; they're only called from other methods that have the correct context (`this`). It doesn't seem like this binding is necessary.

The only reason methods like `_onScroll` and `_setScrollingContainerRef` are bound this way is that they're passed as properties to React and may not be called with the correct context.

While writing this comment I noticed that `_updateScrollLeftForScrollToColumn` and `_updateScrollTopForScrollToRow` no longer need to be bound since they aren't passed directly to `updateScrollIndexHelper` anymore. 😄

source/Grid/Grid.js
```diff
@@ -381,6 +384,31 @@ export default class Grid extends PureComponent {
     })
   }
 
+  /**
+   * Ensure offset position is visible
+   * Useful for animating position changes
+   */
+  scrollToPosition ({
+    scrollLeft = 0,
+    scrollTop = 0
```
bvaughn: I don't think `scrollLeft` or `scrollTop` should have default values of 0. That will make them _required_ parameters. The `_setScrollPosition` helper method allows either param to be omitted, in which case the current value in `state` won't be modified. This seems convenient if you want to scroll one axis of `Grid` without affecting the other.

source/Grid/Grid.js
```diff
@@ -381,6 +384,31 @@ export default class Grid extends PureComponent {
...
+  } = {}) {
+    this._setScrollPosition({ scrollLeft, scrollTop })
+  }
+
+  /**
+   * Gets a calculated value to be used for `scrollLeft`
+   */
+  getScrollLeft (props = this.props, state = this.state) {
```
bvaughn: What's the reason for these 2 methods to accept `props` and `state` as parameters instead of just using the current values (`this.props` and `this.state`)? External components really shouldn't know or care about or touch `instance.state`. And passing in `props` seems like it would make the method more complicated to use.

I was imagining something more like `getScrollLeft(columnIndex: number): number`- or maybe even `getOffsetForCell({ columnIndex, rowIndex }) => { scrollLeft, scrollTop }`.

Were you trying to allow this method to be called during `componentWillReceiveProps` / `componentWillUpdate` with a newer version of props than `Grid` currently had? I'm wary of that- in part because of the disconnect between externally-visible `props` and internally managed `state`.

**Edit for clarity**: It's okay for methods like `_getCalculatedScrollLeft` to accept params for `props` and `state` since they're private/internal methods. I just don't think we should expose this publicly.

source/List/List.js
```diff
@@ -122,6 +122,11 @@ export default class List extends PureComponent {
     })
   }
 
+  /** See Grid#getScrollTop */
+  getScrollTop (props = this.props) {
```
bvaughn: Same comment as above RE `Grid` ^

source/Grid/Grid.js
imadha: They accept props and state mainly because of laziness and not breaking out the parts I think we _actually_ care about. 😅 

 The main reason I see this being used for is computing a "target" `scrollTop` (or `scrollLeft`) based on row or column index. Really, the only things I need to pass are the indexes and the alignment.

```
getOffsetForCell({ columnIndex, rowIndex, scrollToAlignment }) => { scrollLeft, ScrollTop }
```

seems perfectly reasonable.

source/Grid/Grid.js
imadha: Oh that's a really good point. I was imagining you'd always pass in your x and y here but it's perfectly reasonable to only want to move on one access.

imadha:
Thanks for taking the time to review my changes @bvaughn! Don't hesitate to let me know if you'd like anything else updated.

On this note:

> I'm a little wary of adding more public instance methods for things that could be controlled by props because it adds some maintenance complexity, but I understand why you might want these to avoid having to explicitly un-set properties after an animation completes.

I fully understand this, and the main reason for adding `scrollToPosition` is exactly what you said, basically avoid unsetting props in the same way one might use `scrollToCell`. If that's a pattern you'd rather implementers handle, then let's retract that change.

source/Grid/Grid.jest.js
```diff
@@ -379,34 +379,20 @@ describe('Grid', () => {
...
-    it('should support getScrollLeft() method', () => {
+    it('should support getOffsetForCell() public method', () => {
       const grid = render(getMarkup())
-      const initialScrollLeft = grid.getScrollLeft()
-      const calculatedScrollLeft = grid.getScrollLeft({
-        ...grid.props,
-        scrollToColumn: 24
+      const { scrollLeft, scrollTop } = grid.getOffsetForCell({
+        columnIndex: 24,
+        rowIndex: 49
```
bvaughn: Much nicer :+1:

source/Grid/Grid.js
```diff
@@ -389,24 +384,33 @@ export default class Grid extends PureComponent {
...
-  getScrollLeft (props = this.props, state = this.state) {
-    return this._getCalculatedScrollLeft(props, state) || 0;
-  }
+  getOffsetForCell ({
+    columnIndex,
+    rowIndex,
+    scrollToAlignment = this.props.scrollToAlignment
+  } = {}) {
+    const scrollToColumn = columnIndex >= 0 ? columnIndex : this.props.scrollToColumn
+    const scrollToRow = rowIndex >= 0 ? rowIndex : this.props.scrollToRow
```
bvaughn: Tiny nit: You could set the default `columnIndex` and `rowIndex` values in the same way as you do the `scrollToAlignment` for consistency.

bvaughn:
I made a couple of small tweaks (eg renamed `scrollToAlignment` parameter to `alignment` and added a few more tests and docs. Otherwise I'm merging this. Thanks again!

bvaughn:
Releasing as 9.7.0 now. Gave you credit in the changelog. Thanks!

imadha:
image

bvaughn:
You are welcome!

I look forward to seeing what you build with these new methods. Do share!

PR #644, opened 2017-04-05
Enable autoWidth on Grid in combination with horizontal WindowScroller

maxnowack:
Adds the ability to use `WindowScroller` to detect horizontal scrolling. This allows the usage of `autoWidth` on the `Grid` component.
I'm not sure why these two cases weren't already implemented.

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -45,6 +45,12 @@ export default class Grid extends PureComponent {
     autoHeight: PropTypes.bool,
 
     /**
+     * Removes fixed width from the scrollingContainer so that the total width
+     * of rows can stretch the window. Intended for use with WindowScroller
+     */
+    autoWidth: PropTypes.bool,
+
+    /**
      * Responsible for rendering a cell given an row and column index.
      * Should implement the following interface: ({ columnIndex: number, rowIndex: number }): PropTypes.node
      */
@@ -423,7 +429,7 @@ export default class Grid extends PureComponent {
    * 1) New scroll-to-cell props have been set
    */
   componentDidUpdate (prevProps, prevState) {
-    const { autoHeight, columnCount, height, rowCount, scrollToAlignment, scrollToColumn, scrollToRow, width } = this.props
+    const { autoHeight, autoWidth, columnCount, height, rowCount, scrollToAlignment, scrollToColumn, scrollToRow, width } = this.props
     const { scrollLeft, scrollPositionChangeReason, scrollTop } = this.state
 
     // If cell sizes have been invalidated (eg we are using CellMeasurer) then reset cached positions.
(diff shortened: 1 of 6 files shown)


bvaughn:
> I'm not sure why these two cases weren't already implemented.

Because no one has needed it before now and it takes time and effort to implement things. 😛

I'm not really convinced this is a useful feature. Why would you want a `Grid` to have a width wider than the viewport? Vertical "auto" is useful for building UIs with fixed headers but I don't think it's common for the same type of UIs to be horizontal.

bvaughn:
To be clear~ I appreciate the PR! 😄

I'm just not sure about this feature.

maxnowack:
This wasn't meant as a reproach 😉
I was looking at the code, saw all the height stuff and wondered about where the width is.

I'm building a kanban/trello like application (teamgridapp.com). At this time, I'm kicking out our legacy UI framework and rewrite the views in React. Our main view shows all team members in horizontal aligned lists, which I want to virtualize. There're some reasons (styling, sortable, etc.), why these lists don't have a own scrollable div and use one of the parents for scrolling instead.
I've tried to implement virtualize using my PR and it works very well.

source/Grid/Grid.js
```diff
@@ -1104,6 +1122,9 @@ export default class Grid extends PureComponent {
       if (!autoHeight) {
         newState.scrollTop = scrollTop
       }
+      if (!autoWidth) {
+        newState.scrollLeft = scrollLeft
```
bvaughn: Same type of comment here. `scrollLeft` has already been set above, inline, within `newState`.

source/WindowScroller/WindowScroller.jest.js
```diff
@@ -241,44 +253,51 @@ describe('WindowScroller', () => {
...
         }
       }))
 
       expect(windowScroller._positionFromTop).toBe(100)
+      expect(windowScroller._positionFromLeft).toBe(100)
 
       mockGetBoundingClientRectForHeader({
-        height: 200
+        height: 200,
+        width: 200
```
bvaughn: I usually like to use different values for width/height or scrollLeft/scrollTop to avoid possibly missing a bug where I transpose the 2 internally.

source/WindowScroller/WindowScroller.js
```diff
@@ -32,12 +32,13 @@ export default class WindowScroller extends PureComponent {
...
-    const height = typeof window !== 'undefined'
-      ? getHeight(props.scrollElement || window)
-      : 0
+    const { width, height } = typeof window !== 'undefined'
+      ? getDimensions(props.scrollElement || window)
+      : { width: 0, height: 0 }
 
     this.state = {
       height,
+      width,
```
bvaughn: `state` should probably initialize `scrollLeft` to 0 as well

bvaughn:
No worries 😄 Sorry if my reaction seemed defensive. Harder to communicate when sleepy.

bvaughn:
9.6.0 release just went out with this.

maxnowack:
Wow, this was fast! Thank you for applying the changes and all the great work with this package!
I'm deeply impressed of how you manage your open source projects and how fast you react to pull requests and issues.
Again: Thank you!

PR #649, opened 2017-04-11
Clear cell cache when external scrolling ends

leoasis:
With the fix in 9.5.0 for using the cell cache in`WindowScroller` (https://github.com/bvaughn/react-virtualized/blob/master/CHANGELOG.md#950), another bug was unveiled:

Since the `Grid` when used with `WindowScroller` does not emit `onScroll` events when the user scrolls, the cell cache is never cleaned up because the `_debounceScrollEnded` function is never called.

This PR fixes that by detecting if `props.isScrolling` changed from `true` to `false`, and call `this._debounceScrollEnded()` in that case.

### DIFF
--- a/source/Grid/Grid.jest.js
+++ b/source/Grid/Grid.jest.js
@@ -1383,6 +1383,45 @@ describe('Grid', () => {
       done()
     })
 
+    it('should clear cache once :isScrolling via props is false', async () => {
+      const cellRendererCalls = []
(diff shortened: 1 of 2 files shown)


bvaughn:
Thanks for this contribution! Released as 9.7.3

leoasis:
Thanks for accepting it and for this library!

PR #652, opened 2017-04-12
Add onRowContextMenu callback

difelice:
## Add onRowContextMenu callback to Table
Add the ability to handle the mouse context-menu event for each row.

### DIFF
--- a/source/Table/Table.jest.js
+++ b/source/Table/Table.jest.js
@@ -574,6 +574,20 @@ describe('Table', () => {
     })
   })
 
+  describe('onRowContextMenu', () => {
+    it.only('should call :onRowContextMenu with the correct :rowIndex when a row is clicked', () => {
+      const onRowContextMenu = jest.fn()
+      const rendered = findDOMNode(render(getMarkup({
+        onRowContextMenu
+      })))
+      const rows = rendered.querySelectorAll('.ReactVirtualized__Table__row')
+      Simulate.contextMenu(rows[0])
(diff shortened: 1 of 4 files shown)


bvaughn:
Hi @difelice! Thanks for contributing and for taking the time to add tests too! 😄

In the past other people have asked for more row-level events and I generally push back b'c it's hard to draw the line once I start adding more of them. I think the ones that are currently available are by far the most common and additional ones- like contextmenu- should probably stay in application code to avoid adding weight to the `Table` component. (Otherwise, in the extreme, I'd end up adding a prop for each possible event a `<div>` might dispatch.)

In this case you can implement this event yourself pretty easily using the `rowRenderer` prop:

```jsx
const { defaultTableRowRenderer, Column, Table } = ReactVirtualized;

function onRowContextMenu ({ event, index, rowData }) {
  console.log('onContextMenu()', { index, rowData })
}

function rowRenderer (props) {
  const { index, rowData } = props

  const row = defaultTableRowRenderer(props)

  return React.cloneElement(row, {
    onContextMenu: (event) => onRowContextMenu({ event, index, rowData })
  })
}

function renderTable(props) {
  return (
    <Table
      {...props}
      rowRenderer={rowRenderer}
    />
  )
);
```

Here's a [Plnkr](https://plnkr.co/edit/cOkYF4?p=preview) for you.

difelice:
Makes sense. Thanks for the example!

bvaughn:
No problem! Thanks for understanding.

PR #662, opened 2017-04-27
columnIndex added to table column's cellRenderer function

BamaBoy:
Actually, this thing started, with `CellMeasurer` ☺️. 
`Table`'s [demo](http://localhost:3001/#/components/CellMeasurer) works fine if we have only one dynamic column.
But if we want to have two or more dynamic columns, something like [this](https://www.webpackbin.com/bins/-Kik5ZirudqCXo_U_7JE) happens. 

We have to pass `columnIndex` to `cellRenderer`, and then to `CellMeasurer`, to make it work.
Of course we can do it manually, but it would be nice if `cellRenderer` function had access to `columnIndex`.

### DIFF
--- a/source/Table/Table.jest.js
+++ b/source/Table/Table.jest.js
@@ -259,7 +259,7 @@ describe('Table', () => {
   describe('forceUpdateGrid', () => {
     it('should refresh inner Grid content when called', () => {
       let marker = 'a'
-      function cellRenderer ({ cellData, columnData, dataKey, rowData, rowIndex }) {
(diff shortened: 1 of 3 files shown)


bvaughn:
This seems like a reasonable change. Can confirm that `Table` with multi-columns using `CellMeasurer` works fine with this. Thanks for submitting the PR!

bvaughn:
Released as 9.7.4

BamaBoy:
Thanks!

Issue #665, opened 2017-04-28
sub-links in docs page don't work on refresh


djkirby:
ex:

- Go to https://bvaughn.github.io/react-virtualized
- Click InfiniteLoader tab (https://bvaughn.github.io/react-virtualized/#/components/InfiniteLoader)
- Refresh
- It brings me to https://bvaughn.github.io/react-virtualized/#/components/List

bvaughn:
Thanks for pointing this out. Was probably broken by the recent update from RR 4 beta to final.

Going to tag it as a good first bug in case anyone wants to contribute a fix. 😄

joecritch:
I'd love to take this one! Just going to set up my fork.

bvaughn:
Nice! It's all yours 😁

joecritch:
@bvaughn @djkirby Done! Although not sure if I should contribute again with this number... #666

bvaughn:
Haha! Thanks!

PR #666, opened 2017-04-28
Changed exactly prop on Route component to exact for RRV4 final

joecritch:
React Router V4 changed the `exactly` prop to `exact` for path matches. This PR mirrors that change to in `demo/Application.js`

Refs #665


bvaughn:
Woo hoo!

PR #669, opened 2017-05-01
allow skipping of prerendering when using CellMeasurer

bolandrm:
I am using CellMeasurer to render a Grid with unknown column sizes.

The Grid has many rows, and by default it renders all of them so that CellMeasurer can determine the width of the largest cells for each column.  Because it renders all of the rows, the performance is poor.

I don't really care if the cell widths change slightly as I scroll down and larger cells are found, so i've added an option to skip this prerendering of all the cells to dramatically improve performance.

### DIFF
--- a/docs/Grid.md
+++ b/docs/Grid.md
@@ -15,6 +15,7 @@ A windowed grid of elements. `Grid` only renders cells necessary to fill itself 
 | columnWidth | Number or Function | ✓ | Either a fixed column width (number) or a function that returns the width of a column given its index: `({ index: number }): number` |
 | containerStyle | Object |  | Optional custom inline style to attach to inner cell-container element. |
 | deferredMeasurementCache | `CellMeasurer` |  | If CellMeasurer is used to measure this Grid's children, this should be a pointer to its CellMeasurerCache. A shared CellMeasurerCache reference enables Grid and CellMeasurer to share measurement data. |
+| deferredMeasurementBatchAll | Boolean |  | When Using deferredMeasurementCache, set this to false to skip pre-rendering of all cells for size calculations.
 | estimatedColumnSize | Number |  | Used to estimate the total width of a `Grid` before all of its columns have actually been measured. The estimated total width is adjusted as columns are rendered. |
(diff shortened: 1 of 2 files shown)


bvaughn:
I don't know how I feel about this change. This is already a pretty complex area (deferred measurements) and adding more variations to support makes me uneasy. The actual effects of this could be pretty weird too. Imagine if you're scrolling left-to-right and suddenly rows start shifting down and off of your screen until what you're looking at isn't even the same rows you started scrolling from. That's the potential case that this could introduce.

bolandrm:
The use case for it would be mostly around data that you know has similar, but not exactly the same size.

bvaughn:
I understand this _specific_ use case for it, but I'm concerned about adding support to the broader lib. Once it's in, I'm committed to supporting it.

bolandrm:
Understood!  If you decided it's best to not add it to the library, is there some other way I could override this behavior without maintaining a fork?

bvaughn:
I'll think on it a little bit. My initial reaction to this feature is negative but sometimes that changes.

There is not a great way for you to work around this that I know of.

bvaughn:
Hey @bolandrm. I really appreciate the PR and the explanation you provided. I think I'm going to pass on this feature for now though. I'm reluctant to try to add any more complexity to this particular area of react-virtualized. I hope you understand!

I _believe_ you might be able to work around this (and achieve the same behavior as you're going for) by doing the following- although I haven't tested to verify it works:

1. Don't specify a `deferredMeasurementCache` prop on your `Grid` (so it doesn't try to batch cells).
1. Be sure to set `width` or `height` (whichever you're measuring) to "auto" manually in your `cellRenderer` the first time a given cell is rendered. (You can check `CellMeasurerCache` to see if it's been measured.) This will mimic [what `defaultCellRangeRenderer` does normally](https://github.com/bvaughn/react-virtualized/blob/770e29430b5997b64bb26503ba1ff2ae172c8f33/source/Grid/defaultCellRangeRenderer.js#L66-L75) to enable `CellMeasurer` to get accurate measurements.

PR #674, opened 2017-05-05
use keyMapper of deferredMeasurementCache to cache styles

maxnowack:
The defaultCellRangeRenderer should use the same keyMapper as deferredMeasurementCache to ensure that the cached styles are applied to the correct elements.


maxnowack:
This doesn't fix anything. Instead I have to call `recomputeGridSize` if the order of the items has changed.

bvaughn:
Yeah, the cell measurements are cached with the key mapper so cells can move around (or have things added/removed in between) without losing their sizes. (Size doesn't change when a cell is moved.) But position info is specific to the order of cells so there's no equivalent mapper.

Issue #660, opened 2017-04-26
MultiGrid with fixedRowCount is rendering incorrect width.


marcelmokos:
Using MultiGrid when the render with fixedRowCount then some of the columns do not render with width from the CellMeasurerCache.

Basically, the column has `width: auto;` instead of for example`width: 150;`. This problem is not showing every time more likely is the problem visible on larger tables and the problem and row on which it occurs is random as well. (Tested on mac in chrome, and firefox).

Am I missing something that can cause this problem? 

![widthauto](https://cloud.githubusercontent.com/assets/6388074/25433247/fb228cec-2a87-11e7-889e-6ceb37592894.png)

bvaughn:
Hi @marcelmokos,

Unfortunately I need more than a screenshot to help diagnose a problem.

Please provide a Plnkr that shows the issue you're encountering and I'll be happy to take a look. In the meanwhile I'm going to close this issue since it's not really actionable.

> ### Are you reporting a bug or runtime error?
> Please include either a failing unit test or a Plnkr demonstrating the bug you are reporting. You can start by forking this Plnk! https://plnkr.co/edit/6syKo8cx3RfoO96hXFT1
>
> You may also find the online Babel tool (https://babeljs.io/repl/) quite helpful if you wish to use ES6/ES7 syntax not yet supported by the browser you are using.

In the meanwhile you may also want to check out [this demo page](https://bvaughn.github.io/react-virtualized/#/components/CellMeasurer)- specifically [the source for the page](https://github.com/bvaughn/react-virtualized/blob/master/source/CellMeasurer/CellMeasurer.DynamiWidthMultiGrid.example.js)- to for an example of using `CellMeasurer` with `MultiGrid`.

PR #675, opened 2017-05-06
MultiGrid with fixedRowCount is rendering incorrect width

marcelmokos:
Issue: https://github.com/bvaughn/react-virtualized/issues/660

I am first time contributor to react-virtualized. Please excuse me for not providing a code example. I can create one if it will be necessary for accepting PR.

CellMeasurer during rerendering of the table is setting width: "auto" to the table cells in measure methods. This width "auto" is causing incorrect cell rendering mostly visible on cell backgrounds.

widthauto

This fix of CellMeasurer is remembering width and height of the cell before measuring. 
Measuring is setting styles for width and height to "auto" prior measuring. 
The fix is setting remembered width and height back after measuring.

### DIFF
--- a/source/CellMeasurer/CellMeasurer.jest.js
+++ b/source/CellMeasurer/CellMeasurer.jest.js
@@ -223,6 +223,39 @@ describe('CellMeasurer', () => {
       (params) => <div style={{ width: 100, height: 30 }}></div>
     )
 
+    let measurer;
+    const node = findDOMNode(render(
+      <CellMeasurer
+        ref={(ref) => { measurer = ref }}
+        cache={cache}
+        columnIndex={0}
+        parent={parent}
+        rowIndex={0}
+        style={{}}
+      >
+        {child}
+      </CellMeasurer>
+    ))
+
+    const { height, width } = measurer._getCellMeasures(node)
+    expect(height).toBeGreaterThan(0)
+    expect(width).toBeGreaterThan(0)
+
+    expect(node.style.height).toBe('auto')
+    expect(node.style.width).not.toBe('auto')
+  })
+
+  // See issue #660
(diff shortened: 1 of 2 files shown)


bvaughn:
Hey @marcelmokos! Welcome 😄

I'm not sure I understand the need for this fix. `CellMeasurer` only sets width to "auto" if its cache is configured not to have a fixed-width- and it only does so initially (when the cell hasn't yet been measured). Once the cell(s) have been measured, the list/table/grid is _always_ re-rendered (because of how positions are calculated and cached) at which point the "auto" value _won't_ be set again because size is already known.

I'd need to know more (probably a Plnkr that repros the unexpected issue you're seeing, that I could actually take a look at) before I want to change this code because it seems like the current implementation is reasonable.

marcelmokos:
Thanks, I will create a public example.

marcelmokos:
Thanks for your awesome library. 
I was able to create examples.

**This example uses not modified CellMeasurer, CellMeasurerCache**
I am using the width of the component to determine if the cache should be invalidated. When the cache invalidates items are rendered incorrectly.
[Edit react-virtualized](https://codesandbox.io/s/v2m3EQyY5)
react-virtualized

**This example uses modified CellMeasurer, CellMeasurerCache that allows resizing of the table when width of the table changes**
This PR is not about CellMeasurerCache change am just showing why I am proposing the change of CellMeasurer. With proposed change in CellMeasurer, I am no longer experiencing the issue after deleting cache.
[Edit react-virtualized](https://codesandbox.io/s/RgzAYAMRq)

resizible-react-virtualized

bvaughn:
Thanks for the demos. 😄

bvaughn:
This has been released via 9.7.6

PR #681, opened 2017-05-17
Fix aria roles for table headers and cells

### LINKED ISSUE #680: a11y: Table header and cell roles
The table implementation is missing a few other ARIA roles:

- Table headers should have `role="columnheader"`, an `id` to associate the column's cells, and `aria-sort` if the column is sorted
- Table cells should have `role="gridcell"` and `aria-describedby` that points to the associated column header

ARIA documentation: [columnheader](https://www.w3.org/TR/wai-aria/roles#columnheader) and [gridcell](https://www.w3.org/TR/wai-aria/roles#gridcell).

jchen527:
Fixes #680.
- Add id, aria role, and aria sort to table column headers
- Add aria describedby and role to table cells


bvaughn:
I'm traveling at the moment and it's late- so I'll give this a real review later. But at a glance it looks great. Thanks!

bvaughn:
Thanks!

bvaughn:
This has been released via 9.7.6

PR #686, opened 2017-05-24
Add a comment about pointerEvents

jgoux:
I added a mention about pointerEvents for highly interactive rows / items. :)

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -95,7 +95,8 @@ function rowRenderer ({
   //    top
   //    height
   //    width
-  // You can add additional class names or style properties as you would like.
+  // You can add additional class names or style properties as you would like,
+  // for example adding the property pointerEvents: 'auto' can be useful to get immediate interactivity.
   // Key is also required by React to more efficiently manage the array of rows.
   return (
     <div


bvaughn:
Thanks for starting the ball rolling on this! 😁 

My initial reaction is that the docs location is probably too deep. Since `pointer-events` are disabled for the entire scroll-container, in `Grid`, it might make more sense to override them there (eg `<Grid containerStyle={...}>`). If not- we should at least mention this style in the `Grid` docs (and maybe `Table` as well) since it impacts more than just `List`.

I also think the wording, "_immediate interactivity_", is too vague. I haven't thought about this much but it seems like we should explain briefly that RV- and every other windowing lib, eg [react-infinite's `Infinite`](https://github.com/seatgeek/react-infinite/blob/6c38c2ef3c473394a57c80c06339754e621f2909/src/react-infinite.jsx#L420)and [fixed-data-table's `FixedDataTableBufferedRows`](https://github.com/facebook/fixed-data-table/blob/5cbf487dd038c2b741cd0298922c1c30dacb38d7/src/FixedDataTableBufferedRows.react.js#L161)- disable `pointer-events` while a scroll is in progress to improve performance. A few milliseconds after scrolling stops, they are re-enabled. This delay defaults to 150ms but is customizable via the `scrollingResetTimeInterval` as noted in the [`Grid` docs](https://github.com/bvaughn/react-virtualized/blob/master/docs/Grid.md):

> **scrollingResetTimeInterval** `Number` _Wait this amount of time after the last scroll event before resetting Grid pointer-events; defaults to 150ms._

Once we explain this concept and why it's done, we could add a note about how users can disable this if they want by overriding the style- either on the container level _or_ the row/cell level.

What do you think? 😁

bvaughn:
Closing for now. Feel free to re-open if you would like to address the feedback above. 😄

Issue #687, opened 2017-05-28
Feature request: Externalize ArrowKeyStepper state


mking-clari:
I would like to manage the scrollToColumn/scrollToRow state in Redux. I see that I can set the ArrowKeyStepper's scrollToColumn/scrollToRow via its props. In addition, I would like an onChange function to capture scrollToColumn/scrollToRow changes.

I think [ReactMeasure](https://github.com/souporserious/react-measure) is a good example of a component that provides both options (function child vs onMeasure).

bvaughn:
`ArrowKeyStepper` works with a child function, so you already have access to the scrolled-to row and column.

```jsx
  <ArrowKeyStepper
    columnCount={columnCount}
    rowCount={rowCount}
  >
    {({ onSectionRendered, scrollToColumn, scrollToRow }) => {
      // Do whatever you want with scrollToColumn and scrollToRow here
      return (
        <Grid
          scrollToColumn={scrollToColumn}
          scrollToRow={scrollToRow}
          onSectionRendered={onSectionRendered}
          {...otherGridProps}
        />
      );
    }}
  </ArrowKeyStepper>
```

mking-clari:
If I call a Redux action there, that causes a setState function to be called within render, which results in a React warning. I'll create an example repo to show it

mking-clari:
Here is the issue (let me know if you know a better way to transfer the child args to Redux).

<img width="978" alt="screen shot 2017-05-28 at 10 07 10 am" src="https://cloud.githubusercontent.com/assets/15120143/26530655/dbd4a2a2-438d-11e7-9f16-9198415a9b59.png">

Repo: https://github.com/mking-clari/stepper-warning

bvaughn:
> If I call a Redux action there, that causes a setState function to be called within render, which results in a React warning. I'll create an example repo to show it

You could dispatch your action on the next tick to avoid this.

bvaughn:
Anyway, I'm pretty swamped at the moment so I haven't been taking on feature requests for this lib. Happy to review a PR if you'd like to submit one though. 😄

mking-clari:
I'll close this for now. If I have time I'll open a PR. Thanks for taking the time to review!

bvaughn:
Sounds reasonable. Thanks for understanding!

PR #688, opened 2017-05-28
Allow stepper to be used as a controlled component

mking-clari:
Implements #687
* Provides the ability to use `ArrowKeyStepper` as a controlled component
* My specific use case is the ability to select cells via click. Generally this allows the user to provide a single source of truth for all selection methods (include the `ArrowKeyStepper`'s state).
* When using `ArrowKeyStepper` as a controlled component:
  * Provide `onScrollToChange({ scrollToColumn, scrollToRow })` as a prop
  * Provide `scrollToColumn` as a prop
  * Provide `scrollToRow` as a prop
  * If `ArrowKeyStepper` is being used as a controlled component, it does not call `setState` for performance reasons (instead it uses the props directly).

PR checklist
* Tests are passing
* Lint is passing
* I added two new test cases: 
  * Check that `onScrollToChange` is called when key down occurs
  * Check that `onScrollToChange` is not called when props are updated (if the change occurs via props, the source of truth already knows about the change)
* I added an additional demo setting, `Enable click selection`. I have enabled this setting by default in the demo because I think users normally want to be able to click to select.

Let me know your feedback. I will be happy to make changes.

### DIFF
--- a/source/ArrowKeyStepper/ArrowKeyStepper.js
+++ b/source/ArrowKeyStepper/ArrowKeyStepper.js
@@ -8,16 +8,18 @@ import React, { PureComponent } from 'react'
 export default class ArrowKeyStepper extends PureComponent {
   static defaultProps = {
     disabled: false,
+    isControlled: false,
     mode: 'edges',
     scrollToColumn: 0,
     scrollToRow: 0
   };
 
   static propTypes = {
-    children: PropTypes.oneOfType([PropTypes.func, PropTypes.node]).isRequired,
+    children: PropTypes.func.isRequired,
     className: PropTypes.string,
     columnCount: PropTypes.number.isRequired,
     disabled: PropTypes.bool.isRequired,
+    isControlled: PropTypes.bool.isRequired,
     mode: PropTypes.oneOf(['cells', 'edges']),
     onScrollToChange: PropTypes.func,
     rowCount: PropTypes.number.isRequired,
@@ -43,7 +45,7 @@ export default class ArrowKeyStepper extends PureComponent {
   }
 
   componentWillReceiveProps (nextProps) {
-    if (!this._isComponentState()) {
+    if (this.props.isControlled) {
       return
     }
 
@@ -81,7 +83,7 @@ export default class ArrowKeyStepper extends PureComponent {
 
   render () {
     const { className, children } = this.props
-    const { scrollToColumn, scrollToRow } = this._getRelevantState()
+    const { scrollToColumn, scrollToRow } = this._getScrollState()
 
     return (
       <div
@@ -97,10 +99,6 @@ export default class ArrowKeyStepper extends PureComponent {
     )
   }
 
-  _isComponentState () {
-    return typeof this.props.onScrollToChange !== 'function'
-  }
-
   _onKeyDown (event) {
     const { columnCount, disabled, mode, rowCount } = this.props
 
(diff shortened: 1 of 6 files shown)


source/ArrowKeyStepper/ArrowKeyStepper.js
```diff
@@ -92,6 +97,10 @@ export default class ArrowKeyStepper extends PureComponent {
     )
   }
 
+  _isComponentState () {
```
bvaughn: nit: Maybe `useComponentState` ?

source/ArrowKeyStepper/ArrowKeyStepper.js
```diff
@@ -147,4 +156,16 @@ export default class ArrowKeyStepper extends PureComponent {
     this._rowStartIndex = rowStartIndex
     this._rowStopIndex = rowStopIndex
   }
+
+  _getRelevantState () {
+    return this._isComponentState() ? this.state : this.props
+  }
+
+  _setRelevantState ({ scrollToColumn, scrollToRow }) {
```
bvaughn: nit: Maybe `updateScrollState` ?

source/ArrowKeyStepper/ArrowKeyStepper.js
```diff
@@ -147,4 +156,16 @@ export default class ArrowKeyStepper extends PureComponent {
     this._rowStartIndex = rowStartIndex
     this._rowStopIndex = rowStopIndex
   }
+
+  _getRelevantState () {
```
bvaughn: nit: Maybe `getScrollState` ?

yarn.lock
```diff
@@ -3619,13 +3619,13 @@ longest@^1.0.1:
...
   resolved "https://registry.yarnpkg.com/longest/-/longest-1.0.1.tgz#30a0b2da38f73770e8294a0d22e6625ed77d0097"
 
-loose-envify@^1.0.0, loose-envify@^1.1.0, loose-envify@^1.2.0, loose-envify@^1.3.1:
+loose-envify@^1.0.0, loose-envify@^1.2.0, loose-envify@^1.3.1:
   version "1.3.1"
   resolved "https://registry.yarnpkg.com/loose-envify/-/loose-envify-1.3.1.tgz#d1a8ad33fa9ce0e713d65fdd0ac8b748d478c848"
   dependencies:
     js-tokens "^3.0.0"
 
-loose-envify@^1.3.0:
```
bvaughn: Probably fine, but can we drop the `yarn.lock` changes from this PR since it's not actually introducing any new deps?

source/ArrowKeyStepper/ArrowKeyStepper.js
```diff
@@ -14,11 +14,12 @@ export default class ArrowKeyStepper extends PureComponent {
   };
 
   static propTypes = {
-    children: PropTypes.func.isRequired,
+    children: PropTypes.oneOfType([PropTypes.func, PropTypes.node]).isRequired,
```
bvaughn: This change seems wrong. If you pass a node, wouldn't things break now since the `render` function expect `children` to always be a function?

source/ArrowKeyStepper/ArrowKeyStepper.js
```diff
@@ -14,11 +14,12 @@ export default class ArrowKeyStepper extends PureComponent {
...
     className: PropTypes.string,
     columnCount: PropTypes.number.isRequired,
     disabled: PropTypes.bool.isRequired,
     mode: PropTypes.oneOf(['cells', 'edges']),
+    onScrollToChange: PropTypes.func,
```
bvaughn: Seems like it's at least feasible that someone _might_ want to observe scroll changes without being in controlled mode. I wonder if controlled mode should be a more explicit flag (eg a `isControlled` boolean prop or something)?

mking-clari:
@bvaughn Thanks for your feedback. Updated per your review.

bvaughn:

```
```

bvaughn:
Haven't heard anything negative about the RC so I just released 9.8.0

PR #689, opened 2017-05-29
Prevent calling setState on an unmounted WindowScroller component

liorbrauer:
This is a bit of an edge case, when rendering several WindowScroller components on the same page inside a virtual list, the scroll listener can sometimes call setState on an unmounted WindowScroller component (that has been scrolled out of view in the virtual list), triggering react's warning.

This PR prevents calling setState in this scenario if the component has been unmounted.

### DIFF
--- a/source/WindowScroller/WindowScroller.js
+++ b/source/WindowScroller/WindowScroller.js
@@ -88,6 +88,8 @@ export default class WindowScroller extends PureComponent {
     registerScrollListener(this, scrollElement)
 
     window.addEventListener('resize', this._onResize, false)
+    
+    this._isMounted = true
   }
 
   componentWillReceiveProps (nextProps) {
@@ -106,6 +108,8 @@ export default class WindowScroller extends PureComponent {
     unregisterScrollListener(this, this.props.scrollElement || window)
 
     window.removeEventListener('resize', this._onResize, false)
+    
+    this._isMounted = false
   }
 
   render () {
@@ -127,6 +131,8 @@ export default class WindowScroller extends PureComponent {
 
   // Referenced by utils/onScroll
   __handleWindowScrollEvent (event) {
+    if (!this._isMounted) return;
+    
     const { onScroll } = this.props
 
     const scrollElement = this.props.scrollElement || window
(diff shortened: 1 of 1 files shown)


source/WindowScroller/WindowScroller.js
```diff
@@ -127,6 +131,8 @@ export default class WindowScroller extends PureComponent {
 
   // Referenced by utils/onScroll
   __handleWindowScrollEvent (event) {
+    if (!this._isMounted) return
```
mking-clari: @liorbrauer I think ideally we should determine why the mountedInstances array is not accurate, rather than fixing it with a flag

https://github.com/bvaughn/react-virtualized/blob/master/source/WindowScroller/utils/onScroll.js

liorbrauer: @mking-clari Good point, though looking at the code and playing around with it I couldn't pin point why it isn't accurate. New WindowScrollers are added to `mountedInstances` array on mount and are removed on unmount. My best guess is that there's a race condition since scroll events are fired rapidly. If that's the case, I'm not sure what the best course of action is, perhaps `debounce` the scroll event handler?

Alternatively, I can move the check to see if the instance is mounted to the `onScroll.js` helper, changing line 47 to:

```js
if (instance._isMounted && instance.scrollElement === event.currentTarget) {
```

though that still requires using a flag. Any other thoughts?

bvaughn: I'm willing to accept this fix, but I'd like to know more about the case that it's fixing so I can determine if there's a better fix. Can you provide a Plnkr (or similar) repro @liorbrauer ?

liorbrauer: @bvaughn here's a plnkr with the problematic setup:

http://plnkr.co/edit/lquCfWiYbNnypf6JrvgW?p=preview

Just open the console and scroll fast in the preview pane (or simply drag the scrollbar fast). Let me know what you think :)

bvaughn: Cool. Thanks!

bvaughn:
Much appreciated!

bvaughn:

```
```

bvaughn:
Haven't heard anything negative about the RC so I just released 9.8.0

Issue #642, opened 2017-04-04
Column label, string => node


hnordt:
For some columns, I would like to be able to send React nodes, e.g.:

![image](https://cloud.githubusercontent.com/assets/1625399/24675010/f73d8c7a-1953-11e7-9d7d-ac9922834fb6.png)

So I recommend changing `label: PropTypes.string,` to `label: PropTypes.node,` on https://github.com/bvaughn/react-virtualized/blob/master/source/Table/Column.js#L55

bvaughn:
If you'd like to submit a PR for this that:
1. Makes the suggested change
1. Adds at least one Jest test
1. Updates `Column` docs

Then I'd be happy to review it. Otherwise though, I don't plan on making this change. 😄

hnordt:
@bvaughn sure, I'll do that when I have time.

That's good to know you'll accept it if I do. 👍

bvaughn:
Yup! Sounds like a change I'd be happy to accept.

bvaughn:
FYI @hnordt looks like @EnrikoLabriko submitted a PR for this change and it will go out with the next release. 😄

hnordt:
Awesome! Thank you.

I forgot to send the pill request. I'm sorry. 😐

PR #694, opened 2017-06-03
Change Column label propType to node

EnrikoLabriko:
Changes Column label propType to `node`
See #642 

- Added tests
- Updated Column docs
- Tests are passing
- Lint is passing


bvaughn:
Looks great! Thanks for being thorough. 😄

PR #706, opened 2017-06-23
Please create possibility to pass containerStyle prop through List to Grid

GLiberman:
Grid Component have prop which pass styles to inner cell-container -
 "ReactVirtualized__Grid__innerScrollContainer"

I'm sure that small fix wouldn't break anything - it's just optional prop.

### DIFF
--- a/source/List/List.js
+++ b/source/List/List.js
@@ -24,6 +24,9 @@ export default class List extends PureComponent {
 
     /** Optional CSS class name */
     className: PropTypes.string,
+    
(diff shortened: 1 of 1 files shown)


bvaughn:
This is already being done by [this line](https://github.com/GLiberman/react-virtualized/blob/c41239d975afbd71905d426b6894098ec1433733/source/List/List.js#L172) :smile: `List` passes through all unknown props to `Grid`

Issue #725, opened 2017-07-06
List/Grid scroll sync fails when using wheel when total height exceeds DEFAULT_MAX_SCROLL_SIZE


nathanpower:
It seems that when syncing two or more scrollable lists whose total height exceeds DEFAULT_MAX_SCROLL_SIZE here: https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/utils/ScalingCellSizeAndPositionManager.js#L9, the cell offset calculation does not produce correct results.

Here is a fiddle demonstrating the problem: https://jsfiddle.net/nathanpower/823w11xr/1/
Note that using the scroll-bar is fine, but using a scroll wheel or two-finger trackpad scroll results in messed up rendering of the synced list.

I will try and look into this myself too, just wanted to bring it to your attention. If I get anywhere I will update.

nathanpower:
whoops, incorrect link to the fiddle, fixed now

bvaughn:
Looks like a bug. Would love a PR if you can find the time to look into it.

nathanpower:
Do you think something like this would be appropriate, or would it sacrifice perf elsewhere?

<img width="998" alt="screen shot 2017-07-06 at 22 51 30" src="https://user-images.githubusercontent.com/13329684/27934453-c6966fa4-629d-11e7-9ae1-85e47bddf771.png">

nathanpower:
Basically, if the list/grid which is following the scroll of another list/grid exceeds the height of the browser limit, _and_ it's style is cached, the problem occurs.

nathanpower:
Actually maybe I am overthinking it, and the bug is just an `||` when an `&&` was intended. Will PR that first, as it is the simplest change which fixes this issue.

<img width="719" alt="screen shot 2017-07-07 at 09 42 28" src="https://user-images.githubusercontent.com/13329684/27950373-e631daa0-62f8-11e7-8d0f-09a8cc9b4709.png">

bvaughn:
I think that's correct! Looks like my mistake there. 😄

nathanpower:
Fixed in 9.9.0, many thanks for your efforts @bvaughn, sorry that I didn't get further with the tests there.

bvaughn:
No problem. Thanks for following up about this!

On Jul 10, 2017 1:14 AM, "Nathan Power" <notifications@github.com> wrote:

Closed #725 <https://github.com/bvaughn/react-virtualized/issues/725>.

—
[...]

PR #727, opened 2017-07-07
Fix invalid rendering in scroll synced list with height/width exceeding browser limit

nathanpower:
See issue: https://github.com/bvaughn/react-virtualized/issues/725

It seems that when syncing two or more scrollable lists whose total height exceeds DEFAULT_MAX_SCROLL_SIZE here: https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/utils/ScalingCellSizeAndPositionManager.js#L9, the cell offset calculation does not produce correct results.

Here is a fiddle demonstrating the problem: https://jsfiddle.net/nathanpower/823w11xr/1/
Note that using the scroll-bar is fine, but using a scroll wheel or two-finger trackpad scroll results in messed up rendering of the synced list.

This PR changes to only caching the style when not scrolling AND offsets are not adjusted. Maybe this was the original intention here? 

Also some minor typos are corrected and `.vscode` folder added to `.gitignore`, hope that's ok.

### DIFF
--- a/source/Grid/defaultCellRangeRenderer.js
+++ b/source/Grid/defaultCellRangeRenderer.js
@@ -37,7 +37,7 @@ export default function defaultCellRangeRenderer ({
     rowSizeAndPositionManager.areOffsetsAdjusted()
   )
 
-  const canCacheStyle = !isScrolling || !areOffsetsAdjusted
+  const canCacheStyle = !isScrolling && !areOffsetsAdjusted;
 
   for (let rowIndex = rowStartIndex; rowIndex <= rowStopIndex; rowIndex++) {
     let rowDatum = rowSizeAndPositionManager.getSizeAndPositionOfCell(rowIndex)
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist
 node_modules
 npm-debug.log
 styles.css
+.vscode
--- a/source/Collection/CollectionView.js
+++ b/source/Collection/CollectionView.js
@@ -8,7 +8,7 @@ import getScrollbarSize from 'dom-helpers/util/scrollbarSize'
 // @TODO Merge Collection and CollectionView
 
 /**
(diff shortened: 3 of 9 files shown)


source/Grid/defaultCellRangeRenderer.js
```diff
@@ -37,7 +37,7 @@ export default function defaultCellRangeRenderer ({
     rowSizeAndPositionManager.areOffsetsAdjusted()
   )
 
-  const canCacheStyle = !isScrolling || !areOffsetsAdjusted
+  const canCacheStyle = !isScrolling && !areOffsetsAdjusted;
```
bvaughn: nit: This project uses standard (no semis)

nathanpower: whoops! will update

nathanpower:
>How'd you spot/fix all my little typos in docs comments. Do you have some sort of tool that catches those? I'd love to know about it.

I installed a simple [spell checker plugin](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker) for vscode a couple of weeks ago, it's working pretty well.

nathanpower:
Just realized that my test is not testing what I thought it was, will revisit that too.

bvaughn:
> I installed a simple spell checker plugin for vscode a couple of weeks ago, it's working pretty well.

That's awesome! Clearly I need to install the same plug-in!

> Just realized that my test is not testing what I thought it was, will revisit that too.

Okay! No hurry. Let me know when it's ready to re-review. 😄

nathanpower:
mmmm... this change breaks some other tests, I'm not clear on whether those tests were relying on incorrect behavior or not

bvaughn:
Thank you for this bug-fix!

This feature has just been released in version 9.9.0 of the library and I've given you credit in the [CHANGELOG](https://github.com/bvaughn/react-virtualized/blob/master/CHANGELOG.md#990).

Issue #405, opened 2016-09-20
Ship flow typings


andreypopp:
I see React Virtualized annotated with Flow. It would nice if React Virtualized was distributed with flow typings.

bvaughn:
Hey @andreypopp. This seems like a nice suggestion. 👍

Know of any libraries that do this [that I could use as a reference]?

andreypopp:
So what I do usually if library already uses flow is just copy original sources along by change their extension to `.js.flow` (Flow uses them instead of compiled JS to get type info from).

andreypopp:
Also you are using `propTypes` now, it might be beneficial to switch to types for `props` because types can express much more type info: for example callbacks will describes what arguments they accept.

Flow types for props then can be compiled to `propTypes` (with the help of corresponding [Babel plugin](https://github.com/brigand/babel-plugin-flow-react-proptypes)) for consumers who don't use Flow yet.

I can help with that.

bvaughn:
> So what I do usually if library already uses flow is just copy original sources along by change their extension to .js.flow (Flow uses them instead of compiled JS to get type info from).

Copy to..? (Would it go where the ES6 modules go?)

> Also you are using `propTypes` now, it might be beneficial to switch to types for `props` because types can express much more type info

Yeah, I've been meaning to make the `propTypes` => `props` switch for a while but I just haven't had the time. If you'd like to help- I'd welcome it! 😄

andreypopp:
>> So what I do usually if library already uses flow is just copy original sources along by change their extension to .js.flow (Flow uses them instead of compiled JS to get type info from).

> Copy to..? (Would it go where the ES6 modules go?)

That's an interesting question... probably in both locations then... this needs a little more investigation.

bvaughn:
Agreed 👍

turadg:
> So what I do usually if library already uses flow is just copy original sources along by change their extension to .js.flow

Would `flow gen-flow-files` be preferable?
```
  gen-flow-files  EXPERIMENTAL: Generate minimal .js.flow files for publishing to npm.
```

I mean, if it works :)

bvaughn:
> Would `flow gen-flow-files` be preferable?

Yes! 😄

sammkj:
[flow-copy-source](https://github.com/AgentME/flow-copy-source) is another alternative. It's used in React Motion

bvaughn:
[...]

PR #730, opened 2017-07-09
WIP flow types

TrySound:
Continue this threads
https://github.com/bvaughn/react-virtualized/issues/405
https://github.com/bvaughn/react-virtualized/pull/426

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
     "express": "^4.13.3",
     "extract-text-webpack-plugin": "^1.0.1",
     "file-loader": "^0.8.5",
-    "flow-bin": "^0.39.0",
+    "flow-bin": "^0.49.1",
     "fs-extra": "^0.30.0",
     "gh-pages": "^0.11.0",
     "html-webpack-plugin": "^2.16.1",
--- a/source/ArrowKeyStepper/ArrowKeyStepper.example.js
+++ b/source/ArrowKeyStepper/ArrowKeyStepper.example.js
@@ -7,22 +7,17 @@ import Grid from '../Grid'
 import cn from 'classnames'
 import styles from './ArrowKeyStepper.example.css'
 
+type Scroll = {
+  scrollToColumn: number,
+  scrollToRow: number
+};
+
 export default class ArrowKeyStepperExample extends PureComponent {
-  constructor (props) {
-    super(props)
-
-    this.state = {
-      mode: 'edges',
-      isClickable: true,
-      scrollToColumn: 0,
-      scrollToRow: 0
-    }
-
-    this._getColumnWidth = this._getColumnWidth.bind(this)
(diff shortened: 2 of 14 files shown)


TrySound:
I found CellMeasurer `index` prop is not documented

bvaughn:
Hey! I appreciate the work you've done here! 😄

I'm not sure that _replacing_ `propTypes` with Flow types is a good user experience though for the (probably large) percentage of react-virtualized users who _don't_ use the Flow type checker.

It's possible this concern could be addressed with the use of a plugin like [`flow-react-proptypes`](https://github.com/brigand/babel-plugin-flow-react-proptypes) but we would need to make sure there is no user-facing change to the bundled NPM-dist before I'd be willing to move forward with an approach like this.

> I found CellMeasurer `index` prop is not documented

This prob isn't (publicly) documented because it's only a [fallback prop used if the component accidentally passes `index` instead of `rowIndex`](https://github.com/bvaughn/react-virtualized/blob/c828cf2ab0cf35ff2cc7462af02c5269e352df1c/source/CellMeasurer/CellMeasurer.js#L89). I prefer to leave it as kind of a hidden feature.

TrySound:
But propTypes is not used everywhere. Looks like current code base is a bit frankenstein with flow and prop types. But flow fails with 1.5k errors. Should it be used then?
https://github.com/bvaughn/react-virtualized/blob/master/source/CellMeasurer/CellMeasurer.js

bvaughn:
propTypes are used on all user-facing React components. Where is that not the case?

Flow errors because I never actually got around to using it in this project.

TrySound:
I mean maybe then remove flow types completely?

bvaughn:
I don't understand why this is the direction the discussion is taking. Why
does the partial presence of Flow annotations bother you?

If you're interested in RV being fully Flow-typed, I'm supportive of that-
provided it does not break type checking for the majority of library users.
(Also provided it's a feature you're willing to do the work on, because I
don't have the bandwidth for it.)

TrySound:
@bvaughn Yeah. I'm all for that. But can we cover RV with atomic PRs like per component? This let you do more detailed review.

bvaughn:
Great. Then I'm happy to have your help.

I think we have 2 options for how to proceed:

1. Get `flow-react-proptypes` setup and working first (to generate `propTypes` so that the NPM bundle isn't impacted for non-Flow users). After this we could incrementally convert components in master to use Flow through several focused PRs.
1. Incrementally migrate components to use Flow by creating several PRs into a branch _other than master_ and then merge to master once `propTypes` compatibility is settled.

Either way, I think it's important that RV can be released from master in case any critical bugs come up (or in case you lose interest or are otherwise unable to complete the migration).

What do you think?

TrySound:
Agree

PR #734, opened 2017-07-13
Added "scrollToOffset" feature to Grid/List

scf4:
This lets the user add to (or subtract from) the calculated `scrollTo` position. 

E.g., if you have a list of posts in a `WindowScroller` and need to have a margin between the top of the page and the top of the post (when using `scrollTo`), this let's you do that with the `scrollToOffset` prop.

### DIFF
--- a/source/Grid/Grid.jest.js
+++ b/source/Grid/Grid.jest.js
@@ -331,6 +331,22 @@ describe('Grid', () => {
       expect(grid.state.scrollTop).toEqual(940)
     })
 
+    it('should scroll to the correct position for :scrollToOffset', () => {
+      const grid = render(getMarkup({
+        scrollToAlignment: 'start',
+        scrollToOffset: 20,
+        scrollToColumn: 24,
+        scrollToRow: 49
+      }))
+      // 100 columns * 50 item width = 5,000 total item width
+      // 100 rows * 20 item height = 2,000 total item height
+      // 4 columns and 5 rows can be visible at a time.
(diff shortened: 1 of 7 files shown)


bvaughn:
Hey @scf4! I appreciate the pull request!

Unfortunately I don't really want to add more complexity to a feature that's already pretty complex. There are 2 scroll-to methods ([`scrollToCell`](https://github.com/bvaughn/react-virtualized/blob/master/docs/Grid.md#scrolltocell--columnindex-number-rowindex-number-) and [`scrollToPosition`](https://github.com/bvaughn/react-virtualized/blob/master/docs/Grid.md#scrolltoposition--scrollleft-scrolltop-)) as well as 2 props (`scrollToRow` and `scrollTop`). I think that's sufficient to implement similar functionality to what this PR contains in user code without adding it to RV.

eg you could use [`getOffsetForCell`](https://github.com/bvaughn/react-virtualized/blob/master/docs/Grid.md#getoffsetforcell--alignment-string-columnindex-number-rowindex-number-) to get the location of a cell and then [`scrollToPosition `](https://github.com/bvaughn/react-virtualized/blob/master/docs/Grid.md#scrolltoposition--scrollleft-scrolltop-) to scroll to that position plus some arbitrary offset you want to add for margin. If this is something you find yourself doing often, you could decorate a `List` or `Grid` and add this behavior in.

Once a feature like this goes into the core library though, I'm committed to supporting it going forward and I think in this case, I prefer a user-code solution. Hope you understand! 😄

scf4:
@bvaughn No worries, I understand. Thanks anyway!

PR #737, opened 2017-07-13
Cover CellMeasurer with types

TrySound:
Temporary removed flow from examples until `List` will be covered

### DIFF
--- a/source/CellMeasurer/CellMeasurer.js
+++ b/source/CellMeasurer/CellMeasurer.js
@@ -1,7 +1,7 @@
 /** @flow */
 import React from 'react'
 import { findDOMNode } from 'react-dom'
-import CellMeasurerCache from './CellMeasurerCache.js';
+import CellMeasurerCache from './CellMeasurerCache.js'
 
 type Children = (params: { measure: () => void }) => React.Element<*>
 
--- a/source/CellMeasurer/CellMeasurer.js
+++ b/source/CellMeasurer/CellMeasurer.js
@@ -3,7 +3,7 @@ import React from 'react'
 import { findDOMNode } from 'react-dom'
 import CellMeasurerCache from './CellMeasurerCache.js';
 
-type Children = (params: { measure: () => void }) => React.Children
+type Children = (params: { measure: () => void }) => React.Element<*>
 
 type Cell = {
   columnIndex: number,
@@ -12,7 +12,7 @@ type Cell = {
 
 type Props = {
   cache: CellMeasurerCache,
-  children: Children | React.Children,
+  children: Children | React.Element<*>,
   columnIndex?: number,
   index?: number,
   parent: {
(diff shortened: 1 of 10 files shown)


source/CellMeasurer/CellMeasurer.js
```diff
@@ -1,35 +1,42 @@
 /** @flow */
-import { PureComponent } from 'react'
+import React from 'react'
 import { findDOMNode } from 'react-dom'
+import CellMeasurerCache from './CellMeasurerCache.js';
+
+type Children = (params: { measure: () => void }) => React.Children
```
bvaughn: I don't think `React.Children` is what you think it is. (I think it must be explicitly imported via `import type {Children} from 'react'` in the current implementation of Flow.)

bvaughn:
Need more time to actually review this. Just quickly scanned through it between meetings. But so far it looks great!

source/CellMeasurer/CellMeasurer.js
TrySound: Looking at typings, it's any :) Maybe better use `React$Element<*>`?

bvaughn: Yeah, Flow doesn't support pulling in types this way yet (although I think they're planning to add this soon!)

In this case, I think `React.Element<any>` yeah

TrySound:
There's two warnings about using class getters
> Potentially unsafe get/set usage. Getters and setters with side effects are potentially unsafe and disabled by default.

source/CellMeasurer/CellMeasurer.js
```diff
@@ -1,35 +1,42 @@
 /** @flow */
-import { PureComponent } from 'react'
+import React from 'react'
 import { findDOMNode } from 'react-dom'
+import CellMeasurerCache from './CellMeasurerCache.js';
```
bvaughn: nit: This project uses standard so the semi causes lint to fail

source/CellMeasurer/CellMeasurer.js
```diff
@@ -48,45 +55,49 @@ export default class CellMeasurer extends PureComponent {
...
-    // Explicitly clear width/height before measuring to avoid being tainted by another Grid.
-    // eg top/left Grid renders before bottom/right Grid
-    // Since the CellMeasurerCache is shared between them this taints derived cell size values.
-    if (!cache.hasFixedWidth()) {
-      node.style.width = 'auto'
-    }
-    if (!cache.hasFixedHeight()) {
-      node.style.height = 'auto'
-    }
+    if (node instanceof HTMLElement) {
```
bvaughn: `HTMLElement` is not defined so far as lint knows

TrySound: Lint know nothing. It's global class.

bvaughn: Sure. Lint is dumb. We need to give Lint hints for global things 😄

bvaughn:
Sweet. Thanks!

PR #740, opened 2017-07-15
Replace standard with prettier+eslint

bvaughn:
Move linting from pre-test to pre-commit phase.


bvaughn:
Too big to review! 🤞

package.json
```diff
@@ -10,31 +10,37 @@
   "jsnext:main": "dist/es/index.js",
   "license": "MIT",
   "scripts": {
-    "build": "yarn run build:commonjs && yarn run build:css && yarn run build:es && yarn run build:demo && yarn run build:umd",
-    "build:commonjs": "yarn run clean:commonjs && cross-env NODE_ENV=production cross-env BABEL_ENV=commonjs babel source --out-dir dist/commonjs --ignore *.example.js,*.test.js,source/demo/,source/tests.js",
+    "build": "npm run build:commonjs && npm run build:css && npm run build:es && npm run build:demo && npm run build:umd",
```
TrySound: Yarn doesn't work?

bvaughn: It should, but I'm trying to debug [this](https://circleci.com/gh/bvaughn/react-virtualized/1514) failure:
```
$yarn
bash: line 1: yarn: command not found

yarn returned exit code 127

Action failed: yarn
```

Just trying different things out. Not sure why it's failing. ☹️ Tried upgrading Circle from Ubuntu 12 to 14 and that didn't help. Tried modifying my circle.yml as shown in the examples. Ran out of ideas so I tried this.

.eslintrc
```diff
@@ -0,0 +1,16 @@
...
+  "parser": "babel-eslint",
+  "parserOptions": {
+    "sourceType": "module"
+  },
+  "plugins": [
+    "flowtype",
+    "prettier"
+  ],
+  "rules": {
+    "prettier/prettier": "error"
```
TrySound: Why both and precommit and linting?

package.json
bvaughn: TBH though I wasn't sure why you changed `npm` -> `yarn` in the scripts to begin with. Yarn improves the installation/resolution process but just shells out to NPM for running scripts, no?

bvaughn: Ah! "Rebuild" in the CircleCI interface wasn't enough to pick up the Ubuntu version change. Had to push another commit.

TrySound: I think using both things here is quite unnecessary. For example I don't have npm on my new Mac, just yarn.)

bvaughn:
Look at that 😁 CI is passing again

.eslintrc
bvaughn: I'm not super knowledgable about Eslint. I just followed the suggested config from the Prettier site.

package.json
bvaughn: I use Yarn only for installation. I still run scripts via `npm` and `npx`.

.eslintrc
TrySound: Lint will always success with this rules if prettier runs before.

package.json
bvaughn: To be clear, I don't really object to the `npm` -> `yarn` switch in `package.json`. It just felt unnecessary.

.eslintrc
bvaughn: My understanding is that this rule configures Eslint to be _compatible_ with Prettier styles (or maybe rather, not to _conflict_ with Prettier's preferences), but they still do different things. For example, if I add the `no-unused-expressions` rule (which I probably will in a follow-up) to Eslint's config, it will complain about variables that are declared but not used. Prettier will ignore them.

PR #743, opened 2017-07-17
Cover Grid component with types

TrySound:
Wooh. It's heavy. I splitted a few types in Grid. Maybe then it will be lifted for all components. I found a lot effects with comparing `undefined` and `number` and tried to get rid in favour of `-1`. Will try to consolidate further.

PR #744, opened 2017-07-17
Grid: Improve accessibility by adding props `aria-readonly` & `role="rowgroup"`

smockle:
## What changed?

1. [`aria-readonly`](https://www.w3.org/TR/wai-aria/states_and_properties#aria-readonly) is set on the _outer `<div>` with [`role="grid"`](https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/Grid.js#L259)_ to indicate it contains elements which are operable but not editable (contra the [`grid` role](https://www.w3.org/TR/wai-aria/roles#grid) default). It can be overridden. The default value is `true`.

2. [`role="rowgroup"`](https://www.w3.org/TR/wai-aria/roles#rowgroup) is set on the _inner container `<div>`_ to establish the relationship between the _outer `<div>` with [`role="grid"`](https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/Grid.js#L259)_ and _children with `role="row"`_. It helps assistive technologies identify the inner container as something akin to a `<tbody>`. It can be overridden. The default value is `"rowgroup"`.

## Why did it change?

Together, these props allow [JAWS](http://www.freedomscientific.com/downloads/jaws) to recognize `Grid` as a table and enable “Table Layer” reading [commands](http://www.freedomscientific.com/training/Surfs-Up/Table_Reading_Commands.htm).

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -35,6 +35,7 @@ const SCROLL_POSITION_CHANGE_REASONS = {
 export default class Grid extends PureComponent {
   static propTypes = {
     "aria-label": PropTypes.string,
+    "aria-readonly": PropTypes.bool,
 
     /**
      * Set the width of the inner scrollable container to 'auto'.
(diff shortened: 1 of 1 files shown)


bvaughn:
This is excellent! Thanks for improving accessibility! ❤️

I'm going to hold off on merging this until #743 has been merged since that PR is pretty complicated and I don't want to cause any conflicts.

bvaughn:
I've merged master into this branch and manually resolved conflicts

bvaughn:

Issue #722, opened 2017-07-05
Fix Chrome setTimeout throttling


uglywizkid:
Hi there.

We have faced a scrolling issue using react-virtualized table with the Chrome browser. After scrolling with the mouse wheel, we experience a quite long time of 1.5 to 2 seconds for which our web application stops responding as we would expect. However this issue does not occur when moving the scroll bar. We also experienced better performance in other browsers, e.g. Firefox.

Performance profiling in Chrome showed us, that after a scroll event the timer does not fire as expected after 150ms (how it is implemented, see DEFAULT_SCROLLING_RESET_TIME_INTERVAL in Grid.js) but after 1.5 seconds or even more.

![screenshot-1](https://user-images.githubusercontent.com/7262707/27861564-a493fc54-6181-11e7-8cd0-c1e1ceba892f.png)

We researched for an explanation and found a similar issue explained in this bug report
[https://bugs.chromium.org/p/chromium/issues/detail?id=570845](https://bugs.chromium.org/p/chromium/issues/detail?id=570845)

It seems that this is not a bug, but a feature of Chrome. 
I followed the advice at the very end of the bug report and replaced the setTimer with requestAnimationFrame and clearTimer with cancelAnimationFrame (see also my fork on [https://github.com/uglywizkid/react-virtualized/blob/master/source/Grid/Grid.js](https://github.com/uglywizkid/react-virtualized/blob/master/source/Grid/Grid.js) ). This seems to result in much better performance.

A fix in your library would be very appreciated

bvaughn:
> A fix in your library would be very appreciated

A PR would be a good start to getting said fix into RV.

uglywizkid:
Any chance you could fix your project first? Your build is failing. In this case it is hard for me to check whether my changes break other things.

bvaughn:
There's nothing to fix. Use `yarn install` for dependencies and things work fine. CI failed recently for reasons unknown. I am unable to reproduce the failure locally.

bvaughn:
CI is passing again. Upgraded Circle from Ubuntu 12 -> 14 and that seems to have resolved the test issue.

TrySound:
@uglywizkid Will this behaviour be fixed with https://github.com/bvaughn/react-virtualized/pull/742?

giuseppepaul:
Hey guys, ran into this issue too, was driving me mad so thanks for the explanation @uglywizkid

Any news on when a fix for this will be available?

[...]

PR #742, opened 2017-07-17
Improving detection of scroll ended

guilhermefloriani:
We are facing some issues like #697 #722 #322 #564 and few others, reporting that there is a problem related to the detection of scroll ended.

Like some people already said in this kind of issue, this is happening because the browser is delaying the `setTimeout` callback on `_debounceScrollEnded` in exchange of performance reasons.

This is a proposal to fix this kind of problems using `requestAnimationFrame` instead of `setTimeout` and it works fine for me.

This PR includes:
- Using `requestAnimationFrame` instead of `setTimeout` in `Grid` component and `Masonry` component.
- Fixing jest setup to polyfill `requestAnimationFrame` and `cancelAnimationFrame`.
- Change a `Masonry` test to wait before all "animations" finish before assert the result. (I did it with a `setTimeout` callback because `jest.runOnlyPendingTimers()` does not wait a task scheduled with `requestAnimationFrame`, and this kind of task have priority for `setTimeout` "tasks", so the assert expression will be executed only after `_debounceScrollEnded` was called).

Issue #697, opened 2017-06-06
Scrolling timeout hover performance issue


Maxincredible52:
In my particular case I use the Table element and I have to set the hovered row index to the component state in order to activate special functionality in my sub-components that are rendered inside of a table row. 

The hovered row index is set by using the provided `onRowMouseOver` prop of the Table component.  

When scrolling sometimes the hover event takes several seconds to kick in (up to 2 to 3 seconds(!)) and prevents the table from being able to be hovered during that time span which is horrible for user experience as it can be seen in the following gif (note that I did not slow down, this is realtime)

![hover_delay](https://cloud.githubusercontent.com/assets/15246773/26803535/dca1de62-4a44-11e7-96ef-2aa30e61e879.gif)

The issue only occurs in chrome as far as i could test it, other browsers like Edge or Firefox do not have that issue. Another strange behavior of the problem is, that it seems to resolve itself. After a certain amount of time playing around with the table by scrolling and hovering over some rows the hover events behave as they should. When the page is reloaded the issue comes back though.

After research I came to the conclusion that this issue might relate to the issues #564 #322 and maybe #518 but I thought I have to open a new issue as there is no issue that exactly covers my use case.

bvaughn:
This is out of my control, as noted in those other issues you linked to ([here](https://github.com/bvaughn/react-virtualized/issues/564#issuecomment-277789650) and [here](https://github.com/bvaughn/react-virtualized/issues/322#issuecomment-234742880)). I'm not sure what more I can say that I've already said in those issues. If you don't want pointer events to be disabled while scrolling, you can override that behavior as noticed in the replies I linked to.

I'm closing this because it's not actionable. We can talk more here or on Slack if you have follow up questions or concerns though.

Maxincredible52:
Setting pointer-events to `"auto"` does **not** resolve the issue as the `onRowMouseOver` callback does not fire neither. It is not the pointer that is bad for user experience, it is the inability to get the visual feedback of the row being hovered. That is the reason why i opened an issue.

bvaughn:
If you read my replies to the issues you mentioned above, including [the one I linked to](https://github.com/bvaughn/react-virtualized/issues/564#issuecomment-277789650), you'll see this:

> I believe this is the correct default approach for performance reasons. However if you disagree that it's appropriate for your app then you can override by passing in your own `containerStyle` that sets `pointer-events: "auto"`.

Maxincredible52:
[...]

Issue #564, opened 2017-02-06
Delay in hover and click 


prog8:
React-virtualized breaks hover events in Grid. This means that cell can be hovered or clicked with delay after rendering it.

After debugging it truned out that setTimeout method in Grid.js file in react-virtualized (function _debounceScrollEnded) is not always working properly. After calling setTimeout with timeout=0 or 1 (or any other value) the callback (_debounceScrollEndedCallback) method was triggered with 1.5 to 2 seconds delay. What is more when I tried to call setTimeout (timeout=1ms) with simple concole.log message was also printed with long delay. I can see the problem on Chrome 55.0.2883.87 (64-bit) on Ubuntu 16.04.1 LTS. I also tried the same version of Chrome in Windows and could not reproduce the problem. The problem was not reproducible on Firefox.
I decided to report it here because I am not sure if this is a problem in React Virtualized or this is browser problem.

bvaughn:
> After calling setTimeout with timeout=0 or 1 (or any other value) the callback (_debounceScrollEndedCallback) method was triggered with 1.5 to 2 seconds delay.

`setTimeout` doesn't guarantee a callback will be invoked exactly after the specified `delay`- only that it won't be invoked _before_ then. The [Mozilla's docs](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout#Reasons_for_delays_longer_than_specified) list a few reasons why the callback may be delayed, including:
> In addition to "clamping", the timeout can also fire later when the page (or the OS/browser itself) is busy with other tasks

There's also blog posts like [this one](https://blog.chromium.org/2015/08/scheduling-timers-to-improve-input.html):
> The scheduler now delays impending expensive timers after a tap in anticipation of these tasks, allowing many web pages to be scheduled more efficiently. In practice, this can result in up to a 50% input latency improvement on websites that use timers heavily.

[...]

PR #745, opened 2017-07-18
issues/722

uglywizkid:
Pull Request for issue https://github.com/bvaughn/react-virtualized/issues/722


TrySound:
Browser specific hacks is bad practice. The problem should be fixed in a different way.

bvaughn:
FWIW I'm not against replacing `setTimeout` with `requestAnimationFrame` but I think this is solved with a smaller footprint in #742 (and that pull request came first anyway) so I'm going to close this one for now.

bvaughn:
FWIW I'm not against replacing `setTimeout` with `requestAnimationFrame`. There's been a lot of discussion about this (see #697 #722 #322 #564) and I think it's probably the right thing to do. So thanks for taking initiative and creating this PR! 😄 

This same issue is solved with a smaller footprint in #742 though- (and that pull request came first anyway)- so I'm going to close this one for now in favor of it.

PR #748, opened 2017-07-19
Add support to Table to allow components that extend Column

CptLemming:
This change allows the `Table` component to accept children that directly extend the `Column` component.

This allows me to create many Column types I can use directly in a Table, making code reuse simpler.

### Creating custom columns

```js
// Changing the default cellDataGetter to use lodash's `get` function
const cellDataGetter = ({ dataKey, rowData }) => {
  return get(rowData, dataKey);
};

export class GetColumn extends Column {
  static defaultProps = {
    ...Column.defaultProps,
    cellDataGetter: cellDataGetter,
  };
}

// Adding a column to render links using `react-router`
const cellLinkRenderer = ({ cellData, rowData }) => {
  if ('to' in rowData) {
    let linkUrl = rowData.to;
    if (typeof rowData.to === 'function') {
      linkUrl = rowData.to({ rowData, cellData });
    }
    return (<Link to={ linkUrl }>{ cellData }</Link>);
  }
  return cellData;
};

export class LinkColumn extends Column {
  static defaultProps = {
    ...Column.defaultProps,
    cellRenderer: cellLinkRenderer,
  };
}
```

### Using custom columns

```js
const list = [{ node: { name: 'foo' }, link: '/bar' }];

<Table
  height={100}
  width={100}
  rowCount={list.length}
  rowGetter={({ index }) => list[index]}
>
  <GetColumn
    label="Name"
    dataKey="node.name"
    width={50}
  />
  <LinkColumn
    label="More info"
    dataKey="link"
    width={50}
  />
</Table>
```

### DIFF
--- a/source/Table/Table.jest.js
+++ b/source/Table/Table.jest.js
@@ -103,6 +103,18 @@ describe("Table", () => {
       expect(result instanceof Error).toEqual(false);
     });
 
+    it("should accept subclasses of Column as children", () => {
+      class AnotherColumn extends Column {}
+
+      const children = [<AnotherColumn dataKey="foo" width={100} />];
+      const result = Table.propTypes.children(
+        { children },
+        "children",
+        "Table"
+      );
+      expect(result instanceof Error).toEqual(false);
+    });
+
     it("should not accept non-Column children", () => {
       const children = [<div />];
       const result = Table.propTypes.children(
--- a/source/Table/Table.js
+++ b/source/Table/Table.js
@@ -27,7 +27,8 @@ export default class Table extends PureComponent {
     children: props => {
       const children = React.Children.toArray(props.children);
       for (let i = 0; i < children.length; i++) {
-        if (children[i].type !== Column) {
+        const childType = children[i].type;
+        if (childType !== Column && !(childType.prototype instanceof Column)) {
           return new Error("Table only accepts children of type Column");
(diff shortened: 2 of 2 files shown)


bvaughn:

genomics-geek:
Is there a way to do this with composition instead of inheritance?  Seems like `Table component` is very strict on only accepting instance of `Column`

bvaughn:
Not currently.
https://github.com/bvaughn/react-virtualized/blob/master/source/Table/Table.js#L30-L38

genomics-geek:
@bvaughn - Would you be against changing this to `PropTypes.node` ?  Is there something from making this change?

bvaughn:
It just doesn't feel super compelling to me. I guess I'm not really _against_ the idea but I don't intend to make the change b'c I don't see it as a pressing need and I'm super strapped for time.

genomics-geek:
that's fair -  It's already an awesome package :).  

I will submit a PR for this, you can decide later if this makes sense to include it or not in a future release. Thanks for this package!

genomics-geek:
After digging into the code a bit more, I realized `columnData` is a prop that could be used to pass in more variables to cellRenderer.  Makes the need for composition less.  Maybe I missed it, but this wasn't very clear in the Columns doc.

PR #752, opened 2017-07-23
Remove babel-plugin-typecheck to fix flow bindings error

TrySound:
typecheck is deprecated so conflict with flow ast won't be fixed.
https://github.com/codemix/babel-plugin-typecheck#deprecated


bvaughn:
Nice work!

PR #753, opened 2017-07-23
Cover List component with types

TrySound:
There's one weird error I didn't solved yet.

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -574,7 +574,10 @@ export default class Grid extends PureComponent {
       this.setState({});
     }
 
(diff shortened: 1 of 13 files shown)


source/Grid/Grid.js
```diff
@@ -284,9 +265,9 @@ export default class Grid extends PureComponent {
     estimatedRowSize: 30,
     getScrollbarSize: scrollbarSize,
     noContentRenderer: () => null,
```
andreypopp: I think this should be annotated explicitly as `() => React.Element<*> | null` — my guess is flow messes this up with `$DIff<Props, DefaultProps>` somehow.

TrySound: Okay :( Just can't reproduce with simple example.

andreypopp: The problem with repro is probably because in tryflow you operate within the same module.

PR #768, opened 2017-07-31
Update WindowScroller documentation

jondubin:
The function WindowScroller takes in as a child will also pass `width` as a property, but the documentation doesn't currently reflect this.

https://github.com/bvaughn/react-virtualized/blob/master/source/WindowScroller/WindowScroller.js#L26


bvaughn:
Thanks 😄

PR #775, opened 2017-08-11
Update CellMeasurer docs

mattmilburn:
Suggest mentioning `clear()` and `clearAll()` as public methods for `CellMeasurerCache` in the docs.

These two functions were very useful for me in managing dynamic row heights for cells that are temporarily used to display any kind of "loading..." indicator while working with an infinite loading `List` that renders data requested from an API, unless its currently busy fetching data.

In my case, I had an animated, pure-css loader in a `List` displaying rows with dynamic heights. Once the loading indicator was replaced with the next batch of rows, the row that was once occupied by my loading indicator required a re-measuring via the use of `clear()`.

In another case where `AutoSizer` would trigger its `onResize` callback due to a window resize event, causing a Bootstrap container to change its width, I used `clearAll()` to completely re-measure all rows in my `List`.

### DIFF
--- a/docs/CellMeasurer.md
+++ b/docs/CellMeasurer.md
@@ -38,6 +38,14 @@ Note that while all of the individual parameters above are optional, you must su
 It would be inefficient to do so since the size of a row (or column) is equal to the largest cell within that row.
 See [below](#limitations-and-performance-considerations) for more information.
 
+### Public Methods
+
+##### clear (rowIndex: number, columnIndex: number)
+Reset cached data about a specific cell. This should be called when a cell needs to be re-measured to handle dynamic content. (eg. Rendering a loading indicator at the end of a series will cache its `rowIndex` and `columnIndex` to the height of the loading indicator. Once your actual data loads and replaces the loading indicator, you may need to use `clear()` to re-measure the cell that was previously occupied by the loading indicator.)
+
+##### clearAll ()
+Reset any cached data about already-loaded cells. This can be used in situations where your `Grid`, `List` or `Table` needs to reflow its content due to a resizing scenario. (eg. Window resizes and re-renders a responsive layout that may affect all cells.)
+
 ### Examples
 
 ###### Grid
--- a/docs/CellMeasurer.md
+++ b/docs/CellMeasurer.md
@@ -38,13 +38,17 @@ Note that while all of the individual parameters above are optional, you must su
 It would be inefficient to do so since the size of a row (or column) is equal to the largest cell within that row.
 See [below](#limitations-and-performance-considerations) for more information.
 
-### Public Methods
+### `CellMeasurerCache` Public Methods
 
 ##### clear (rowIndex: number, columnIndex: number)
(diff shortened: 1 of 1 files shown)


mattmilburn:
Excellent! Thanks for simplifying (and understanding) my verbiage. Much cleaner and concise :)

bvaughn:
Happy to collaborate! Feel free to submit more docs PRs if you have suggestions! 😄

tafelito:
@mattmilburn I have the same issue you mentioned about the loading. I did use the clear when I know a row has changed, but how do you know when the loading row has been replaced so you can clear the cache for that specific row? 

Also I have another issue regarding this that sometimes the loading is shown at the bottom, but then when new rows are loaded, they get loaded before replacing the loading row, so the loading ends up in the middle of the list before getting replaced. Did you have that issue?

mattmilburn:
@tafelito It sounds like you may need to keep track of exactly which `rowIndex` is being used to hold a loading indicator so you can reference that value later. While working through my issue, I created this Plunker (below) to isolate my problems.

You can see on LINE 186 I'm setting the value for `this._lastLoadingIndex` which will be used later on with the `clear()` method on LINE 68 in `componentWillUpdate()`. I'm just checking to see if the number of rows in my list has increased. If so, we clear cached measurements for every column in that row (which is why I'm using a `for` loop).

https://embed.plnkr.co/uoRdanlB1rXsgBe2Ej8i/

In my `componentWillUpdate()` function I'm checking to see if the number of rows is greater than the last number of rows, which will indicate that my data has had more data appended to it. 

As for the second part of your question, it sounds like something isn't used correctly. Any links to code examples you may have would be helpful.

tafelito:
Thanks @mattmilburn for the link, I was able to make it work now! There was only 1 issue with your example if you have a limit of rows to load. So instead of returning` index < rowCount - 1`  on `isRowLoaded` I just return` !!this.list[index]`, otherwise the `loadMoreRows` keeps being called, even if all the rows have been loaded before.

Also the key for this to work, that I also wasn't doing it, was to set the threshold to 1, without that the `loadMoreRows` it's also being called a few times on the initial load.

As per the other issue, apparently it was a mix of things, and I haven't had the issue again. But If I do I will isolate it in an example code

mattmilburn:
@tafelito Excellent! And yes, I intentionally setup that example to load more rows infinitely (no pun intended) because I was trying to keep the example very minimal while I narrowed down other issues related to cached cell measurements.

Happy to have helped!

tafelito:
Hey @mattmilburn, here is a small vid of the error I commented before, where the loading is shown even if there are some more rows already loaded. You can see that the loading is even shown twice. 

oct-06-2017 00-10-11

I was even able to reproduce it on this codepen (I took your example as a reference)

https://codepen.io/tafelito/pen/rGpbQd/?editors=0010

oct-06-2017 01-00-12

Sorry @bvaughn this is probably not the proper way to follow up with an issue, I just didn't wanted to create an new issue if it's not really an issue

PR #778, opened 2017-08-15
add onWheel prop

implausible:
This is a follow up to https://github.com/bvaughn/react-virtualized/pull/541.

The PR simply exposes the onWheel prop for Grid.

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -86,6 +86,9 @@ type Props = {
   /** Either a fixed column width (number) or a function that returns the width of a column given its index.  */
   columnWidth: CellSize,
 
+  /** Unfiltered props for the Grid container. */
+  containerProps?: Object,
+
   /** ARIA role for the cell-container.  */
   containerRole: string,
 
@@ -143,11 +146,6 @@ type Props = {
 
   /** Callback invoked with information about the section of the Grid that was just rendered.  */
   onSectionRendered: (params: RenderedSection) => void,
-
-  /**
-   * Optional callback invoked whenever onWheel is called.
-   */
-  onWheel?: (params: WheelEvent) => void,
 
   /**
    * Number of columns to render before/after the visible section of the grid.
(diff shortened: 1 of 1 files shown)


bvaughn:
I'm not sure I want to expose a "wheel" event handler on `Grid`. While scroll events are async, wheel and touch* events are sync (at least by default) and so can cause performance problem.s

implausible:
Well, without the ability to force react-virtualized to run wheel events synchronously, react-virtualized has huge tearing issues in electron. We talked about this previously, what would you suggest we do now?

implausible:
> Perhaps you could modify this approach to enable your onWheel handler to be injected in? eg what if Grid exposed a new onWheel: PropTypes.func (optinoal) property that enabled you to inject your own throttled wheel handler? This way if it works well for your use-case you could continue to use it without potentially causing regressions for existing apps.

This is where we left off last time, we haven't had the need to update react-virtualized in awhile, but we're planning on updating react-virtualized, soon. Before we can update react-virtualized, we need to make sure that we can fix the tearing that occurs in multiple grid scrolling. Since this seemed to be the direction we wanted to go last time, if it's no longer the direction you intend, what do we do now?

bvaughn:
Don't know. Need to think about it. Not immediately comfortable with encouraging usage of "wheel" events (with an explicit `onWheel` prop) since they have known perf impact. (Also there are no other top-level event handlers, excluding `onScroll`, which is a custom thing and _not_ a pass-thru for the built-in "scroll" event.

implausible:
I'll revise for a rest prop.

athomann:
Also was looking for an `onWheel` prop.  We are attempting to prevent the parent container from scrolling when you reach the bottom of the virtual list.  So far I can only implement that with onWheel, unless you know of another way?

athomann:
Nevermind, I should be able to solve it when this is released.

bvaughn:

Issue #769, opened 2017-07-31
Wizard is completely broken (no errors)


0xCMP:
When you visit the wizard I expected that checking the boxes would do something and I'd have an idea on which components to use for a full page, infinite scrolling, list of items with an unknown height.

This image shows that no matter what I do nothing changes. I checked the console and saw zero errors. I made sure all my content blockers were not affecting some possible `data.json` or similar that might have powered it and saw nothing was being blocked at all.
![2017-07-31-react-virtualize-wizard-broken](https://user-images.githubusercontent.com/505649/28794321-edacec84-7603-11e7-8ffb-81c78724fa02.gif)

# Are you asking a question?
No

# Are you reporting a bug or runtime error?
Yes. There were no errors

# Are you making a feature request?
No.

iamrobinpham:
https://github.com/JedWatson/react-codemirror/issues/106

bvaughn:
PRs welcome 😄

PR #784, opened 2017-08-22
Wizard demo: Add a key prop to CodeMirror that increments on state change

cvpcasada:
This is my attempt to fix https://github.com/bvaughn/react-virtualized/issues/769 by adding a `key` prop that updates when the state changes as suggested by https://github.com/JedWatson/react-codemirror/issues/106#issuecomment-318781325. This fix then will not be needed when that react-codemirror issue is resolved but until then, this workaround will do.

### DIFF
--- a/source/demo/Wizard/Wizard.js
+++ b/source/demo/Wizard/Wizard.js
@@ -18,6 +18,7 @@ export default class Wizard extends Component {
     super(props);
 
     this.state = {
+      key: 0,
       cellsHaveKnownHeight: true,
       cellsHaveKnownWidth: true,
       cellsHaveUniformHeight: true,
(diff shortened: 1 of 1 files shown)


source/demo/Wizard/Wizard.js
```diff
@@ -31,6 +32,9 @@ export default class Wizard extends Component {
     };
   }
 
+  updateState = obj =>
+    this.setState(state => ({ ...state, ...obj, key: state.key + 1 }));
```
TrySound: You don't need to assign state, setState do you by itself.

cvpcasada: Oh, thanks, I didn't know that setState functional callback will also merge with previous state. I'll update the code.

PR #785, opened 2017-08-23
cellMeasurerCache.clear() add 0 as default columnIndex

tihuan:
Bug: When calling `cellMeasurerCache.clear()` without a `columnIndex`, no cache will be deleted.

Expected behavior: Calling `cellMeasurerCache.clear()` without a `columnIndex` will clear cache at provided `rowIndex` and `columnIndex` at `0`.

Test added as well! Not sure if the test description is clear enough :D

Thank you!

### DIFF
--- a/source/CellMeasurer/CellMeasurerCache.jest.js
+++ b/source/CellMeasurer/CellMeasurerCache.jest.js
@@ -67,6 +67,20 @@ describe("CellMeasurerCache", () => {
     expect(cache.has(1, 0)).toBe(true);
   });
 
+  it("should clear a single cached row cell in column 0 when columnIndex param is absent", () => {
+    const cache = new CellMeasurerCache({
+      fixedHeight: true,
+      fixedWidth: true
+    });
+    cache.set(0, 0, 100, 20);
+    cache.set(1, 0, 100, 20);
+    expect(cache.has(0, 0)).toBe(true);
+    expect(cache.has(1, 0)).toBe(true);
+    cache.clear(0);
+    expect(cache.has(0, 0)).toBe(false);
+    expect(cache.has(1, 0)).toBe(true);
+  })
+
   it("should clear all cached cells", () => {
     const cache = new CellMeasurerCache({
       fixedHeight: true,
--- a/source/CellMeasurer/CellMeasurerCache.js
+++ b/source/CellMeasurer/CellMeasurerCache.js
@@ -94,7 +94,7 @@ export default class CellMeasurerCache {
     }
   }
 
-  clear(rowIndex: number, columnIndex: number) {
+  clear(rowIndex: number, columnIndex: number = 0) {
     const key = this._keyMapper(rowIndex, columnIndex);
 
     delete this._cellHeightCache[key];


TrySound:
Both parameters are required to be specified. It's not the place for defaults.

bvaughn:
I'm actually not opposed to this change, although I'd like to hear your reasoning @TrySound.

There is precedent for this with [`CellMeasurerCacheDecorator`](https://github.com/bvaughn/react-virtualized/blob/master/source/MultiGrid/CellMeasurerCacheDecorator.js) which defaults all `columnIndex` values to 0. I think it's reasonable to do the same with `CellMeasurerCache`.

TrySound:
It's just about API simplicity. I think rowIndex shouldn't be passed without columnIndex and vice versa. Better to pass both or nothing.

bvaughn:
I tend to think default parameter values can help make APIs simpler. Particularly in this case, since other methods in `CellMeasurerCache` such as [`getWidth`](https://github.com/bvaughn/react-virtualized/blob/master/source/CellMeasurer/CellMeasurerCache.js#L149), [`getHeight`](https://github.com/bvaughn/react-virtualized/blob/master/source/CellMeasurer/CellMeasurerCache.js#L137), and [`has`](https://github.com/bvaughn/react-virtualized/blob/master/source/CellMeasurer/CellMeasurerCache.js#L161) already have this defaulting behavior.

TrySound:
Good defaults should be complete, for example
```js
const fn = (position = { x: 0, y: 0 }) => {};
```
but this way introduces more conditions
```js
const fn = (x = 0, y = 0) => {};
```
well here you call with specific x and omit y for simplicity, but then why to change y I should specify x (want I use current one or 0).
Here we go for simpler API with single responsibility methods
```js
const fnX = x => {};
const fnY = y => {};
```
Zero is not the value which is hard to specify in difference with current value.

TrySound:
cell and row here are not separated entities which can be passed separately, they are one thing - position, let's keep 'em together.

TrySound:
Also in ten we should remove those defaults.

bvaughn:
I think you're branching off to a broader topic, where as I'm discussing this particular component's API. The reason there are default values for `columnIndex` in several of the cache's methods is to make it easier to use with `List` or `Table` - components that themselves don't make mention of a column index due to the fact that they only have 1.

On a broader level, I don't believe it's wrong for tail parameters to a function to have default values. And I don't think it's necessary for _all_ parameters to have default values if _any_ parameters have them, so long as the "optional" parameters comes last. This is very common.

TrySound:
Then maybe we need RowMeasurer which will wrap CellMeasurer API.

TrySound:
I just prefer separation and composition over conditions. And yeah I'm a bit lost topic, but grid has or had such problems too.

bvaughn:
> Then maybe we need RowMeasurer which will wrap CellMeasurer API.

Perhaps, although composition has slight perf costs (additional instances and function calls) and usability costs (knowing the wrapper exists and when to use it). I'm not convinced this is a problem with the `CellMeasurerCache`.

> And yeah I'm a bit lost topic, but grid has or had such problems too.

Clearly we will have to discuss this more (but not in this PR).

bvaughn:
I'm going to accept this for now, given the precedent set by the other `CellMeasurerCache` methods, because I think consistency is good . We can revisit the topic before v10.

PR #789, opened 2017-08-25
Add `keyGetter` prop (10x performance improvement)

ashtonsix:
**Motivation**

My app has complex rows which take a long time to mount, but negligible time to re-render. As I scroll down React Virtualized seems to unmount rows from the top of my list & create brand new rows at the bottom. I want to move rows from the top to the bottom instead. Additionally, I want to avoid unnecessary rendering when inserting/deleting rows & sorting short lists. ie, I require custom logic to manage the keys passed to each row in a list.

**Solution**

> Made possible with the new `keyGetter` prop (`cellKeyGetter` in `Grid`)

Track the index each document should be displayed at manually & associate a key to that document. If the document has no associated key it should grab one from the released key pool or generate a new key & then reinitialize the row associated with that document by dispatching an action.

Release keys into the pool by watching for changes in `startIndex` & `endIndex` (indicates a section re-render - the intersection can be released).

**API**

```js
keyGetter={({startIndex, stopIndex, index}) => key}
```

**FYI**

My app is a relatively unique use case, with lots of features built around React Virtualized. It uses `List`, connects each visible row to a state management library for re-rendering, has a custom height cache & calls `List.prototype.repositionRowsAfterIndex` directly.

11 tests were failing when I tested locally (before & after I made any changes), haven't added any myself yet.

I would have done this without modifying React Virtualized if `Grid` didn't have style/cell caches.

Thanks for making this library! ❤

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -1,6 +1,7 @@
 /** @flow */
 
 import type {
+  CellKeyGetter,
   CellRenderer,
   CellRangeRenderer,
   CellPosition,
@@ -25,6 +26,7 @@ import defaultOverscanIndicesGetter, {
   SCROLL_DIRECTION_FORWARD
 } from "./defaultOverscanIndicesGetter";
 import updateScrollIndexHelper from "./utils/updateScrollIndexHelper";
+import defaultCellKeyGetter from "./defaultCellKeyGetter";
 import defaultCellRangeRenderer from "./defaultCellRangeRenderer";
 import scrollbarSize from "dom-helpers/util/scrollbarSize";
 import {
@@ -75,6 +77,8 @@ type Props = {
    * Intended for use with WindowScroller
    */
   autoWidth: boolean,
+
+  cellKeyGetter: CellKeyGetter,
 
   /** Responsible for rendering a cell given an row and column index.  */
   cellRenderer: CellRenderer,
@@ -234,6 +238,7 @@ export default class Grid extends React.PureComponent {
     autoContainerWidth: false,
     autoHeight: false,
     autoWidth: false,
+    cellKeyGetter: defaultCellKeyGetter,
     cellRangeRenderer: defaultCellRangeRenderer,
     containerRole: "rowgroup",
     containerStyle: Object,
@@ -961,6 +966,7 @@ export default class Grid extends React.PureComponent {
     state: State = this.state
   ) {
     const {
+      cellKeyGetter,
       cellRenderer,
       cellRangeRenderer,
       columnCount,
@@ -1057,6 +1063,7 @@ export default class Grid extends React.PureComponent {
 
       this._childrenToDisplay = cellRangeRenderer({
         cellCache: this._cellCache,
+        cellKeyGetter,
         cellRenderer,
         columnSizeAndPositionManager: this._columnSizeAndPositionManager,
         columnStartIndex: this._columnStartIndex,
--- a/source/Grid/defaultCellKeyGetter.js
+++ b/source/Grid/defaultCellKeyGetter.js
@@ -1,0 +1,10 @@
+/** @flow */
+
+import type { CellKeyGetterParams } from "./types";
+
+export default function defaultKeyGetter({
+  rowIndex,
+  columnIndex
+}: CellKeyGetterParams) {
+  return `${rowIndex}-${columnIndex}`
+}
--- a/source/Grid/defaultCellRangeRenderer.js
+++ b/source/Grid/defaultCellRangeRenderer.js
@@ -16,6 +16,7 @@ export default function defaultCellRangeRenderer({
   deferredMeasurementCache,
   horizontalOffsetAdjustment,
   isScrolling,
+  cellKeyGetter,
   parent, // Grid (or List or Table)
   rowSizeAndPositionManager,
   rowStartIndex,
@@ -54,7 +55,14 @@ export default function defaultCellRangeRenderer({
         columnIndex <= visibleColumnIndices.stop &&
         rowIndex >= visibleRowIndices.start &&
         rowIndex <= visibleRowIndices.stop;
-      let key = `${rowIndex}-${columnIndex}`;
+      let key = cellKeyGetter({
+        rowStartIndex,
+        rowStopIndex,
+        rowIndex,
+        columnStartIndex,
+        columnStopIndex,
+        columnIndex
+      });
       let style;
 
       // Cache style objects so shallow-compare doesn't re-render unnecessarily.
--- a/source/Grid/index.js
+++ b/source/Grid/index.js
@@ -7,6 +7,7 @@ export type {
   CellSize,
   OverscanIndicesGetter,
   RenderedSection,
+  CellKeyGetterParams,
   CellRendererParams,
   Scroll
 } from "./types";
--- a/source/Grid/types.js
+++ b/source/Grid/types.js
@@ -3,6 +3,17 @@
 import React from "react";
 import ScalingCellSizeAndPositionManager from "./utils/ScalingCellSizeAndPositionManager";
 
+export type CellKeyGetterParams = {
+  rowStartIndex: number,
+  rowStopIndex: number,
+  rowIndex: number,
+  columnStartIndex: number,
+  columnStopIndex: number,
+  columnIndex: number
+};
+
+export type CellKeyGetter = (props: CellKeyGetterParams) => string;
+
 export type CellPosition = { columnIndex: number, rowIndex: number };
 
 export type CellRendererParams = {
@@ -19,6 +30,7 @@ export type CellRenderer = (props: CellRendererParams) => React.Element<*>;
 
 export type CellRangeRendererParams = {
   cellCache: Object,
+  cellKeyGetter: CellKeyGetter,
   cellRenderer: CellRenderer,
   columnSizeAndPositionManager: ScalingCellSizeAndPositionManager,
   columnStartIndex: number,
--- a/source/List/List.js
+++ b/source/List/List.js
@@ -8,9 +8,10 @@ import type {
   OverscanIndicesGetter,
   RenderedSection,
   CellRendererParams,
+  CellKeyGetterParams,
   Scroll as GridScroll
 } from "../Grid";
-import type { RowRenderer, RenderedRows, Scroll } from "./types";
+import type { KeyGetter, RowRenderer, RenderedRows, Scroll } from "./types";
 
 import Grid, { accessibilityOverscanIndicesGetter } from "../Grid";
 import React from "react";
@@ -45,6 +46,8 @@ type Props = {
 
   /** Height constraint for list (determines how many actual rows are rendered) */
   height: number,
+
+  keyGetter: KeyGetter,
 
   /** Optional renderer to be used in place of rows when rowCount is 0 */
   noRowsRenderer: NoContentRenderer,
@@ -100,6 +103,7 @@ export default class List extends React.PureComponent {
   static defaultProps = {
     autoHeight: false,
     estimatedRowSize: 30,
+    keyGetter: ({index}) => `${index}`,
     onScroll: () => {},
     noRowsRenderer: () => null,
     onRowsRendered: () => {},
@@ -195,7 +199,7 @@ export default class List extends React.PureComponent {
   }
 
   render() {
-    const { className, noRowsRenderer, scrollToIndex, width } = this.props;
+    const { className, keyGetter, noRowsRenderer, scrollToIndex, width } = this.props;
 
     const classNames = cn("ReactVirtualized__List", className);
 
@@ -207,6 +211,7 @@ export default class List extends React.PureComponent {
         className={classNames}
         columnWidth={width}
         columnCount={1}
+        cellKeyGetter={this._keyGetter}
         noContentRenderer={noRowsRenderer}
         onScroll={this._onScroll}
         onSectionRendered={this._onSectionRendered}
@@ -256,6 +261,23 @@ export default class List extends React.PureComponent {
 
     onScroll({ clientHeight, scrollHeight, scrollTop });
   };
+
+  _keyGetter = ({
+    rowStartIndex,
+    rowStopIndex,
+    rowIndex,
+    columnStartIndex,
+    columnStopIndex,
+    columnIndex
+  }: CellKeyGetterParams) => {
+    const {keyGetter} = this.props
+
+    return keyGetter({
+      startIndex: rowStartIndex,
+      stopIndex: rowStopIndex,
+      index: rowIndex
+    })
+  }
 
   _onSectionRendered = ({
     rowOverscanStartIndex,
--- a/source/List/types.js
+++ b/source/List/types.js
@@ -2,6 +2,14 @@
 
 import React from "react";
 
+export type KeyGetterParams = {
+  startIndex: number,
+  stopIndex: number,
+  index: number
+};
+
+export type KeyGetter = (params: KeyGetterParams) => string;
+
 export type RowRendererParams = {
   index: number,
   isScrolling: boolean,


bvaughn:
This is cool. Thanks for the back story!

Looking at the `KeyGetter ` snippet you shared, I'm not sure what `indexToBridgeKey` is. Could you elaborate?

Also, do you have any interest in contributing your key-getter to RV as well? Seems like it would be a useful thing to provide as an opt-in for folks with expensive rows.

bvaughn:
Interesting data point: I tried adding a naive version of this into a big `Grid` with lots of little cells (each very lightweight) and performance got dramatically worse.

ashtonsix:
re the snippet - I'm loading rows asynchronously, so they start without unique identifiers (bridge keys). Until a row has loaded it should always get the same React key (identified by index), and then seamlessly transition to using a bridge key once one is available. Mine looks like this:

```js
const indexToBridgeKey = index => {
  const row = this.getRowByIndex(index)
  return row.loaded ? row.data._id : null
}
```

And the whole thing:

```js
class TableBody extends Component {
  constructor() {
    super()
    // blah...
    this.keyGetter = new KeyGetter(indexToBridgeKey)
  }
  // other methods
  render() {
    return (
      <List
        // props...
        keyGetter={this.keyGetter}
      />
    )
  }
}
```

Sure, happy to get a naive `KeyGetter` into RV (maybe the final API could be either a bool which always uses the index or a function [`indexToBridgeKey`])

The perf could probably be improved a whole bunch by focusing on the common case where the index can be used as the bridge key and/or doing some tweaks.

bvaughn:
Hmm. So you're just using a data id as the key instead of the row index?
And that actually makes a perf difference?

I would not expect it to, since I would expect both data-id and row-index
to be stable while a user is scrolling around.

Maybe I'm misunderstanding something? 😊

ashtonsix:
If you remove a row then every row after the deleted row will have a different `index`, but the `id` stays the same. By using the `id` you can avoid some rendering if nested components use `shouldComponentUpdate` (just moving the containers), same bonus for adding and sorting very short lists.

w/o this the animation when adding a new row looks ugly

ashtonsix:
(using react-mirror)

ashtonsix:
motion

bvaughn:
Ah. I see. So it's not scrolling performance it's when you remove rows?

On Aug 30, 2017 7:27 PM, "Ashton" <notifications@github.com> wrote:

motion

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/bvaughn/react-virtualized/pull/789#issuecomment-326171436>,
or mute the thread

bvaughn:
Actually, now that I think about this a bit more- can't you just insert your `key` directly within the `cellRenderer` (without this feature, basically just ignore the incoming `key` param)?

```jsx
const cellRenderer = ({ key, ...rest }) => (
  <div {...rest} key={yourCustomKey}>...</div>
);
```

ashtonsix:
There's a prerequisite for this feature - at the start of each full re-render you need to know which cells were previously rendered & won't show in the next render so you can reuse their keys (or use a heurestic based on `rowIndex` like me).

I started by watching `rowIndex` & releasing some keys whenever the value was less than the previous value - but this only worked when scrolling _down_. I came up with 4 possible solutions:

1. modify `defaultCellRangeRenderer` (non-configurable, & I wanted to iterate w/o having to rebuild constantly/figure out how to run + live rebuild RV's demo)
2. add a hook that runs immediately before `cellRangeRenderer` is called (intrusive, more changes required)
3. add `startRowIndex`, `stopRowIndex` as arguments to `cellRenderer` & watch for changes (unsure of precise caching behaviour inside `defaultCellRangeRenderer`)
4. same as `3` but with a new option (re-renders are still inferred indirectly which is non-ideal)

`1` or `2` might be better solutions but I can't see how to implement this w/o changing RV whatsoever.

ashtonsix:
FYI 2: Using the `data.id` as the key directly is very inefficient as the row remounts once it's finished loading almost halving overall perf.

bvaughn:
> y, the scroll perf improvement (by reusing keys) is the most significant thing here, improving row deletion perf (by mapping ids to react keys) is an added bonus - this solution does both

Hm. That doesn't match what I've observed when trying this with a `Grid`, as I mentioned above. Things got way slower.

I used a counter to create new keys whenever the pool was empty and I used `onSectionRendered` (for `Grid`) to track the currently rendered range of cells so that I'd know when to release keys. It was super slow though.

Can I see your app by chance?

> FYI 2: Using the data.id as the key directly is very inefficient as the row remounts once it's finished loading almost halving overall perf. Instead, this associates an index to a particular react key, & then associates the id to the same react key (via bridge keys) when available.

Ah. Is the added complexity really worth it for the deletion edge case? Maybe your cells really are _that_ expensive and it _is_ worth it. I dunno.

ashtonsix:
Perhaps your naive implementation has a bug? The only perf hit should come from calling `keyGetter`, when I benchmarked `KeyGetter.prototype.activate` (from pastebin) I was getting negligible times. If you're generating new keys when `keyGetter` is called your performance would be predictably awful (for uniform height tables there should be no new keys at all after the first render).

> Can I see your app by chance?

The current version uses more advanced tables than what's publically accessible, but you can see a demo at [demo.bebeyond.co](https://demo.bebeyond.co)

screenshot of demo.bebeyond.co

> Ah. Is the added complexity really worth it for the deletion edge case? Maybe your cells really are that expensive and it is worth it. I dunno.

Totally worth it, fyi some rows in my app can be expanded/collapsed (individual payments for a recurring forecast in a budgeting context) & have a bunch of subrows (most common creation/deletion case). If we're including the default `KeyGetter` this would be much better API for users:

```js
// scroll optimisation
list = <List {...props} pooling={true} />

// scroll optimisation + extras
list = <List {...props} pooling={index => getRow(index)._id} />
```

bvaughn:
> Perhaps your naive implementation has a bug?

Totally possible! 😄  I didn't spend a ton of time on it so I may have made a stupid error.

Now that I look closer, some of the perceived slowdown I was talking about is actually coming from [cell caching](https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/defaultCellRangeRenderer.js#L105-L127). This would requires some thought since key recycling would invalidate some of the assumptions made by this caching.

> The only perf hit should come from calling keyGetter

I don't think this is necessarily true. There's perf overhead in maintaining the key pool. In my test harness, the `Grid` has many small cells (~500 visible at any given point) so there's a lot of key cycling.

I _think_ it's likely that no matter how lightweight the cells are, it shouldn't cost more to update them than to create from scratch, but I haven't thought about it much.

> If you're generating new keys when keyGetter is called your performance would be predictably awful

Only generating when the pool has no more available keys (so it grows to a point and then holds steady).

> If we're including the default KeyGetter this would be much better API for users:

I like the API proposal 😄  I just want to make sure this _actually_ improves performance for other common use-cases.

bvaughn:
Here's my naive pool btw:
```js
function createKeyPool() {
  const availableKeys = [];
  const inUseKeys = {};

  let previousCycle = null;
  let keyCounter = 0;

  function clearColumn(columnIndex) {
    for (let rowIndex = previousCycle.rowStartIndex; rowIndex <= previousCycle.rowStopIndex; rowIndex++) {
      freeKey(rowIndex, columnIndex);
    }
  }

  function clearRow(rowIndex) {
    for (let columnIndex = previousCycle.columnStartIndex; columnIndex <= previousCycle.columnStopIndex; columnIndex++) {
      freeKey(rowIndex, columnIndex);
    }
  }

  function freeKey(rowIndex, columnIndex) {
    const hash = `${rowIndex}-${columnIndex}`;
    const key = inUseKeys[hash];

    if (key !== undefined) {
      availableKeys.push(inUseKeys[hash]);

      delete inUseKeys[hash];
    }
  }

  function onSectionRendered({
    columnOverscanStartIndex,
    columnOverscanStopIndex,
    columnStartIndex,
    columnStopIndex,
    rowOverscanStartIndex,
    rowOverscanStopIndex,
    rowStartIndex,
    rowStopIndex
  }) {
    if (previousCycle !== null) {
      const {
        columnStartIndex: prevColumnStartIndex,
        columnStopIndex: prevColumnStopIndex,
        rowStartIndex: prevRowStopIndex,
        rowStopIndex: prevRowStartIndex
      } = previousCycle;

      for (let i = prevColumnStartIndex; i < columnStartIndex; i++) {
        clearColumn(i);
      }
      for (let i = prevColumnStopIndex; i > columnStopIndex; i--) {
        clearColumn(i);
      }
      for (let i = prevRowStartIndex; i < rowStartIndex; i++) {
        clearRow(i);
      }
      for (let i = prevRowStopIndex; i > rowStopIndex; i--) {
        clearRow(i);
      }
    }

    previousCycle = {
      columnStartIndex: columnOverscanStartIndex,
      columnStopIndex: columnOverscanStopIndex,
      rowStartIndex: rowOverscanStopIndex,
      rowStopIndex: rowOverscanStartIndex
    };
  }

  function cellKeyGetter({
    columnStartIndex,
    columnStopIndex,
    columnIndex,
    rowStartIndex,
    rowStopIndex,
    rowIndex
  }) {
    const hash = `${rowIndex}-${columnIndex}`;
    let key;

    if (inUseKeys.hasOwnProperty(hash)) {
      key = inUseKeys[hash];
    } else if (availableKeys.length > 0) {
      key = availableKeys.pop();
    } else {
      key = keyCounter++;
    }

    inUseKeys[hash] = key;

    return key;
  }

  return {
    cellKeyGetter,
    onSectionRendered
  };
}
```

ashtonsix:
I wrote this snippet to test implementations of `keyPoolGetter`:

```js
var {onSectionRendered, cellKeyGetter} = createKeyPool()
var prevIndex = 0

var VISIBLE_ROW_COUNT = 3
var OVERSCAN_ROW_COUNT = 2

var _onSectionRendered = (startIndex) => {
  const down = startIndex >= prevIndex
  prevIndex = startIndex
  onSectionRendered({
    columnOverscanStartIndex: 0,
    columnOverscanStopIndex: 1,
    columnStartIndex: 0,
    columnStopIndex: 1,
    rowOverscanStartIndex: startIndex - (down ? 0 : OVERSCAN_ROW_COUNT),
    rowOverscanStopIndex:
      startIndex + VISIBLE_ROW_COUNT + (down ? OVERSCAN_ROW_COUNT : 0),
    rowStartIndex: startIndex,
    rowStopIndex: startIndex + VISIBLE_ROW_COUNT
  })
}

var allKeyGetter = (startIndex) => {
  return Array.from(
    {length: VISIBLE_ROW_COUNT + OVERSCAN_ROW_COUNT},
    (x, i) => cellKeyGetter({rowIndex: startIndex + i, columnIndex: 0})
  )
}

var scrollTo = (index) => {
  _onSectionRendered(index)
  console.log(allKeyGetter(index))
}
```

But **500?!** That's a bit more than I expected 😅

Maybe we could drop the id support after all & find an arithmetic solution eg, the key pool is reduced to 4 numbers `xStart`, `xEnd`, `yStart`, `yEnd` we update with `onSectionRendered` & `keyGetter` uses `%` & the current offset to calculate keys.

Doesn't look like any real perf errors in your implementation (however, if you scroll up [`scrollTo(1); scrollTo(0)`] it takes the key second from last instead of last)

bvaughn:
(Disclaimer: I'm at work so I'm a bit distracted; apologies if I misinterpret)

I'm aware of anything based on % because the number of visible rows will change as a user scrolls (by at least 1, perhaps more than 1 if dynamic heights are involved). This can make perf worse. That's why I used a pool that I push and pop from instead, even though it's more work to track and release things.

bvaughn:
I'm going to invite some others to play with this branch too. I'd like more input on perf impact.

bvaughn:
I've released a version of this branch (with a couple of small tweaks) to NPM as `react-virtualized@alpha`

Here's an example of the type of `Grid` I've been testing locally:
https://codesandbox.io/s/4j66o8xp3x

ashtonsix:
Your right - the modulus idea isn't good.

But you could use two simpler pools (horizontal & vertical) so you only have `M + N` instead of `M * N` keys to update - releasing keys would be as simple as splicing each pool from both ends (4 operations). Then combine those in the `keyGetter`.

bvaughn:
Unless I'm picturing that wrong in my head, wouldn't that result in more unique keys than we really want? Maybe you could sketch it out in code or pseudo-code?

ashtonsix:
> Maybe you could sketch it out in code or pseudo-code?

Sure: [codesandbox.io/s/xv3noj7o9o](https://codesandbox.io/s/xv3noj7o9o)

bvaughn:
Ooh! This looks super clever. I like it!

As I scroll around, columns/rows start disappearing. Looks like your pool has the same logic bug mine did initially:

```js
let targetIndex = Math.min(prevStopIndex + 1, startIndex); // Don't do too much work
for (let i = prevStartIndex; i < targetIndex; i++) {
  const key = inUseKeys[i];
  delete inUseKeys[i];
  if (key) { // Don't push undefined
    availableKeys.push(key);
  }
}
targetIndex = Math.max(prevStartIndex - 1, stopIndex); // Don't do too much work
for (let i = prevStopIndex; i > stopIndex; i--) {
  const key = inUseKeys[i];
  delete inUseKeys[i];
  if (key) { // Don't push undefined
    availableKeys.push(key);
  }
}
```

ashtonsix:
Where are the other people you invited?

bvaughn:
Haha, they're all dead 😈 

Nah, I tweeted about it with a link to this issue 😇  haven't gotten any metrics back yet though.

bvaughn:
FWI I created a visual debugger for the key pooler: [bvaughn/react-virtualized/playground/key-debugger.html](https://rawgit.com/bvaughn/react-virtualized/beyond-labs-master/playground/key-debugger.html)

After you scroll a bit, it will print all available and in-use keys.

bvaughn:
I think it's showing a problem with the key pool approach, unless I'm misreading it. The distribution of keys is decent, but each newly visible cell starts off with a render count of 1 (meaning that it was newly-created rather than recycled).

I think that's because the total number of cells isn't constant. At one time there may be N rows and at another time there's N-1. RV isn't specifically hanging onto the pooled row(s) and so the renderers are being destroyed and later recreated.

It's early and it's possible I'm missing or misinterpreting something. I'd love to hear your thoughts.

bvaughn:
I think I need to step away from this a bit and let my brain recharge. 😄 I'm starting to feel like it's going to be overly complicated to do within the constraints of RV's current implementation, but maybe I'm overlooking an obvious solution or making it harder than it needs to be. Probably!

bvaughn:
Sorry to ask again, and I assume the answer is "no" but...doesn't hurt to ask. 😄 

I wonder if I might be able to get access to run, even if it's kind just a rough reproduction, your app code? I've been working on something today that might help but I don't have a great way to evaluate its performance yet.

ashtonsix:
I brought up sharing the app source, the answer is essentially "no" but we came up w/ 3 compromises that could work:

1. sharing just the proprietary table package that wraps RV, but only w/ an NDA & w/o issues from Facebook
2. carrying out experiments you describe ourselves
3. loosely replicating the perf characteristics we have in a separate example:

```js
class Row extends Component {
  shouldComponentUpdate() {
    return false
  }
  render() {
    var d = Date.now()
    while (true) {
      if (Date.now() - d > 10) break
    }
    return <div style={{height: 35}}>Row</div>
  }
}

const SomeTable = () => (
  <List rowHeight={35} rowRenderer={Row} />
)
```

PS: Thanks for all the time you've put into this :) I've got other priorities today, but going to take another stab at this tomorrow

ashtonsix:
I did that stabbing thing & figured it out. You need to release the keys **before** calling `cellRangeRenderer`, that's why it worked in my first example where I was checking for changes in the props passed to `cellKeyGetter`.

[Copy/paste this](https://pastebin.com/6VKirzzD) into `playground/key-debugger` & you'll stop seeing `render: 1`

ashtonsix:
FYI: RV could do with more hooks (even secret undocumented hooks!) - some of the workarounds we've come up with to detect certain events are strange. eg, to detect when RV has finished measuring cells after an update we count calls to `cellMeasurerCache.has`,  and determine whether to call `list.repositionRowsAfter` once every rendered cell has been accounted for.

PS: hope your ok, saw the accident on Twitter

jeremyk:
Hi, I am kind of wondering if this is dead or not as it sounds like it will help out performance on my app.

Also, I have another idea and not sure where best to bring it up so I thought I would through it out here:
A performance optimization I made before on a virtual scroller was to render in each item at its spot initially, but when scrolling only move the outer div. This was an angular 1 app, so a pretty different ecosystem but I did see a big improvement in scrolling performance. One of the ideas was that you could always make the outer div as big as you want and it would have no affect on performance while scrolling as long as you were rendering the same numbers of elements. I think it could help here as the style prop passed to each element wouldn't have to update while scrolling. Does that make sense? I figure I am not thinking of some limitation or you would have done it that way to begin with but I thought I would ask.

bvaughn:
It's kind of stalled. My initial testing showed perf regressions for my test harnesses in some conditions. Ashton couldn't share his app source (understandable). I never really had the chance to try his follow-up suggestion.

bvaughn:
Your suggestion sounds interesting. I don't know if it would help perf in some cases, but I fear it might worsen perf for variable height content. (Since the placement and number of items would change.)

ashtonsix:
FYI we created a private fork of react-virtualised (copy-pasted source into the table package, deleted everything we didn't need and made alterations). We improved key pooling with an `overscanIndicesGetter` that remembers the total number of displayed rows and only changes it by increments of 25% to avoid mounting new rows when alternating between, for example, 29 & 30 visible rows.

jeremyk:
@ashtonsix - I don't suppose you are going to try to get this back in the main line at this point? Too bad if not.

@bvaughn - re: variable height content - variable heights do make it a little tricky but I don't think it should have an impact as you could use your "guess" that you do now for the height until you know for sure. I think the more cells on the screen the more it will help so if you only have a few large cells then yes, it isn't going to help much. But I feel like that case isn't interesting anyway from a performance/stress perspective.

bvaughn:
Variable height rows means either a variable number of children or
over-rendering rows in potentially many cases. Seems like the latter is no
good and the former would (I assume?) undo much of the performance gains
from what you described. Maybe I'm wrong?

robobeau:
Greetings! Has there been any traction with this feature? The company I work for has a use case for this feature. We're using List to display a tree structure as a flat list, and the caching presents a bug. Take this tree structure, for instance:

- Documents
- Standards Manuals
- Assessments

Our key format is composed of 3 parts to ensure uniqueness:

1. The item's indices from the original tree data structure,
2. The object type
3. The object ID

For example, the first document inside `Documents` would be `0-0|Document|1`.

We have 50 documents under `Documents`, and we load 10 at a time (as an example). So, we open `Documents`, and now we have the following structure:

- Documents
-- Document 1
-- yadda yadda yadda...
-- Document 10
- Standards Manuals
- Assessments

As you scroll down through the `Documents` folder, we trigger loading whenever the last rendered document is rendered in the viewport, the tree structure is updated, and it tries to render the new items. What happens (as far as I can tell) is that we get into a state where we've cached `Standards Manuals` (for example) and it tries to render it twice, one for the cached version, and one for the actual new location of `Standards Manuals` once it's been pushed down by the new documents. Since we're using unique keys, React balks at the duplicates and the rendering goes wonky. 

I hope most of that made sense... And thank you for all the hard work on this library. We refactored our tree using ReactVirtualized, and the performance is _dramatically_ better than our Angular version.

ashtonsix:
@robobeau I recommend taking a glance at react-window, the successor to this project. fyi I'm muting this thread (not a problem for me anymore)

PR #792, opened 2017-08-30
Make CellSizeAndPositionManager.getTotalSize() more explicit so it's easier to understand.

ayc92:
- The underlying logic is untouched.
- Adds intermediate variables so it's clear as to what the logic is doing.

My goal here is to make this function more explicit so that it's easier for the next person who reads it to understand what it's doing. My hope is that setting some intermediate variables will accomplish this, but if comments are preferable, or you feel like it's too verbose, I can totally understand.


bvaughn:
Seems reasonable 😄 Thanks!

Issue #780, opened 2017-08-16
Items all dissapear in list


ReinAkane:
# Bug or runtime error
I have verified that this is occurring on master. I'll see about adding a failing unit test tonight. Error is reproducible in `react-virtualized` list example.

1. Open list example in Firefox.
2. Scroll down (until originally visible items + overscan are not visible).
3. Reduce `Num rows` until you don't have a scroll.
4. Increase `Num rows` back to its previous value.
5. Notice that no items are visible.

Scrolling at all immediately brings them back into view.

Gif of reproduction in examples:
---
![virtualized_bug](https://user-images.githubusercontent.com/8212518/29366562-f5a188c4-825f-11e7-89da-95941b96aa53.gif)

edit: updated to specify the bug is only in Firefox.

bvaughn:
I'm not able to reproduce this on the demo site, using Chrome or Safari. It seems to only happen for Firefox. Can you confirm if this is the browser you're using also?

ReinAkane:
Let me try and reproduce in Chrome, as I have done so previously, but the examples reproduction was in Firefox. I'll try in all my browsers and let you know which I can and cannot reproduce in.

ReinAkane:
Can only reproduce in Firefox. Tried in Chrome, IE11, and Firefox all on windows.

bvaughn:
That's interesting.

bvaughn:
I should add that my testing above was on Mac OS. So it seems like this issue affects Firefox on both platforms, but no other browsers.

ReinAkane:
Working on a fix, hope to have a PR soon. The problem stems from the fact that when a container is shortened past the current scroll location Chrome fires a scroll event but Firefox does not. The fix we're going to try and make is to calculate our location when props change.

bvaughn:
Great! Thanks for the update.

ReinAkane:
So I've spent about two hours in a futile attempt to get tests to run at all. Something deciding not to babelify. I spent 30 minutes trying to get yarn to work, on a similarly futile attempt. Here's the commit that fixes this issue: https://github.com/ReinAkane/react-virtualized/commits/fix-items-all-disappear-in-list-780
If I remember to, I will try to test when I get home.

bvaughn:
> So I've spent about two hours in a futile attempt to get tests to run at all.

```sh
yarn install
yarn test
```

ReinAkane:
[...]

PR #798, opened 2017-09-01
Scroll when container contracted to hide current scroll

ReinAkane:
Fixing issue #780: scrolling then reducing list size could cause all items to disappear in Firefox. See issue for a GIF.

The root cause is that Firefox does not fire a scroll event when a scrolled element's size decreases. This causes the Grid's `scrollTop` and `scrollLeft` values to be out of sync with the actual scroll location of the element.

This PR makes the Grid check for item count changes to determine if it needs to scroll the grid back into view.

### DIFF
--- a/source/Grid/Grid.jest.js
+++ b/source/Grid/Grid.jest.js
@@ -105,6 +105,78 @@ describe("Grid", () => {
       expect(rendered.textContent).not.toContain("column:0");
       expect(rendered.textContent).not.toContain("row:0");
     });
+
+    it("should scroll to the last existing point when rows are removed", () => {
+      const grid = render(
+        getMarkup({
+          rowCount: 15
+        })
+      );
+
+      simulateScroll({
+        grid,
+        scrollTop: 200
+      });
+
+      const updatedGrid = render(
+        getMarkup({
+          rowCount: 10
+        })
+      );
+
+      expect(updatedGrid.state.scrollTop).toEqual(100);
+    });
+
+    it("should scroll to the last existing point when columns are removed", () => {
+      const grid = render(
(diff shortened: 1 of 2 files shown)


bvaughn:

ReinAkane:
@mcordova47 I'm not going to have time to test for a while, sadly. If you want to pull it down and test yourself, I added unit tests to ensure that the bug this PR fixed continues to be fixed.

bvaughn:
I think this introduced the regression described in #816 and bvaughn/react-virtualized-select/issues/75

Happy to review a fix if you'd like to put up a PR @mcordova47

mcordova47:
@bvaughn @ReinAkane I looked a little further into it, and the problem I am facing is a little different than the problem described in #816.  It seems to be a problem with the `MultiGrid` component that was exposed by this commit.  `MultiGrid` has no default value for `scrollToRow`, and it sets its bottom right grid's `scrollToRow` to `props.scrollToRow - props.fixedRowCount`, causing it to be `NaN` when it should default to `-1`.

Adding defaults of `-1` to `scrollToRow` and `scrollToColumn` in `MultiGrid` seems to fix my issue.  I created a pull request  (#820), if you'd like to review that.  Some tests for other components were failing before my fix, but all tests for `MultiGrid` pass.  Added two tests that were failing before the fix.

mcordova47:
@bvaughn Actually, looks like #816 could also be fixed by setting `scrollToRow` and `scrollToColumn` defaults to `-1` instead of `0`.  I could be misunderstanding something, but to me it makes sense that the `ArrowKeyStepper` keeps jumping to the first row/column since the default values are both `0`.

PR #801, opened 2017-09-03
Add childrenClassName for inner Grid element 

jesuGMZ:
Using a List element is not possible to attach a custom CSS class name for the Grid element that is being used internally. This solution allows it.

Example
```
<List
  className="wraper"
  childrenClassName="items"
  height={height}
  width={width}
  ref={registerChild}
  rowCount={this.state.animals.length}
  rowHeight={rowHeight}
  rowRenderer={this._rowRenderer}
/>
```
will generate
```
<div aria-label="grid" aria-readonly="true" class="ReactVirtualized__Grid ReactVirtualized__List wraper" role="grid" tabindex="0" style="box-sizing: border-box; direction: ltr; height: 915px; position: relative; width: 723px; will-change: transform; overflow: auto;">
  <div class="ReactVirtualized__Grid__innerScrollContainer items" role="rowgroup" style="width: auto; height: 2400px; max-width: 723px; max-height: 2400px; overflow: hidden; position: relative;">
    <div class="item">Item value</div>
  </div>
</div>
```

Hope could be useful :)

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -85,6 +85,9 @@ type Props = {
   /** Optional custom CSS class name to attach to root Grid element.  */
   className?: string,
 
+  /** Optional custom CSS class name to attach to inner children Grid element. */
+  childrenClassName?: string,
+
   /** Number of columns in grid.  */
   columnCount: number,
 
@@ -857,6 +860,7 @@ export default class Grid extends React.PureComponent {
       autoHeight,
       autoWidth,
       className,
+      childrenClassName,
       containerProps,
       containerRole,
       containerStyle,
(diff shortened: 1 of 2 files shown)


bvaughn:
Thanks for submitting a PR! I hope you'll understand my reluctance to add _yet another_ prop to `Grid` though. There are so many already and each one I accept pretty much needs to be supported forever. 😄

The `containerStyle` prop is necessary in order to override the scroll-container's inline styles, but the `className` prop, coupled with a descendant `. ReactVirtualized__Grid__innerScrollContainer` selector, is sufficient to support CSS-powered overrides for this element.

For that reason, I prefer to pass on this new property. Again, I hope you'll understand.

Cheers!

jesuGMZ:
Problems comes when you have the need to use a selector to apply CSS style to the inner items. In case you are using a CSS framework for example (I am with Semantic UI) is not possible to solve the CSS children selector.

I understand anyway your decision, and thanks for this project :)

albinotonnina:
I'm in the same situation of @jesuGMZ

albinotonnina:
Too bad 🙁

Issue #812, opened 2017-09-18
Warning: Failed prop type: Invalid prop `containerStyle` of type `function` supplied to `Grid`, expected `object`


willemx:
I'm getting this warning after upgrading to RV 9.10.0
Looking through the code in `Grid.js` I see `containerStyle: Object` as a default prop.
I have never used default props in my code myself, but should this not be: `containerStyle: {}` ?

TrySound:
Yep. You are right.

bvaughn:
Thanks! I fixed this but my fix was in a branch ☹️  Realized on the ride to work this morning. Glad Bogdan posted a fix 😄

PR #813, opened 2017-09-18
Fix Grid.defaultProps

TrySound:
Ref https://github.com/bvaughn/react-virtualized/issues/812


bvaughn:
Thanks buddy.

bvaughn:
Check out 9.10.1. This fix should be in there.

Issue #815, opened 2017-09-18
tabIndex error on MultiGrid component - version 9.10.1


kupkovski:
After upgrading to version 9.10.1, I started getting the following error on using `<MultiGrid>` component:
```
Warning: Failed prop type: The prop `tabIndex` is marked as required in `Grid`, but its value is `null`.
    in Grid (created by MultiGrid)
    in MultiGrid (at MyGrid.jsx:258)
    in div (created by AutoSizer)
    in AutoSizer (at MyGrid.jsx:256)
```

The problem is `<MultiGrid>` creates all their `<Grid>`s setting the `tabIndex` prop as `null` on the following lines:
- https://github.com/bvaughn/react-virtualized/blob/9.10.1/source/MultiGrid/MultiGrid.js#L636
- https://github.com/bvaughn/react-virtualized/blob/9.10.1/source/MultiGrid/MultiGrid.js#L690
- https://github.com/bvaughn/react-virtualized/blob/9.10.1/source/MultiGrid/MultiGrid.js#L728

But `<Grid>` component allows only numbers:
- https://github.com/bvaughn/react-virtualized/blob/9.10.1/source/Grid/Grid.js#L211

Creating a `<MultiGrid>` passing the `tabIndex` prop as a number does not help, because the component itself always creates the Grids using `null`.

bvaughn:
A PR would be welcome if you have the time

PR #818, opened 2017-09-20
Make Grid.props.tabIndex: ?number

kalley:
Fixes: https://github.com/bvaughn/react-virtualized/issues/815

Since `MultiGrid` sends `null` for `tabIndex`, this just adds it as a possible value in the flow props, which get compiled into `propTypes` for distribution.

Note: Reverted my previous update since it seems we just don't want focus on those grids at all.

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -11,52 +11,52 @@ import type {
   ScrollbarPresenceChange,
   RenderedSection,
   OverscanIndicesGetter,
-  Alignment,
-} from './types'
-import type { AnimationTimeoutId } from '../utils/requestAnimationTimeout'
+  Alignment
+} from "./types";
+import type { AnimationTimeoutId } from "../utils/requestAnimationTimeout";
 
-import React from 'react'
-import cn from 'classnames'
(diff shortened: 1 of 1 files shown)


source/Grid/Grid.js
```diff
@@ -208,7 +208,7 @@ type Props = {
   style: Object,
 
   /** Tab index for focus */
-  tabIndex: number,
+  tabIndex: number | null,
```
bvaughn: I think the typical Flow syntax to use for this is:
```js
tabIndex: ?number,
```

bvaughn:
Looks like d6704d6 included some unintentional formatting changes also? 😄

kalley:
oops... all fixed up.

bvaughn:
Thanks!

ahutchings:
@bvaughn would you be open to cutting a release to include this? emojione-picker is also running into the issue: https://github.com/tommoor/emojione-picker/issues/77

bvaughn:
@ahutchings It was included in the [9.11.0 release](https://github.com/bvaughn/react-virtualized/blob/master/CHANGELOG.md#9110) this morning.

ahutchings:
Awesome, thanks!

Issue #822, opened 2017-09-25
Problems with scrolling


yuristrelets:
I am trying to use `List`, something like this
```
<AutoSizer>
  {({ height, width }) => (
    <List
      width={width}
      height={height}
      rowHeight={48}
      rowRenderer={({ index }) => (
        <div style={{ height: 48 }} key={index}>Item {index}</div>
      )}
      rowCount={rows.size}
    />
  )}
</AutoSizer>
```
But instead of smoothly scrolling I get terribly jerky on chrome 61.0.3163.100 https://www.screencast.com/t/I7yRYcu71
on FF 55.0.3 I had empty space at the end https://www.screencast.com/t/JOT1Bswz

Without `AutoSizer` the problem doesn't disappear. 
I'm using the latest versions of react and react-virtualized.

What am I doing wrong? 
Who faced such a problem?

christiantinauer:
Same problem here. Introduced with 9.10.0. Downgrading to 9.9.0 fixes the problem.

istarkov:
Same here

yuristrelets:
Unfortunetely, even rolling back to version 9.9.0, and then 9.0.0 and 8.11.4, didn't solve the problem. I used version 7 for a long time and wanted to try the latest version in the new project. On version 7 there is no such problem. I have no ideas.

TrySound:
@yuristrelets That's because you should pass style and key (not index) props to your div
```js
      rowRenderer={({ style, key, index }) => (
        <div style={{ height: 48 }} style={style} key={key}>Item {index}</div>
      )}
```

yuristrelets:
@TrySound You're right! My fault, I didn't see that it's necessary to pass `style` ad `key`. Downgrading to version 9.9.0 and changing `rowRenderer` solved the problem. Thank you!

mtycvy:
I have experienced this problem with v9.10 downgrading to 9.9.0 fixed the issue.

bradbirnbaum:
We are seeing the same issue on 9.10.1 as well.

istarkov:
Guys here is the fix https://github.com/bvaughn/react-virtualized/pull/825

bvaughn:
Fixed by #825

Will be released momentarily as 9.11

w10036w:
react-virtualized ^9.20.1 still has this issue.
After add some dummy style it is resolved. 
Please reopen this issue or at least mark "style" as a compulsory parameter to avoid this.

env: 
[...]

PR #825, opened 2017-09-26
Fix scrollTo issue

TrySound:
Ref #822 

Done, but there should be a test case I guess.
/cc @istarkov


istarkov:
mb add test on `updateScrollIndexHelper` checking that it's not causes updates if rowHeight is a function?

Mr-Wallet:
This is a big pain point for my team because we need fixes in 9.10.0 but we need to also not have #822 ... is this still moving?

bvaughn:
Looks like this fixes bvaughn/react-virtualized-select/issues/75

Thanks buddy 👍

PR #833, opened 2017-10-02
Adding in the defaultSortOrder for the column. Currently only SortDirection.ASC is the uneditable default.

mbseid:
**Motivation**
Currently, when a user selects a column to sort, the default direction is hard coded to `SortDirection.ASC`. It would be great to have this as a configurable option so that each column can have their own default sorting direction.

**Implementation** 
I added a prop to the `Column` object and updated the `_createHeader` function to use the column's information to handle the sorting click.

Side Note: I added a test to ensure that it works and also added the new prop to the docs also.

Let me know what you think and thanks for the awesome library!

### DIFF
--- a/source/Table/Column.js
+++ b/source/Table/Column.js
@@ -4,6 +4,7 @@ import { Component } from "react";
 import defaultHeaderRenderer from "./defaultHeaderRenderer";
 import defaultCellRenderer from "./defaultCellRenderer";
 import defaultCellDataGetter from "./defaultCellDataGetter";
+import SortDirection from "./SortDirection";
 
 /**
  * Describes the header and cell contents of a table column.
@@ -33,6 +34,9 @@ export default class Column extends Component {
 
     /** Uniquely identifies the row-data attribute corresponding to this cell */
     dataKey: PropTypes.any.isRequired,
+
+    /** Optional direction to be used when clicked the first time */
+    defaultSortDirection: PropTypes.oneOf([SortDirection.ASC, SortDirection.DESC]),
 
     /** If sort is enabled for the table at large, disable it for this column */
     disableSort: PropTypes.bool,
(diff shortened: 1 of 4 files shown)


source/Table/Table.js
```diff
@@ -518,10 +519,13 @@ export default class Table extends PureComponent {
...
-      const newSortDirection =
-        sortBy !== dataKey || sortDirection === SortDirection.DESC
-          ? SortDirection.ASC
-          : SortDirection.DESC;
+      const isFirstTimeSort = sortBy !== dataKey;
+
+      // If this is the firstTime sort of this column, use the column default sort order OR the default, asc.
+      // Otherwise, invert the direction of the sort.
+      const newSortDirection = (isFirstTimeSort ?
+      				(defaultSortOrder || SortDirection.ASC) :
```
bvaughn: We could get around this || if we set the default value via `static defaultProps`

source/Table/Column.js
```diff
@@ -34,6 +35,9 @@ export default class Column extends Component {
     /** Uniquely identifies the row-data attribute corresponding to this cell */
     dataKey: PropTypes.any.isRequired,
 
+    /** Optional direction to be used when clicked the first time */
+    defaultSortDirection: PropTypes.oneOf([SortDirection.ASC, SortDirection.DESC]),
```
bvaughn: This is referenced as `defaultSortOrder` everywhere else but `defaultSortDirection` here.

I should have caught this during the initial pass. Whoops. I'll fix in master...

source/Table/Table.js
mbseid: True. Good point.

source/Table/Column.js
mbseid: Good catch. Not sure how I missed it. 

I saw you fixed it here: https://github.com/bvaughn/react-virtualized/commit/6cac6be834b65a8aa689798a404e2387e4aaf69e

Thanks so much!

mbseid:
Thanks so much! Greatly appreciate the awesome work and the library.

bvaughn:
You're welcome 👍

PR #840, opened 2017-10-10
Support for massive datasets in CellSizeAndPositionManager

rakeljane:
We have a (somewhat ridiculous) use case for Grid. I'm finding that chrome runs out of memory when scrolling through sets with around 1 billion rows. 
It fails in `getSizeAndPositionOfCell` on the loop that caches the size of every row up to the new position, this behavior seems unnecessary when the height of rows is consistently known or fixed.

Passing a callback through to `CellSizeAndPositionManager` that will provide "range" sizes is one solution I've found that seems to be working. 
A cleaner solution might be to have `CellSizeAndPositionManager` toggle it's logic based on knowing that cell sizes are fixed (return the fixed size * number of rows). I played with the callback solution first just because it's a simpler implementation for grids with fixed heights on everything but the header. 

Open to suggestions!


rakeljane:
Saw your tweets, figure I'd provide some more insight into the use case. 
Anything over 10k rows doesn't provide a whole lot of utility, totally agree. "Displaying" all 1B+ rows is purely a way of showing the user that all of the data is there and they can filter on it to get whatever they need. 
The client never holds onto more than it needs for the current display window.

bvaughn:
> Saw your tweets, figure I'd provide some more insight into the use case.

Oh, whoops! Sorry if my tweet seemed like I was poking fun at this issue. I didn't mean to!

I was just really amused that someone was pushing RV that hard. I don't think I've even tried to do _that_ 😁

> (also edited the title, didn't mean this to come across as a bug accusation, it's definitely a feature request)

No worries. I didn't read it as an accusation, just an extreme edge-case.

PR #844, opened 2017-10-15
Update Flow/Prettier/ESlint

bvaughn:
I find the current pre-commit hooks super annoying. This PR changes things to be similar to [the setup I use for reactjs/reactjs.org](https://github.com/reactjs/reactjs.org/blob/master/.circleci/config.yml), but with automated tests (similar to the config file [here](https://circleci.com/docs/2.0/sample-config/)).

Let's see if CircleCI picks up the new configuration correctly.


bvaughn:
This is failing on CI but I'm not sure why b'c Circle seems to be having some problems right now. Will check back once they've got it resolved.

https://twitter.com/brian_d_vaughn/status/919594877729378305

PR #846, opened 2017-10-17
fix item size on re-measuring

ookami-kb:
If the style width or height is set to 'auto', it will override values, that were probably defined in css for this item. Setting it to null allows to use default width/height of the item.


bvaughn:
Can't merge this as-is. If you're interested in collaborating more on supporting this new edge-case, let's talk more! I'm going to close this PR for now though 👍

PR #855, opened 2017-10-22
Bump React peer dep to 16.

milesj:
The React peer dep still points to 16-alpha, which causes unmet peer deps for 16 proper. This fixes it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,11 +144,11 @@
     "classnames": "^2.2.3",
     "dom-helpers": "^2.4.0 || ^3.0.0",
     "loose-envify": "^1.3.0",
-    "prop-types": "^15.5.4"
+    "prop-types": "^15.6.0"
   },
   "peerDependencies": {
-    "react": "^15.3.0 || ^16.0.0-alpha",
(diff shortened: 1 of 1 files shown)


bvaughn:
> The React peer dep still points to 16-alpha, which causes unmet peer deps for 16 proper.

I'm curious what makes you say that?

The semver constraint `^16.0.0-alpha` should match all versions 16.0.0-alpha*, 16.0.0-beta*, 16.0.0-rc*, and 16.0.0+.

These are some handy online tools for verifying:
* http://jubianchi.github.io/semver-check/
* https://semver.npmjs.com/

RV's current react peer deps match the following versions:

screen shot 2017-10-24 at 9 40 52 am

Your proposed change would drop support for alpha, beta, and RC releases. This probably wouldn't matter _that much_ but I don't want to drop support unnecessarily unless it's part of a major.

milesj:
Sorry, must be a yarn issue then, as it would spew "unmet peer" warnings.

bvaughn:
Might be coming from another source?

PS You can tell Yarn which resolution to use by adding a `resolutions` block to `package.json`.

PR #874, opened 2017-11-14
Pull request fix scroll top on non scrolling container wrong children to render

uglywizkid:
Hi there,

We are currently using this library (especially the AutoSizer and Table component) in our application. In our application we are saving the current scroll position in the URL in order to share links. This way we encountered an issue when there is a scroll position set, but the whole list actually fits into the screen/container. In this case it can happen, that it does not render all items, because it begins to render from the nearest item of the set scroll position. We expect it should render all children from the beginning (start 0) if all items fit into the container nevertheless what scroll position (scrollTop) has been set. This is what I tried to fix with this pull request.


source/Grid/utils/CellSizeAndPositionManager.js
```diff
@@ -225,8 +225,13 @@ export default class CellSizeAndPositionManager {
       return {};
     }
 
+    let start;
+    if (totalSize < containerSize) {
+      start = 0;
+    } else {
+      start = this._findNearestCell(offset);
+    }
```
bvaughn: I'm not sure this is the correct place to fix this. `totalSize` could be estimated (eg if `CellMeasurer` is also being used) and therefor not totally accurate. It's a bit of an edge-case but I think it could introduce a problem.

bvaughn: Guarding against this case is why the "scroll" handler in `Grid` has the following check:
```js
scrollLeft = Math.min(
  Math.max(0, totalColumnsWidth - width + scrollbarSize),
  scrollLeft,
);
scrollTop = Math.min(
  Math.max(0, totalRowsHeight - height + scrollbarSize),
  scrollTop,
);
```

bvaughn:
I don't think I feel comfortable introducing more checks/code into something as performance-sensitive as the "scroll" handler code path to cover an edge-case like this (eg a user programmatically setting a scroll offset that's way larger than the container it's in).

The browser prevents the scroll offset from actually going past 0 in this case.

I think in this case it seems most reasonable for your code not to specify an invalid offset. Do you think this is unreasonable? I'm happy to discuss it more if you strongly disagree.

source/Grid/utils/CellSizeAndPositionManager.js
uglywizkid: Yes, but this code is inside handleScrollEvent. But if everything fits into the screen, this code will never be executed.

uglywizkid:
I think we really need a solution here. Our application is bookmarkable, so the scroll position is encoded into the URL. This is very important for us. So it is very likely to have an invalid scroll position, because the list grows an shrinks daily.

Can you provide any other solution for us? Perhaps an exposed function to validate the scroll position?

PR #877, opened 2017-11-16
Introduce Column headerStyle

mucsi96:
Hello this is my first contribution in this repo! 🎉 

I miss the ability to provide extra styles for each column header.
Currently it is possible to provide only extra className for each column header.

I need this for adding IE support for react-virtualized where I pass the width for every column and its header as fallback solution for flex-box.

I am open to submit a PR with IE support for Table if you wish later.


bvaughn:
This feautre will be released with 9.13.0 shortly.

PR #879, opened 2017-11-20
reverse using fixed option of cellMeasureCache

hiradimir:
I have performance issue of MultiGrid with CellMeasure.

I believe the [comment](https://github.com/bvaughn/react-virtualized/blob/656033edec3e33c89a468643ca861625fc5ade6f/source/Grid/utils/CellSizeAndPositionManager.js#L211-L213). But the grid behavior is different from my wish. 
> // eg we can't know a row's height without measuring the height of all columns within that row.

I am thinking Grid can know row's height without measuring for cellMeasureCache with fixedHeight.

I think bug is [here](https://github.com/bvaughn/react-virtualized/blob/656033edec3e33c89a468643ca861625fc5ade6f/source/Grid/Grid.js#L318-L334).
Here is [sample](https://plnkr.co/edit/rypMdjS44EzEsxp6il1D?p=preview) 

I fixed the issue by this PR.

### DIFF
--- a/source/Grid/Grid.jest.js
+++ b/source/Grid/Grid.jest.js
@@ -2328,6 +2328,62 @@ describe('Grid', () => {
     expect(keys).toEqual(['0-0', '1-1']);
   });
 
+  describe('getVisibleCellRange', () => {
+    it('should return all cell-range of row when a :deferredMeasurementCache is provided without fixedHeight', () => {
+      const cache = new CellMeasurerCache({
+        fixedWidth: true,
+      });
+      const grid = render(
+        getMarkup({
+          deferredMeasurementCache: cache,
+        }),
+      );
+
+      expect(grid._columnSizeAndPositionManager.batchAllCells).not.toBe(true);
+      const columnRange = grid._columnSizeAndPositionManager.getVisibleCellRange({
+        containerSize: DEFAULT_WIDTH,
+        offset: 500,
+      });
+      expect(columnRange.start).not.toEqual(0);
+      expect(columnRange.stop).not.toEqual(NUM_COLUMNS - 1);
+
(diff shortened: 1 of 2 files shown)


bvaughn:
This PR completely crushes the `CellMeasurer` demo page (eg visit /components/CellMeasurer and try to scroll down). The `Grid` becomes blank for a couple of seconds before rendering a new section of cells.

bvaughn:
I believe this PR reflects a misunderstanding.

The height of a row _must_ be equal to the height of the tallest cell in that row, and the width of a column _must_ be equal to the width of the largest cell in that column. Otherwise things would overflow in a way that we don't want. We can't just measure as a user scrolls and gradually adjust, because it would be a very poor user experience for rows and columns to shift around and change size as you scroll. So that's the intent behind this section of code.

I agree that it looks a bit confusing for the `_columnSizeAndPositionManager` to check `deferredMeasurementCache.hasFixedHeight` rather than `deferredMeasurementCache.hasFixedWidth`. Unfortunately, this is kind of an awkward side effect of the way the code evolved over time. The only thing the `batchAllCells` param is used for in `CellSizeAndPositionManager` is to determine [how many cells to measure (by rendering them)](https://github.com/bvaughn/react-virtualized/blob/b446761b13cbf142e71de106cf799c0962bc0397/source/Grid/utils/CellSizeAndPositionManager.js#L210-L218) when rendering children.

Since we need to know the height of all columns in a row in order to know the row's height, and since the `batchAllCells` param passed to `_columnSizeAndPositionManager` determines how many columns we end up rendering, this is why it checks to see if the row-height is fixed or dynamic (eg `hasFixedHeight`).

I hope this explanation helps. 😄 I admit this code is a bit confusing even to me sometimes.

PR #881, opened 2017-11-21
Added missing documentation, Lists do support overscanIndicesGetter

benwill:
Hopefully the title explains all, I was looking to override the List components overscanIndicesGetter behaviour as documented by the grid, as a new user to this, it looked like it wasn't possible. After looking at the code and trying it though it worked.. So hopefully this helps people in the future!


bvaughn:
Hm! I certainly empathize with you and apologize for the confusion. I"m not sure I feel about making this change though.

The [List docs](https://github.com/bvaughn/react-virtualized/blob/master/docs/List.md#list) says:
> That means that `List` also accepts [`Grid` props](https://github.com/bvaughn/react-virtualized/blob/master/docs/Grid.md) in addition to the props shown below.

I did this so I didn't have to copy-paste every `Grid` prop into the `List` and `Table` docs as well. (That would make docs even harder to maintain.)

benwill:
Fair enough, just theres already a lot of duplication there

bvaughn:
I wish there was a better way to auto-generate these docs. One of my coworkers, @mitermayer, is working on something that might be helpful for this.

PR #882, opened 2017-11-22
Add virtualized tree component

Lodin:
This PR adds a new component for rendering tree structures: `Tree`. 

`Tree` is built on top of [`Grid`](https://github.com/bvaughn/react-virtualized/blob/master/docs/Grid.md) component. It uses `Grid` rows to display tree nodes as a simple list, but also takes control of nodes opening and closing. It allows end user to avoid additional logic to control this behavior. 

To render nodes conditionally `Tree` uses generator function `nodeGetter` ([detals](https://github.com/Lodin/react-virtualized/blob/tree/docs/Tree.md#nodegetter)). It creates an array of currently opened and visible nodes and sends it to grid. If user opens or closes any node, function restarts, and creates new array to display.

You can read documentation for this component [here](https://github.com/Lodin/react-virtualized/blob/tree/docs/Tree.md). PR also contains example usage of `Tree`. Also there is online [proof of concept](https://lodin.github.io/react-tree-virtualized/).

### DIFF
--- a/source/Tree/Tree.example.js
+++ b/source/Tree/Tree.example.js
@@ -54,6 +54,7 @@ export default class TreeExample extends React.PureComponent {
 
     if (dynamicRowsHeights !== prevDynamicRowsHeight) {
       this.tree.forceUpdate();
+      this.tree.forceUpdateGrid();
     }
   }
 
--- a/source/Tree/Tree.example.js
+++ b/source/Tree/Tree.example.js
@@ -1,5 +1,17 @@
 import React from 'react';
+import {AutoSizer} from '../AutoSizer';
+import {
+  ContentBox,
+  ContentBoxHeader,
+  ContentBoxParagraph,
+} from '../demo/ContentBox';
 import Tree from './Tree';
+import {InputRow, LabeledInput} from '../demo/LabeledInput';
+import styles from './Tree.example.css';
+
+function getRandomInt(min, max) {
+  return Math.floor(Math.random() * (max - min)) + min;
+}
 
 export default class TreeExample extends React.PureComponent {
   static overscanIndicesGetter({startIndex, stopIndex}) {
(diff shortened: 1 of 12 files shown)


bvaughn:
This looks cool. 😄

I'll try to find some time to review it.

source/demo/index.js
```diff
@@ -12,6 +12,4 @@ import Application from './Application';
 render(<Application />, document.getElementById('root'));
 
 // Import and attach the favicon
-document.querySelector(
-  '[rel="shortcut icon"]',
-).href = require('file!./favicon.png');
+document.querySelector('[rel="shortcut icon"]').href = require('./favicon.png');
```
bvaughn: This change (removing the `file!` prefix) breaks the loader for the favicon.
screen shot 2017-11-26 at 8 59 58 am

I'm curious, was this an accident?

source/styles.css
```diff
@@ -76,3 +76,26 @@
...
+  display: flex;
+  align-items: center;
+}
+
+.ReactVirtualized__Tree__node > :first-child {
+  margin-right: 10px;
+  width: 10px;
+}
+
+.ReactVirtualized__Tree__node button {
```
bvaughn: The way RV treats styles is this:
* Functional styles (things that are important to the usability of a component) are defined as inline styles. This means they work even if the user doesn't want to import RV's default styles.
* Presentational styles are most avoid, with the exception of the `Table` component, in which case they're defined in this stylesheet.
* User can provide additional (external) CSS styles by specifying a `className` property.
* User can provide- even _override_ inline styles- by specifying a `style` property.

The `Tree` component styles defined in this CSS file are a mix of presentational and functional. Things like the border style are more presentational (and I slightly prefer we just stuck with the browser defaults for this sort of thing). But eg the width of the button and the margin are very much functional, because if the stylesheet is missing or removed, you end up with this:
screen shot 2017-11-26 at 9 12 21 am

So I think the functional styles should be moved inline and the remaining presentational styles should either live in this file or just go away entirely and be up to the end user to specify.

source/Tree/Tree.example.js
```diff
@@ -0,0 +1,238 @@
...
+import Tree from './Tree';
+import {InputRow, LabeledInput} from '../demo/LabeledInput';
+import styles from './Tree.example.css';
+
+function getRandomInt(min, max) {
+  return Math.floor(Math.random() * (max - min)) + min;
+}
+
+export default class TreeExample extends React.PureComponent {
+  static overscanIndicesGetter({startIndex, stopIndex}) {
```
bvaughn: I'm curious- is this necessary?

This is kind of an advanced property that I don't expect to be used very often, so I'm surprised to see it in the demo. It should almost always be sufficient to specify `overscanRowCount`.

Just curious if this hints at some performance issue or limitation?

source/Tree/Tree.example.js
```diff
@@ -0,0 +1,238 @@
...
+      nestingLevel: 0,
+      node: this.root,
+    });
+
+    while (stack.length !== 0) {
+      const {node, nestingLevel} = stack.pop();
+
+      const isOpened = yield {
+        childrenCount: node.children.length,
+        height: dynamicRowsHeights ? getRandomInt(30, 60) : undefined,
```
bvaughn: The dynamic height should probably come from the `context.list` items, each of which defines a `size`. It doesn't necessarily have to _be_ that size but it should be _based_ on that size so it doesn't change each time the function is called.

source/Tree/Tree.js
```diff
@@ -0,0 +1,439 @@
...
+    const {
+      className,
+      noRowsRenderer,
+      scrollToIndex,
+      width,
+    } = this.props;
+
+    const classNames = cn('ReactVirtualized__Tree', className);
+
+    const rowCount = this._prepareNodes();
```
bvaughn: Creating an array with thousands of objects seems like a potentially expensive thing to compute each `render` (eg any time a `Tree` prop like `width` or `height` changes, any time a node is expanded/collapsed).

I'm not super familiar with this component, but shouldn't it be possible to move this pre-compute step into `componentWillMount` and `componentWillReceiveProps` (where we could only recompute it if it seems likely to have changed- eg not in response to a changed `width` or something)?

Would it be possible to only apply a patch to the computed nodes when a single node is collapsed/expanded (rather than recomputing the entire tree)?

source/demo/index.js
Lodin: Yes, sorry. I had problems with autoreloading in webpack 1, so I changed webpack configuration to use webpack 3. This line left in the PR was an accident. I'll get it back in the next commit

source/Tree/Tree.example.js
Lodin: Hmm... That's strange. I have no idea why this code is here. I think it would be better if I remove it.

source/styles.css
Lodin: Good point. I'll fix it.

source/Tree/Tree.js
Lodin: Yes, creating an internal array was a problem I've stuck from the very start. Thanks for a fresh idea, I'll think about it.

Lodin:
@bvaughn Actually, I [started](https://github.com/Lodin/react-tree-virtualized) with a separate component. However, then I thought that it would be better for users to have a very basic tree component along with other `react-virtualized` components, so I've prepared this PR. And I guess I wanted new ideas about several solutions I've made. 

But if you think it should be a separate component, I'm not against this idea.

bvaughn:
Thanks for being so understanding.

My concern about this is that- once the component was merged into the main react-virtualized library, it would become my responsibility to maintain (answering GitHub issue questions about it, responding to feature requests, etc.) I don't have the bandwidth as it is to stay on top of issues for this library (as my slow PR response time has indicated- and this was a vacation week, when I have more time than usual).

I'd be happy to accept a PR that adds some mentions and links to the tree component, if you decide to release it as a standalone component.

Lodin:
@bvaughn Ok, then I'll finish my component and add new PR with links to it. Thanks for your time!

PR #887, opened 2017-11-27
Really fix coverage 😳

rickhanlonii:
## Overview

In the [previous PR](https://github.com/bvaughn/react-virtualized/pull/886) we forgot to send the results to codecov

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
           key: dependencies-{{ checksum "yarn.lock" }}
       - run:
           name: Test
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           key: dependencies-{{ checksum "yarn.lock" }}
       - run:
           name: Test
-          command: yarn run test:ci
+          command: yarn test:ci
```
rickhanlonii: Updated based on feedback 👌

Issue #891, opened 2017-12-04
server rendering with WindowScroller


landvibe:
I'm using next.js for server rendering.
Below code has a console warning
`Warning: Expected server HTML to contain a matching <div> in <div>.`
I think client rendering is failed with this warning, so the List has no row.
If I change WindowScroller to AutoSizer, then there is no console warning.
I don't know what causes this warning.

![image](https://user-images.githubusercontent.com/3401416/33539601-979a4930-d90a-11e7-859e-07f1a41b409e.png)

If I skip server rendering like below, it also has the warning but the rows are rendered correctly on client side.
```
render() {
    if (!IS_CLIENT) {
      return null;
    }
    ...
```

bvaughn:
`WindowScroller` depends on a `window` object, else it [falls back to a width and height of 0](https://github.com/bvaughn/react-virtualized/blob/43b84762ccb242c2048703463df570ff884ff167/source/WindowScroller/WindowScroller.js#L56-L58). I guess we could add `serverWidth`/`serverHeight` props, but it's possible that the server and client markup would still different based on the actual size of the client window.

**Edit** Looks like you already opened a PR for this. (Thanks!) I don't mind making this addition, although I'm going to tweak it slightly.

bvaughn:
Going out as 9.14.0

PR #892, opened 2017-12-05
add defaultHeight, defaultWidth props on WindowScroller

landvibe:
Fixed: #891 

When I run server-side rendering, browser prints below warning and fail initial rendering.
`Warning: Expected server HTML to contain a matching <div> in <div>`
I added defaultHeight and defaultWidth props like AutoSizer.
I checked There is no warning with the props.


kewats:
+1

I too am using WindowScroller and need server-side rendering to work for me. Looking forward to see this change merged.

PR #899, opened 2017-12-13
Resolves #898 - Allow Table to accept children other than Column type

### LINKED ISSUE #898: Creating a custom Column using composition instead of inheritance?
Are there any plans to support using composition of `Column` in `Table`?  Seems like `Table` is very strict about children types seen [here](https://github.com/bvaughn/react-virtualized/blob/master/source/Table/Table.js#L29-L38).  

I ask because it seems like composition is preferred over inheritance in React... I want to create custom column components and right now I find myself having to extend `Column` instead of `React.Component`.

genomics-geek:
Changed the propType validation for children in Table to accept other types, not just Column.

My reasoning:

I wanted to create some custom columns for Table and I was getting an error unless I used Column or a class that extended Column.  This allows for using composition to create customized Columns.

### DIFF
--- a/source/Table/Table.jest.js
+++ b/source/Table/Table.jest.js
@@ -95,34 +95,6 @@ describe('Table', () => {
 
   beforeEach(() => jest.resetModules());
 
-  describe('children', () => {
-    it('should accept Column children', () => {
-      const children = [<Column dataKey="foo" width={100} />];
-      const result = Table.propTypes.children({children}, 'children', 'Table');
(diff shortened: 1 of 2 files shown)


bvaughn:
Seems like you were able to resolve this using the `Column` `cellRender` prop 😄

genomics-geek:
Yup! I was going to say - I also didn't quite understand `columnData` prop either, but after digging into the react-virtualized code, it does exactly what I needed :).  Thanks again

PR #900, opened 2017-12-13
Update detectElementResize to support rendering into iframes and child windows

ahutchings:
Without these changes, the size of elements rendered in remote windows is incorrectly detected as height 0/width 0.

* Create detectElementResize CSS in the document containing the rendered AutoSizer.
* Create the resize trigger element using the document containing the rendered AutoSizer.

I'm not sure how to go about testing for this change - any suggestions are welcome.

### DIFF
--- a/source/AutoSizer/AutoSizer.js
+++ b/source/AutoSizer/AutoSizer.js
@@ -55,7 +55,12 @@ export default class AutoSizer extends React.PureComponent {
 
   componentDidMount() {
     const {nonce} = this.props;
-    if (this._autoSizer && this._autoSizer.parentNode instanceof HTMLElement) {
+    if (
+      this._autoSizer &&
+      this._autoSizer.parentNode &&
+      this._autoSizer.parentNode instanceof
+        this._autoSizer.parentNode.ownerDocument.defaultView.HTMLElement
+    ) {
       // Delay access of parentNode until mount.
       // This handles edge-cases where the component has already been unmounted before its ref has been set,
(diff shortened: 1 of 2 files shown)


bvaughn:
Can you point me at an example of the type of thing you're trying to do that this change enables? A Codesandbox demo would be great!

ahutchings:
Sure, I created a Codesandbox here: https://codesandbox.io/s/nro7yzrqv0

Two popup windows are opened -

The one with the red square is using the unpatched version of react-virtualized, so the detectElementResize CSS is not appended to its document <head> tag, and so the square does not resize.

The one with the green square is using the changes in this PR, which append the detectElementResize CSS to the owning document of the AutoSizer's parent node.

bvaughn:
Both pop-up windows appear to resize the same for me.

ahutchings:
Oops, turns out some of those changes are needed for Chrome, but not for Firefox. Working on adding them back now.

ahutchings:
Pushed a new commit and updated the Codesandbox. It now works properly in both Chrome and Firefox.

TrySound:
@bvaughn Is it possible to cover it with puppeteer tests?

bvaughn:
I dunno. Maybe?

TrySound:
@ahutchings Would you like to take care about test for this case?

ahutchings:
Sure, I will take a look at creating a test case for this. Looks like I can use the WindowScroller end-to-end test as an example.

bvaughn:
Nice! Thanks!

PR #908, opened 2017-12-19
Don't call _rowSizeAndPositionManager if targetIndex is not changed

### LINKED ISSUE #907: Jumping scroll in a List on options change and scrollToIndex defined
The easiest way to reproduce is to use `react-virtualized-select` as locally I use a List control inside InfiniteLoader based on source of react-virtualized-select. 
I had no such issue on a version 9.9 

Picture:
rv_bug

Example:
https://istarkov.github.io/rv-bug/index.html

Example source:
https://github.com/istarkov/rv-bug/blob/6437eb8f68aff3bccbaeb82cc2bd1dd07c40d5a4/src/App.js#L13-L27

Currently I avoid this issue locally setting `scrollToIndex` property of List control to undefined whenever scrolling occurs.
But its somehow tricky because key up down also a source of scroll events, also works poorly in firefox

istarkov:
Naive solution to fix #907

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -303,6 +303,7 @@ export default class Grid extends React.PureComponent<Props, State> {
   _columnStopIndex: number;
   _rowStartIndex: number;
   _rowStopIndex: number;
+  _previousTargetIndex = -1;
 
   _disablePointerEventsTimeoutId: ?AnimationTimeoutId;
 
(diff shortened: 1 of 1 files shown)


TrySound:
Would you like to write a test?

istarkov:
There is working example source in corresponding issue, do you have some kind of hand tested fixtures here? As I'm not sure I'll be able to write fully automated test for this issue.

istarkov:
PS: Also can easily write fully automatic test based on cypress.io

TrySound:
Okay. Let's try if it works with travis. I believe it may cover more cases in this project.

TrySound:
@bvaughn Are you okay with trying cypress?

bvaughn:
I have no experience with Cypress, but I'd be willing to try it.

PR #918, opened 2017-12-26
Detect non-window element resize

TrySound:
@bvaughn can this feature be minor?


source/WindowScroller/WindowScroller.js
```diff
@@ -145,10 +162,18 @@ export default class WindowScroller extends React.PureComponent<Props, State> {
...
   componentWillUnmount() {
-    if (this.props.scrollElement) {
-      unregisterScrollListener(this, this.props.scrollElement);
+    const scrollElement = this.props.scrollElement;
+    if (scrollElement) {
+      unregisterScrollListener(this, scrollElement);
+    }
+    if (scrollElement === window) {
+      window.removeEventListener('resize', this._onResize, false);
+    } else if (this._detectElementResize && scrollElement) {
```
bvaughn: This doesn't handle the case where the `scrollElement` prop changes from `window` to `!window` (or vice versa) within the life of the instance. I don't know if that's ever going to actually happen though. 😄

TrySound: Done

TrySound:
We can use now or in the future [https://github.com/que-etc/resize-observer-polyfill](resize-observer-polyfill). It's quite nice and have types. However not sure its strategy it is also efficient as current implementation.

bvaughn:
Yeah. I make mention of maybe using that in the next major release in the
umbrella notes issue. Need to think about it more before deciding for
certain.

PR #920, opened 2017-12-27
Fix position offset computing for window

TrySound:
The case is visible when you scroll down 3000px, do small resize and scroll again. Viewport is cleaned. documentElement relativity is different.

PR #928, opened 2017-12-29
Enable e2e test on ci

TrySound:
Ref #920

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/node:8.9.1-stretch-browsers
+      - image: circleci/node:latest
     steps:
       - checkout
       - restore_cache:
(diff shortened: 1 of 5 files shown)


bvaughn:
I assume this is unexpected?
screen shot 2017-12-30 at 9 24 00 am

TrySound:
Yep. Tried a lot of combinations for puppeteer config, nothing helped. My friends said travis works fine with it.

TrySound:
If disable sandbox as troubleshooting section said you will see some native lib is not found.

TrySound:
It's possibly debian problem

bvaughn:
I think I'm missing some context here then. If we merge this, it will break CI. Are you suggesting we add Travis integration for some tests? or instead of CircleCI? or...? What am I missing?

This PR references #920 but I don't see where that explains why CI is expected to fail and what the plan is to fix it?

TrySound:
I added puppeteer for e2e tests, but wasn't successful in running it on circle ci machine.

TrySound:
So disabled it temporary. Wanna fix the problem here.

TrySound:
@bvaughn It works finally!

bvaughn:
Nice!

Issue #926, opened 2017-12-29
Problematic use of babel-plugin-flow-react-proptypes in ESM build


billyjanitsch:
The use of [babel-plugin-flow-react-proptypes](https://github.com/brigand/babel-plugin-flow-react-proptypes) in the [`es` Babel env](https://github.com/bvaughn/react-virtualized/blob/master/.babelrc#L56) causes many CJS `require` calls to end up in what should be an ESM build. You can observe this in the [published build](https://unpkg.com/react-virtualized@9.14.1/dist/es/); for example, [List contains `require('../Grid')`](https://unpkg.com/react-virtualized@9.14.1/dist/es/List/List).

This largely defeats the purpose of an ESM build, since it means that most modules in the build cannot be tree-shaken under any circumstances.

The problem was introduced in #731 and there's some discussion about the need for the plugin in #730.

---

P.S. thanks for your amazing work on this library ❤️

billyjanitsch:
Unfortunately it looks like the Babel plugin cannot currently output ESM, but there is talk of potentially supporting it in brigand/babel-plugin-flow-react-proptypes#159.

I've seen other libraries solve this problem by using Flow types internally and _also_ adding explicit propTypes to user-facing components (rather than using a Babel plugin which transpiles the former into the latter). The duplication sucks, but it more explicitly captures the fact that Flow types can't always map precisely onto PropTypes (e.g. functions), and tests can generically prevent the two from getting out of sync.

TrySound:
@billyjanitsch I have the same decision. Gonna write explicit prop types maybe in next minor release.

bvaughn:
FYI it looks like the `babel-plugin-flow-react-proptypes` functionality was implemented via brigand/babel-plugin-flow-react-proptypes/pull/162 and published as 14.0.0.

TrySound:
Awesome! Accidentally unwatched that repo.

TrySound:
Fixed in 9.18.1

PR #931, opened 2017-12-30
Manual prop types

TrySound:
List of components
- [ ] ArrowKeyStepper
- [x] AutoSizer
- [ ] CellMeasurer
- [ ] Collection
- [x] ColumnSizer
- [ ] Grid
- [x] InfiniteLoader
- [x] List
- [ ] Masonry
- [x] MultiGrid
- [x] ScrollSync
- [x] Table
- [x] WindowScroller

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "jsnext:main": "dist/es/index.js",
   "license": "MIT",
   "scripts": {
-    "build:types": "flow-copy-source --ignore \"**/*.{jest,e2e,example}.js\" source/WindowScroller dist/es/WindowScroller",
+    "build:types": "flow-copy-source --ignore \"**/*.{jest,e2e,example}.js\" source/WindowScroller dist/es/WindowScroller && flow-copy-source --ignore \"**/*.{jest,e2e,example}.js\" source/AutoSizer dist/es/AutoSizer",
     "build": "npm run build:commonjs && npm run build:css && npm run build:es && npm run build:demo && npm run build:umd",
     "build:commonjs": "npm run clean:commonjs && cross-env NODE_ENV=production cross-env BABEL_ENV=commonjs babel source --out-dir dist/commonjs",
     "build:css": "postcss source/styles.css -o styles.css --use autoprefixer",
@@ -27,7 +27,7 @@
     "deploy": "gh-pages -d build",
     "lint": "eslint 'source/**/*.js'",
     "typecheck": "flow check",
-    "prebuild": "npm run typecheck && npm run lint",
+    "postbuild": "npm run typecheck && npm run lint",
     "prettier": "prettier --write '{playground,source}/**/*.js'",
     "prettier:diff": "prettier --list-different '{playground,source}/**/*.js'",
     "postpublish": "npm run deploy",
--- a/source/AutoSizer/AutoSizer.js
+++ b/source/AutoSizer/AutoSizer.js
@@ -1,13 +1,19 @@
 /** @flow */
 
-import type {Size} from './types';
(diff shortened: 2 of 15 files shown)


bvaughn:
Could you explain the motivation for this? It seems undesirable to have to maintain duplicate prop definitions for each component.

TrySound:
https://github.com/bvaughn/react-virtualized/issues/926

bvaughn:
Can we talk about this change first?

TrySound:
commonjs may not work good with esm in the same file. Also I don't like reexporting of prop types. Output code is a bit scary.

bvaughn:
The idea was to use the Flow types to auto-generate the `propTypes`. Flow types are useful for us, while building RV. `propTypes` are useful during runtime, for people using RV.

This PR seems to add redundant Flow and `propTypes` which doubles the maintenance effort. I don't understand why that is required. I'd like to avoid it if possible. Can you explain why it was necessary?

billyjanitsch:
@bvaughn were you able to look at #926? It has a good overview of why this type of change is necessary.

In short, the Babel plugin used to convert Flow types into React PropTypes outputs CJS and does not currently support outputting ESM. This causes [many `require` calls](https://unpkg.com/react-virtualized@9.14.1/dist/es/List/List) to end up in the ESM build of react-virtualized, preventing tree-shaking thereby defeating its purpose. (FWIW it looks like [work has just started](babel-plugin-flow-react-proptypes/pull/162) on supporting ESM output.)

(Another disadvantage of using the plugin is that Flow types can't always map exactly onto PropTypes (e.g. functions), whereas explicit PropTypes make it clear what will be validated for the user at runtime.)

As I mentioned in https://github.com/bvaughn/react-virtualized/issues/926#issuecomment-354404488, the solution that I've seen other libraries use is to add _both_ Flow types and PropTypes, and use a (generic) test to ensure that they stay in sync. It's unfortunate that this solution involves duplication, but I don't see an existing alternative if you want to support a useful ESM build.

bvaughn:
Whoops. Nope, I actually missed the "#926" comment entirely. Thank you for pointing out what I overlooked.

> Another disadvantage of using the plugin is that Flow types can't always map exactly onto PropTypes (e.g. functions), whereas explicit PropTypes make it clear what will be validated for the user at runtime.

I don't follow. Flow types are much more powerful than `PropTypes`. If there was loss of info for, eg a function type, then I wouldn't expect it to be any worse than if you'd defined it as a `PropTypes.func` in the first place. Can you maybe point out an example of what you're talking about?

> the solution that I've seen other libraries use is to add both Flow types and PropTypes, and use a generic test to ensure that they stay in sync. It's unfortunate that this solution involves duplication, but I don't see an existing alternative if you want to support a useful ESM build.

Agreed. This seems like a reasonable solution for the time being. I'd like to see us remove this once brigand/babel-plugin-flow-react-proptypes/pull/162 is merged though! 😄

billyjanitsch:
> If there was loss of info for, eg a function type, then I wouldn't expect it to be any worse than if you'd defined it as a PropTypes.func in the first place.

You're right, I can't imagine a case where you could do _better_ with PropTypes. I'm just pointing out that the loss of info is implicit, i.e. you might use a particular Flow type without realizing that this validation will be lost for the user. Anyway, this is an incidental point.

> This seems like a reasonable solution for the time being. I'd like to see us remove this once brigand/babel-plugin-flow-react-proptypes#162 is merged though!

Yeah, absolutely. Happy to help with either effort this week if that would be useful.

bvaughn:
That would be excellent 😁

Issue #938, opened 2018-01-02
PropTypes related warning with WindowScroller


michaeldzjap:
After updating to _react-virtualized_ v9.15.0 I get a
```
Warning: Failed prop type: Invalid prop `scrollElement` of type `Window` supplied to `WindowScroller`, expected instance of `Element`.
```
warning. My usage is of the form:
```javascript
<SomeCustomElm>
    <WindowScroller onResize={this.handleResize.bind(this)}>
        <AutoSizer disableHeight>
            <List
                ...
            />
        </AutoSizer>
    </WindowScroller>
</SomeCustomElm>
```
Any idea what might be wrong?

bogdansoare:
Also getting this warning

TrySound:
@bogdansoare @michaeldzjap Will release this fix today.

PR #939, opened 2018-01-02
Fix WindowScroller scrollElement prop types for window

TrySound:
Ref #938

### DIFF
--- a/source/WindowScroller/WindowScroller.js
+++ b/source/WindowScroller/WindowScroller.js
@@ -36,8 +36,7 @@ type Props = {
   onScroll: (params: {scrollLeft: number, scrollTop: number}) => void,
 
   /** Element to attach scroll event listeners. Defaults to window. */
-  scrollElement: ?Element,
-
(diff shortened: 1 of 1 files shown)


Hypnosphi:
For some reason, this change didn't get into actual `9.16.0` release:

```
mkdir test
cd test
yarn init --yes
yarn add react-virtualized
cat node_modules/react-virtualized/dist/es/WindowScroller/WindowScroller.js
```

Then you can see those lines:
```
/** Element to attach scroll event listeners. Defaults to window. */
  scrollElement: typeof Element === 'function' ? require('prop-types').instanceOf(Element) : require('prop-types').any,
```

Instead of the expected
```
/** Element to attach scroll event listeners. Defaults to window. */
  scrollElement: require('prop-types').oneOfType([require('prop-types').any, typeof Element === 'function' ? require('prop-types').instanceOf(Element) : require('prop-types').any]),

```

TrySound:
Yeah, merge collision. Will be fixed in 9.16.1

PR #940, opened 2018-01-02
Allow custom WindowScroller child with childRef in children function

TrySound:
`findDOMNode` is not quite flexible with complex layouts. This PR introduces childRef to improve this. Current behavior uses `findDOMNode` result until `childRef` is called
```js
<WindowScroller>
{({ childRef }) => <div ref={childRef} />}
</WindowScroller>
```
Also bit refactored WindowScroller tests with jest mocks.


bvaughn:
I have some concerns about this PR.

* The docs have not been updated.
* The `childRef` name is inconsistent with other components such as `InfiniteLoader` and `ColumnSizer`; (they all uses `registerChild`).

PR #942, opened 2018-01-03
Improved performance on Table with a lot of columns (> 150)

gannunziata:
Improved performance on Table with a lot of columns, avoiding internal use of `_extends()` on 
 `_createColumn()`.

### DIFF
--- a/source/Table/Table.js
+++ b/source/Table/Table.js
@@ -1,6 +1,6 @@
 /** @flow */
 
-import type {CellPosition} from '../Grid';
+// import type {CellPosition} from '../Grid';
 
 import cn from 'classnames';
 import Column from './Column';
@@ -8,6 +8,8 @@ import PropTypes from 'prop-types';
 import React, {PureComponent} from 'react';
 import {findDOMNode} from 'react-dom';
 import Grid, {accessibilityOverscanIndicesGetter} from '../Grid';
+import type {CellPosition} from '../Grid';
+
 import defaultRowRenderer from './defaultRowRenderer';
 import defaultHeaderRowRenderer from './defaultHeaderRowRenderer';
 import SortDirection from './SortDirection';
@@ -426,6 +428,7 @@ export default class Table extends PureComponent {
       className,
       columnData,
       dataKey,
+      title,
       id,
     } = column.props;
 
@@ -443,24 +446,20 @@ export default class Table extends PureComponent {
 
     const style = this._cachedColumnStyles[columnIndex];
 
-    const title = typeof renderedCell === 'string' ? renderedCell : null;
+    const colTitle = title ? title : (typeof renderedCell === 'string' ? renderedCell : null);
 
-    const a11yProps = {
+    const columnProps = {
       role: 'gridcell',
+      'aria-describedby': id || undefined,
+      key: "Row" + rowIndex + "-" + "Col" + columnIndex,
+      className: cn('ReactVirtualized__Table__rowColumn', className),
+      style: style,
+      title: colTitle,
     };
-
-    if (id) {
-      a11yProps['aria-describedby'] = id;
-    }
-
-    a11yProps.key = "Row" + rowIndex + "-" + "Col" + columnIndex;
-    a11yProps.className = cn('ReactVirtualized__Table__rowColumn', className);
-    a11yProps.style = style;
-    a11yProps.title = title;
     
     return (
       <div
-        {...a11yProps}
+        {...columnProps}
        >
         {renderedCell}
       </div>
@@ -482,6 +481,7 @@ export default class Table extends PureComponent {
       headerRenderer,
       id,
       label,
+      title,
       columnData,
       defaultSortDirection,
     } = column.props;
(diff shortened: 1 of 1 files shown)


source/Table/Table.js
```diff
@@ -453,13 +453,15 @@ export default class Table extends PureComponent {
       a11yProps['aria-describedby'] = id;
     }
 
+    a11yProps.key = "Row" + rowIndex + "-" + "Col" + columnIndex;
+    a11yProps.className = cn('ReactVirtualized__Table__rowColumn', className);
+    a11yProps.style = style;
+    a11yProps.title = title;
```
TrySound: Can you move all these props to object literal and rename `a11yProps` to `cellProps`?

source/Table/Table.js
```diff
@@ -553,12 +555,13 @@ export default class Table extends PureComponent {
       a11yProps.id = id;
     }
 
+    a11yProps.key = "Header-Col" + index;
+    a11yProps.className = classNames;
+    a11yProps.style = style;
+
```
TrySound: Same move and rename to headerProps

source/Table/Table.js
```diff
@@ -523,7 +525,7 @@ export default class Table extends PureComponent {
           ? SortDirection.ASC
           : SortDirection.DESC;
 
-      const onClick = event => {
+      const onClick = event => {
```
bvaughn: nit: trailing whitespace

source/Table/Table.js
```diff
@@ -453,13 +453,15 @@ export default class Table extends PureComponent {
...
+    a11yProps.title = title;
+    
     return (
       <div
         {...a11yProps}
-        key={`Row${rowIndex}-Col${columnIndex}`}
-        className={cn('ReactVirtualized__Table__rowColumn', className)}
-        style={style}
-        title={title}>
+       >
```
bvaughn: Looks like the crux of this fix is to avoid generating sub-optimal `React.createElement` boilerplate with an added function call and `for`-loop. But that's not obvious at a glance, and might regress later.

I'd suggest adding a comment about why it's important not to mix inline props and spread props.

source/Table/Table.js
```diff
@@ -1,13 +1,15 @@
...
-import type {CellPosition} from '../Grid';
+// import type {CellPosition} from '../Grid';
 
 import cn from 'classnames';
 import Column from './Column';
 import PropTypes from 'prop-types';
 import React, {PureComponent} from 'react';
 import {findDOMNode} from 'react-dom';
 import Grid, {accessibilityOverscanIndicesGetter} from '../Grid';
+import type {CellPosition} from '../Grid';
```
TrySound: Move it back please

source/Table/Table.js
```diff
@@ -538,27 +540,33 @@ export default class Table extends PureComponent {
         }
       };
 
-      a11yProps['aria-label'] = column.props['aria-label'] || label || dataKey;
-      a11yProps.tabIndex = 0;
-      a11yProps.onClick = onClick;
-      a11yProps.onKeyDown = onKeyDown;
+      ariaLabel = column.props['aria-label'] || label || dataKey;
+      chTabIndex = 0;
+      chOnClick = onClick;
```
TrySound: Can you move these values to object literal?

source/Table/Table.js
```diff
@@ -8,6 +8,7 @@ import PropTypes from 'prop-types';
 import React, {PureComponent} from 'react';
 import {findDOMNode} from 'react-dom';
 import Grid, {accessibilityOverscanIndicesGetter} from '../Grid';
+
```
TrySound: space

source/Table/Table.js
TrySound: Never mind. Didn't see condition

source/Table/Table.js
```diff
@@ -426,6 +428,7 @@ export default class Table extends PureComponent {
       className,
       columnData,
       dataKey,
+      title,
```
bvaughn: nit: Alpha sort prop-lists like this 😄

source/Table/Table.js
```diff
@@ -482,6 +481,7 @@ export default class Table extends PureComponent {
       headerRenderer,
       id,
       label,
+      title,
```
bvaughn: Another props-sort nit

source/Table/Table.js
```diff
@@ -443,23 +445,24 @@ export default class Table extends PureComponent {
...
     };
-
-    if (id) {
-      a11yProps['aria-describedby'] = id;
-    }
-
+    
+    // NOTE: is important to NOT mixin inline property and spread properties,
+    // since internally REACT clone property object to adds inline property
+    // and slow down performances
```
bvaughn: The wording of this comment (and the one below) confuse me. I don't think this has anything to do with "React internals" (unless I'm misunderstanding something). I believe this change is being made to avoid the overhead added by the [`babel-runtime` `extends` method](https://unpkg.com/babel-runtime@6.26.0/helpers/extends).

Before this change:
```js
return _react2.default.createElement(
  'div',
  (0, _extends3.default)({}, a11yProps, {
    key: 'Row' + rowIndex + '-Col' + columnIndex,
    className: (0, _classnames2.default)('ReactVirtualized__Table__rowColumn', className),
    style: style,
    title: title }),
  renderedCell
);
```

After this change:
```js
return (
  <div
    {...columnProps}
   >
    {renderedCell}
  </div>
);
```

bvaughn: That being said, I'm happy to update the wording of the comment myself, so don't worry about it. 😄

bvaughn:
Thanks for the contribution!

PR #943, opened 2018-01-04
Add class to AutoSizer's wrapping div.

manuelmager:
AutoSizer is wrapped in a `div` that has inline styles, but no className. That makes targetting it in css cumbersome.

I am creating a table with lots of columns, so being able to scroll horizontally is necessary. I am using a `WindowScroller` wrapped around `<Autosizer disableHeight>`, which as of now prevents horizontal scrolling due to the wrapping div having inline style:
```
overflow: visible;
width: 0px;
```

By targetting this div, I can change it's effective style to:
```
overflow-x: auto;
width: unset;
```

This enables horizontal scrolling.

### DIFF
--- a/source/AutoSizer/AutoSizer.js
+++ b/source/AutoSizer/AutoSizer.js
@@ -122,7 +122,7 @@ export default class AutoSizer extends React.PureComponent<Props, State> {
     */
 
     return (
-      <div ref={this._setRef} style={outerStyle}>
+      <div ref={this._setRef} style={outerStyle} className={'ReactVirtualized__AutoSizer'}>
         {children(childParams)}
       </div>
     );


TrySound:
@manuelmager Can one more div inside AutoSizer children solve your problem?
```
<AutoSizer>
  {({ width }) => <div style={{ width, overflowX: 'auto' }}></div>}
</AutoSizer>
```

manuelmager:
@TrySound Yes that works, but I try to keep styles I want to unset on the same element where they exist. To me, it feels like a smell to have a parent div set values and then undo them using a nested div. It works, so it's probably "just" opinion. What is your take on that?

On a sidenote, wrapping works as well, but seems like a workaround.
jsx:
```
<div class="AutoSizer_Wrapper">
    <AutoSizer>...</AutoSizer>
</div>
```
css:
```
.AutoSizer_Wrapper > div {
    overflow-x: auto !important;
    width: unset !important;
}
```

TrySound:
The case here is that AutoSizer element should be encapsulated. Whole point of react about this. I'm gonna to propose something like `childRef` of v10, so you could choose any element by yourself. If current solutions (or workarounds) work I'd like to not increase API.

bvaughn:
I've added an optional `className` and `style` prop to `AutoSizer` (94cd678). I don't think either is strictly necessary, but this is at least in line with the convention of other RV components as of version 9.x

This should enable you to accomplish what you're looking to do! Cheers

PR #946, opened 2018-01-05
Added support for table multi-column sorting

CzBuCHi:
Modified how `sort` method is called when used clicks on column header:

- simple click: original behavior
- shift + click:
  - on column, that is used for sorting: reverse column sort direction
  - on new column: add that column with default sort to the end of sort array
- control + click:
  - if column is used for sorting: remove column from sort array, otherwise add to the end

if sorted only by one column return single value instead of array (backward comp.)

preview: (action shows how to get from state in one row above)

|Action||||Sort definition|
|-|-|-|-|-|
||First &uarr;|Second|Third|First ASC|
|click **First**|First &darr;|Second|Third|First DESC|
|shift-click **First**|First &darr;|Second|Third|First DESC|
|control-click **First**|First|Second|Third||
|click **Second**|First|Second &uarr;|Third|Second ASC|
|shift-click **Second**|First &uarr;|Second &uarr;|Third|First ASC then Second ASC|
|control-click **Second**|First|Second &uarr;|Third|Second ASC|
||||
||First &uarr;|Second &uarr;|Third|Second ASC then First ASC|
|click **First**|First &darr;|Second|Third|First DESC|
|shift-click **First**|First &darr;|Second &uarr;|Third|Second ASC then First DESC|
|control-click **First**|First|Second &uarr;|Third|Second ASC|

### DIFF
--- a/source/Table/Table.jest.js
+++ b/source/Table/Table.jest.js
@@ -583,6 +583,185 @@ describe('Table', () => {
     });
   });
 
+  describe('multi-sorting', () => {
+    /*
+    this will write react error in console, but test will succeed:
+
+    console.error node_modules\react-dom\cjs\react-dom.development.js:9747
+    The above error occurred in the <Table> component:
+        in Table
+    
+    Consider adding an error boundary to your tree to customize error handling behavior.
+    Visit https://fb.me/react-error-boundaries to learn more about error boundaries.    
+    */ /*
+    it('should throw error for invalid sort-by and sort-direction props', () => {
+      expect(() => {
+        render(
+          getMarkup({
+            sort: () => {},
+            sortBy: ['name', 'email'],
+            sortDirection: SortDirection.ASC,
+          }),
+        );
+      }).toThrow('props sort-by and sort-direction cannot have different item count!');
+    });
+    */
+
+    it('should render the correct sort indicators by the current sort-by columns', () => {
+      const sortDirections = [SortDirection.ASC, SortDirection.DESC];
+      sortDirections.forEach(sortDirection => {
+        const rendered = findDOMNode(
+          render(
+            getMarkup({
+              sort: () => {},
+              sortBy: ['name', 'email'],
+              sortDirection: [sortDirection, sortDirection],
+            }),
+          ),
+        );
+
+        ['name', 'email'].forEach((name, index) => {
+          const nameColumn = rendered.querySelector(
+            '.ReactVirtualized__Table__headerColumn:nth-of-type(' +
+              (index + 1) +
+              ')',
+          );
+
+          expect(
+            nameColumn.querySelector(
+              '.ReactVirtualized__Table__sortableHeaderIcon',
+            ),
+          ).not.toEqual(null);
+          expect(
+            nameColumn.querySelector(
+              `.ReactVirtualized__Table__sortableHeaderIcon--${sortDirection}`,
+            ),
+          ).not.toEqual(null);
+        });
+      });
+    });
+
+    it('should call sort with the correct arguments when the current sort-by column header is clicked', () => {
+      const sortDirections = [SortDirection.ASC, SortDirection.DESC];
+      sortDirections.forEach(sortDirection => {
+        const sortCalls = [];
+        const rendered = findDOMNode(
+          render(
+            getMarkup({
+              sort: ({sortBy, sortDirection}) =>
+                sortCalls.push({sortBy, sortDirection}),
+              sortBy: ['name', 'email'],
+              sortDirection: [sortDirection, sortDirection],
+            }),
+          ),
+        );
+        const nameColumn = rendered.querySelector(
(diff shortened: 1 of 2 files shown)


bvaughn:
My initial thought: This is a lot of complexity (the behavior you're describing) to be built into `Table`. This sort of logic might be nicer if packaged as something you could optionally add into a `Table` or its `Column`s, perhaps via the `sort` function property.

I _believe_ the only thing preventing you from implementing it that way is the [`Table` does not pass the click `event` to the `sort` function](https://github.com/bvaughn/react-virtualized/blob/1fb6308859d98b61989d470f4f08096e1a26d937/source/Table/Table.js#L528-L531), like it [does with `onHeaderClick`](https://github.com/bvaughn/react-virtualized/blob/1fb6308859d98b61989d470f4f08096e1a26d937/source/Table/Table.js#L532). If we added that parameter to `sort`, would this be an acceptable alternative approach for you?

PS. I think your table has some typos in it, wrt what's being shift or control clicked. Either that or I'm just misunderstanding some of the rows. For example, how does shift-clicking Second end up adding First to the sort? And how does control-clicking Second remove First?

bvaughn:
I am not comfortable with changing the current default behavior of `Table` in the way proposed by this PR. I think an alternate implementation, which allows users to opt into the more advance sort behavior would be something I could feel comfortable about adding to the project though. If you would like to re-open this PR using an approach more like that, I'd be happy to review it.

CzBuCHi:
> I am not comfortable with changing the current default behavior of Table in the way proposed by this PR. I think an alternate implementation, which allows users to opt into the more advance sort behavior would be something I could feel comfortable about adding to the project though. If you would like to re-open this PR using an approach more like that, I'd be happy to review it.

i will create new PR (dont have permission to open this one) - #957

> I believe the only thing preventing you from implementing it that way is the Table does not pass the click event to the sort function, like it does with onHeaderClick. If we added that parameter to sort, would this be an acceptable alternative approach for you?

Thats would allow consumer to set sort by multiple columns, but table would render <SortIndicator /> only for first one...

> PS. I think your table has some typos in it, wrt what's being shift or control clicked. Either that or I'm just misunderstanding some of the rows. For example, how does shift-clicking Second end up adding First to the sort? And how does control-clicking Second remove First?

yea, these are copy-paste bugs ... fixed

bvaughn:
> i will create new PR (dont have permission to open this one) - #957

Yes, thanks! That's what I meant (a new PR) 😄 

My main goal is to not add this complexity into `Table` if it can be encapsulated in a small helper component/function. (This is a general pattern I've used in RV to try to keep things easier to maintain.)

> Thats would allow consumer to set sort by multiple columns, but table would render only for first one...

I'm not sure I understand your wording here. I _think_ you may be saying that because `Table` defines `sortBy` and `sortDirection` as string types, it wouldn't handle the fact that it might be sorted by multiple columns. To that end, I would not be opposed to changing the `prop-type` validations to make those string-or-array-of-strings.

CzBuCHi:
>My main goal is to not add this complexity into Table if it can be encapsulated in a small helpe component/function. 

Okay ... i can do that ... but look how i did it in that second PR first (all multi-sort related code is enabled by single prop)

> I'm not sure I understand your wording here. I think you may be saying that because Table defines sortBy and sortDirection as string types, it wouldn't handle the fact that it might be sorted by multiple columns. 

yea .. that was in my thought ... and my wording is wierd, because english is not my first language :(

CzBuCHi:
>My main goal is to not add this complexity into Table if it can be encapsulated in a small helper component/function.

looked tru the code and im not sure that composition is right way for this - i modified 'private' method _createHeader and added another 'private' method _handleMultiSortColumnClick to handle multi-sort.

i could create HOC which would override _createHeader  but that is just wrong - HOC becomes aware about 'private' method of Table.

Any ideas how to do this?

bvaughn:
"Composition" is kind of a broad term. [What I suggested](https://github.com/bvaughn/react-virtualized/pull/946#issuecomment-355763239) was a utility function to create a complex `sort` callback prop to pass to `Table`. This wouldn't require any special knowledge of `Table` internals?

CzBuCHi:
> This wouldn't require any special knowledge of Table internals?

technically i can write sort function, that would sort by multiple columns right now - but table wouldn't know about it and didnt show correct sort indicators ...

---

Writing this gived me an idea:

1. create new TableBase component - code copied from current Table component except how new sort order is calculated.
2. current Table becomes wrapper around TableBase with current sort order calculations.
3. create new MultiTable (TableEx?) as second wrapper around TableBase  with multi-sort order calculations.

from external perspective behavior of Table component wouldn't change ...

bvaughn:
> but table wouldn't know about it and didnt show correct sort indicators ...

Which is why [I also suggested updating the `sortBy` and `sortDirection` props to support arrays-of-strings](https://github.com/bvaughn/react-virtualized/pull/946#issuecomment-356031605).

PR #947, opened 2018-01-05
Rename childRef to registerChild and add some docs

TrySound:


docs/WindowScroller.md
```diff
@@ -53,3 +53,32 @@ ReactDOM.render(
...
+```javascript
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { List, WindowScroller } from 'react-virtualized';
+import 'react-virtualized/styles.css'; // only needs to be imported once
+
+ReactDOM.render(
+  <WindowScroller>
+    {({ height, isScrolling, registerChild, scrollTop }) => (
+      <div ref={registerChild}>
```
bvaughn: Is the wrapper div necessary? (Why?)

TrySound: It's an example of using `registerChild` Usually it's more complex layout which requires information from WindowScroller.

bvaughn: I think the docs should make that clearer then, probably.

The other `registerChild` props get attached to the `List` (or `Table`, whatever) itself. This one could too, right? So if we show otherwise, we should probably add a comment or explanation about _why_ or _when_ you might want to do this.

docs/WindowScroller.md
```diff
@@ -53,3 +53,37 @@ ReactDOM.render(
   document.getElementById('example')
 );
 ```
+
+or using `registerChild` you can specify grid container deeper in layout (by default `WindowScroller` uses `ReactDOM.findDOMNode` function)
```
TrySound: Looks good?

bvaughn: Much better! 😄  Thanks!

PR #950, opened 2018-01-06
Upgrade flow to prop-types converter to esm support

TrySound:
Ref #926 #931 

Upgraded to alpha version of that plugin. Looks pretty good.
Not sure, should we update to it before 10.

/cc @billyjanitsch

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "babel-eslint": "^8.1.2",
     "babel-jest": "^22.0.4",
     "babel-loader": "7.1.2",
-    "babel-plugin-flow-react-proptypes": "^13.0.1",
+    "babel-plugin-flow-react-proptypes": "^15.1.0",
     "babel-plugin-react-transform": "^3.0.0",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.12",
     "babel-plugin-transform-runtime": "^6.23.0",
@@ -158,7 +158,7 @@
     "classnames": "^2.2.3",
     "dom-helpers": "^2.4.0 || ^3.0.0",
     "loose-envify": "^1.3.0",
-    "prop-types": "^15.5.4"
+    "prop-types": "^15.6.0"
   },
   "peerDependencies": {
     "react": "^15.3.0 || ^16.0.0-alpha",
--- a/.babelrc
+++ b/.babelrc
@@ -12,12 +12,8 @@
       ],
       "plugins": [
         "transform-runtime",
-        "flow-react-proptypes",
-        [
-          "transform-react-remove-prop-types", {
-            "mode": "wrap"
-          }
-        ]
+        ["flow-react-proptypes", { "deadCode": true }],
+        ["transform-react-remove-prop-types", { "mode": "wrap" }]
       ],
       "presets": [
         "env",
@@ -72,11 +68,8 @@
       "plugins": [
         "transform-runtime",
         "flow-react-proptypes",
-        [
-          "transform-react-remove-prop-types", {
-            "mode": "wrap"
-          }
-        ]
+        ["flow-react-proptypes", { "deadCode": true, "useESModules": true }],
+        ["transform-react-remove-prop-types", { "mode": "wrap" }]
       ],
       "presets": [
         ["env", { "modules": false }],
--- a/source/Masonry/Masonry.js
+++ b/source/Masonry/Masonry.js
@@ -9,6 +9,26 @@ import {
 
 import type {AnimationTimeoutId} from '../utils/requestAnimationTimeout';
 
+type Props = {
+  autoHeight: boolean,
+  cellCount: number,
+  cellMeasurerCache: CellMeasurerCache,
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
     "deploy": "gh-pages -d build",
     "lint": "eslint 'source/**/*.js'",
     "typecheck": "flow check",
-    "prebuild": "npm run typecheck && npm run lint",
+    "postbuild": "npm run typecheck && npm run lint",
```
bvaughn: These should both remain pre-build, no? Generating `prop-types` from Flow types is orthogonal to our Flow-type-checking the source.

TrySound: Flow should verify flow types in dist folder

bvaughn: Flow should verify source-types. Building is slow. Waiting for an entire build to pass before finding out there was a Flow or lint error is not good developer experience.

TrySound: Flow should verify correctness of ported types in build. You may ignore some part of code being dependant on it. Types will be broken. Also flow and linting in build script is the last instance of process after ide and manual scripts running, so mostly a poka-yoke.

bvaughn: I feel very strongly that Flow and Lint would be run before build and on the contents of source. If you want to add an additional post build step that's fine too

bvaughn: *should be run (sorry on mobile)

TrySound: Types may fail if for example flow will be updated and build will never run. It's noisy blocker.

bvaughn: If Flow is outdated, build _should_ be blocked, because we can't trust our coverage.

TrySound: No, I mean we upgraded flow and fixed types in source. But we still check built types which are not yet regenerated on build. So it's better to generate latest types and then check them.

bvaughn: I think one of us is misunderstanding the other. I apologize if it's me.

Here's my stance: We should run Flow and Link against the source code _before_ building. This should happen _automatically_ in case we forgot or were in a hurry. This should happen _before_ build because build is slow and we always want to fail fast to tighten feedback loops.

It seems like your stance is: We should also run Flow against the thing we are about to publish to NPM. I don't object to that, but I don't want to do that _instead_ of the above, only _in addition to_.

Please help me understand what I'm missing.

TrySound: Your case is getting fast response about correctness.

My case is not getting response about incorrectness just because some files wasn't automatically built yet.

Working with types I got a lot of errors trying to run build just because checking is run before new types are landed. And it takes some time until I see that it happens in dist folder. So I can't do all machine work with only one command.

TrySound: Also build for me is fast enough. Most errors I get in place from my vim.

bvaughn: > My case is not getting response about incorrectness just because some files wasn't automatically built yet.

> Working with types I got a lot of errors trying to run build just because checking is run before new types are landed.

I think this is the key to our misunderstanding. The Flow types are defined in source. They aren't "built" like the prop-types are. Why would you need to build before checking Flow types?

TrySound: Because we are published flow types for WindowScroller and AutoSizer. This types are copied from source, but still should be checked that all necessary files are copied and dependencies are not ignored.

bvaughn: This (publishing Flow types that were copied) was a recent change. I'm not really open to the idea of regressing dev iteration time (by waiting on the entire _build_ step to Flow-check) because of it. If it's slowing you down, then let's talk about alternate approaches, like copying those Flow defs over _before_ build or something?

TrySound: Okay. I will just copy types before. It takes just a few milliseconds.

bvaughn: Cool beans. Thanks for taking the time to explain

TrySound: Hm.. not that easy. We run rimraf after checking.

bvaughn: I still don't really understand why it can't work like this:
1. Flow check and lint source
2. Clean (rimraf) dist dir
3. Build and/or copy over Flow types (in either order)
4. Flow check dist

I need to step away from the computer for a bit. 😄 Weekend and all. Sorry mate.

TrySound: Because flow can't be run on a subset of files like eslint AFAIK. For different subset you need another config in another folder.

bvaughn: Sorry mate. I think we've gone full circle here and are at an impasse.

I really don't understand why you are comfortable with losing potential type-safety on the source code in favor of the distributed bundle, vs having it for both of them. But, as I've said before, I feel pretty strongly that lint and Flow need to both continue to happen as a _prebuild_ step. Any new solution for _postbuild_ needs to be additive.

TrySound: It's mostly constraints of tools behavior. Never mind. this PR is just a proposal for new  version of prop-types converter. I won't change scripts behavior.

bvaughn: Okay!

PS. I think this is the most we've ever talked on an issue 😅

TrySound:
I finally made it work as we expect
https://github.com/brigand/babel-plugin-flow-react-proptypes/pull/174

PR #957, opened 2018-01-08
Multi column sort support

CzBuCHi:
original PR: #946 

- splitted simple and multi-sort into separate components, which internally render original table component.
- original table no longer gets sortBy and sortDirection via props (using internal fn for that)
- original table no longer calculate arguments for sort function (again delegated to internal fn)

### DIFF
--- a/source/Table/MultiTable.js
+++ b/source/Table/MultiTable.js
@@ -1,0 +1,367 @@
+/** @flow */
+
+import type {CellPosition} from '../Grid';
+
+import cn from 'classnames';
+import Column from './Column';
+import PropTypes from 'prop-types';
+import React, {PureComponent} from 'react';
+import {findDOMNode} from 'react-dom';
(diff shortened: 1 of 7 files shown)


CzBuCHi:
@bvaughn can u please review this version? I think it could be final one, but also welcome any suggestions whats can be improoved ... thx

bvaughn:
This has a bigger footprint than what I had in mind. I've created a counter-proposal (see PR #966) that shows the direction I tried to suggest when we last chatted about this. Please let me know your thoughts?

bvaughn:
I don't think I'm comfortable with the amount of duplication in prop-types and public methods between `SimpleTable`, `MultiTable`, and `Table`. Let's continue discussion this feature on PR #966 and see if we can find a solution with a smaller footprint that will still enable you to get multi-sort working. 😄

PR #965, opened 2018-01-11
Expose overflow prop on Grid component

miketmoore:
This is a small change, to expose the `overflow` property in the `Grid` component. It retains the existing value as the default (`hidden`), but now allows it to be overridden. 

Thanks!


OriR:
You can actually use `containerStyle` to override it.
Something like this should work
```js
<Grid containerStyle={{ overflow: 'flex' }} ...
```

TrySound:
Yep. The example above should work.

miketmoore:
Yes, `containerStyle` solves the problem! Thanks @OriR

Issue #904, opened 2017-12-18
Improved SSR on Masonry by populating PositionCache


djeeg:
# Requesting a Feature/Discussion

My original way to solve this was to have two modes of rendering inside _cellRenderer

Client-side, as per recommended
https://github.com/bvaughn/react-virtualized/blob/master/docs/Masonry.md#cellrenderer

 ```
          <CellMeasurer cache={this._cache} index={index} key={key} parent={parent}>
                    <div
                        style={{
                            ...style,
                            width: this.props.columnWidth,
                        }}
                    >
                        {tilecontent}
                    </div>
                </CellMeasurer>
```

Then server-side, a bit hacky, fix the height of all the cells, then the correct height would be calculated on the client side. Not the best solution, but acceptable.

      
  ```
       <div
                    key={key}
                    style={{
                        ...style,
                        top: Math.floor(index / this.props.columnCount) * 500,
                        left: (index % this.props.columnCount) * this.props.columnWidth,
                        position: "absolute",
                        width: this.props.columnWidth,
                    }}
                >
                    {tilecontent}
                </div>

```

Then I saw this issue https://github.com/bvaughn/react-virtualized/issues/872, which is basically describes the same thing that is happens when trying to implement Masonry SSR. All the cells ends up in the first column. 

The main issue holding back the SSR generation is that the `this._positionCache.range` is empty. Got me thinking that your suggested solution to inject `PositionCache `could help me here.

My proof of concept to try and populate positionCache server side

 - Allow injection of PositionCache into Masonry
https://github.com/bvaughn/react-virtualized/blob/9.14.1/source/Masonry/Masonry.js#L66
`_this._positionCache = props.positionCache ? props.positionCache : new _PositionCache2.default();`

- Change the component render to be

```
    render() {
        if (!this._cache) {
            this._cache = new CellMeasurerCache({
                defaultHeight: CELL_HEIGHT,
                defaultWidth: this.props.columnWidth,
                fixedWidth: true,
            });
        }
        if (!this._cellPositioner) {
            this._cellPositioner = createMasonryCellPositioner({
                cellMeasurerCache: this._cache,
                columnCount: this.props.columnCount,
                columnWidth: this.props.columnWidth,
                spacer: GUTTER_SIZE,
            });
        }
        var _positionCache = null;
        if(Utils.isserver()) {

[...]

PR #967, opened 2018-01-14
[Masonry] add positionCache property

djeeg:
Provide injectable positionCache to Masonry to assist SSR.

From https://github.com/bvaughn/react-virtualized/issues/904


bvaughn:
Hey @djeeg 😄

I think we may have misunderstood each other slightly. I'm not convinced that exposing this property alone does much to improve the SSR experience for the `Masonry` component, for reasons I mentioned in [my comment](https://github.com/bvaughn/react-virtualized/issues/904#issuecomment-357483484) on #904: 
> I guess I _could_ export the `PositionCache` component, and update `Masonry` to accept a default cache as a prop in the way you describe, but the experience feels pretty clunky to me over all.

The ["client hints" technique](https://github.com/jonathantneal/http-client-hints) you mentioned looks like it could address part of my concern about mismatch... but I guess I'm still not convinced that the `ServerRenderingCompat` approach I mentioned is any worse than hand-populating a `PositionCache`.

Could you provide a working, end-to-end example somewhere that I could look at? (It can be a git repo I have to check out and run locally.)

bvaughn:
Closing due to lack of response 😄

We can revisit this if there is a response!

PR #968, opened 2018-01-14
Fixed instanceOf HTMLElement to work across contexts

dfdeagle47:
https://github.com/bvaughn/react-virtualized/pull/900 fixed an issue with iframes/child windows.

One important change was to update the HTMLElement instance check to work across contexts in AutoSizer.js.

There's still one other HTML instance check in CellMeasurer.js which was not updated to work across contexts.

This PR fixes this.


bvaughn:
Nice catch! 😁 Verified locally. Thanks!

PR #969, opened 2018-01-15
Reduce unnecessary renders for CellMeasurer in Grid

bvaughn:
I realized this morning that when `CellMeasurer` is used, it causes `CellSizeAndPositionManager` to [always render every cell in a given column or row](https://github.com/bvaughn/react-virtualized/blob/04de423e9fe9db5f896add1aadc722c82bbc94ce/source/Grid/utils/CellSizeAndPositionManager.js#L210-L218) regardless of whether measurements have already been collected for those cells. This was not the intent of that code.

I believe I've fixed this to only render all cells initially (per the need explained via the inline comments). Local smoke testing seems good. Automated tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Changelog
 
 ##### 9.17.1
 * 🐛 `CellMeasurer` works properly in iframes and popup windows. ([dfdeagle47](https://github.com/dfdeagle47) - [#968](https://github.com/bvaughn/react-virtualized/pull/968))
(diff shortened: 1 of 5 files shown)


bvaughn:
I feel good enough about this to move forward and try it. Will roll back if necessary.

damusnet:
FWIW, I pulled the branch in our project where we use `CellMeasurer` and nothing broke.

And BTW, thanks for this awesome library :)

bvaughn:
Thanks for double checking!

PR #974, opened 2018-01-18
Documenting render props

vexii:
First off thanks for the project :)  
after having struggled with getting `WindowScroller` to work becouse i did not read the example propperly and missed the `onChildScroll` prop i decidet to write down my findings and share it back to you (and hoppefuly have other people enjoy this component)


TrySound:
Nice, thanks!

PR #994, opened 2018-01-31
Rollup

TrySound:
/cc @Andarist


.babelrc
```diff
@@ -83,7 +83,7 @@
         "transform-runtime"
       ],
       "presets": [
-        "env",
+        ["env", { "modules": false }],
```
Andarist: I'd recommend to refactor this whole file. I usually just create `.babelrc.js` for better control (you have to put `{"presets": ["./.babelrc.js"]}` in you `.babelrc` so babel6 can handle it) and then I can just use javascript to construct plugins/presets based on `process.env` variables. It's imho hard to see what changes between various variants here, because of overwhelming duplication

Andarist: Oh, I'd also recommend using `loose` mode with `modules: false` in call cases + smth like
```js
const cjs = process.env.NODE_ENV === 'test' || process.env.BABEL_ENV === 'cjs'
// ...
plugins: [
 // ...
 cjs && 'transform-es2015-modules-commonjs'
].filter(Boolean)
```

Why? `loose` mode for everything except commonjs transform, because it makes `__esModule` property enumerable, so I consider default mode better for this case.

rollup.config.js
```diff
@@ -0,0 +1,35 @@
...
+  },
+  external: ['react', 'react-dom'],
+  plugins: [
+    nodeResolve(),
+    commonjs({
+      include: 'node_modules/**',
+    }),
+    babel({
+      exclude: 'node_modules/**',
+      runtimeHelpers: true,
```
Andarist: would be good to compare actual output for this and for the variant with `plugins: ['external-helpers']` in case of umd build, I suspect that latter might be slightly better or there is no difference - im curious which one of those 2 😅

rollup.config.js
```diff
@@ -0,0 +1,35 @@
...
+    }),
+    uglify({
```
Andarist: I think you are lacking a `rollup-plugin-replace` here which would replace `process.env.NODE_ENV` in your bundle

Andarist:
Good thinking with using rollup for just umd build first 👌

bvaughn:
😮 ... 🎉 😎

TrySound:
@Andarist could you take a look?

package.json
```diff
@@ -12,11 +12,11 @@
...
     "build": "npm run build:commonjs && npm run build:css && npm run build:es && npm run build:demo && npm run build:umd",
-    "build:commonjs": "npm run clean:commonjs && cross-env NODE_ENV=production cross-env BABEL_ENV=commonjs babel source --out-dir dist/commonjs",
+    "build:commonjs": "npm run clean:commonjs && cross-env NODE_ENV=commonjs babel source --out-dir dist/commonjs",
     "build:css": "postcss source/styles.css -o styles.css --use autoprefixer",
     "build:demo": "npm run clean:demo && cross-env NODE_ENV=production webpack --config webpack.config.demo.js -p --bail",
-    "build:es": "npm run clean:es && npm run build:types && cross-env NODE_ENV=production cross-env BABEL_ENV=es babel source --out-dir dist/es",
-    "build:umd": "npm run clean:umd && cross-env NODE_ENV=production webpack --config webpack.config.umd.js --bail",
+    "build:es": "npm run clean:es && npm run build:types && cross-env NODE_ENV=es babel source --out-dir dist/es",
+    "build:umd": "npm run clean:umd && cross-env NODE_ENV=rollup rollup -c",
```
Andarist: i'd just use separate env var for this, let's keep `NODE_ENV` restricted to things like production, development, test etc

.babelrc.js
```diff
@@ -0,0 +1,81 @@
...
+  '*.jest.js',
+  '*.e2e.js',
+  '*.ssr.js',
+  '*.example.js',
+  'source/demo',
+  'source/jest-*.js',
+  'source/TestUtils.js',
+];
+
+if (env === 'commonjs') {
```
Andarist: personally I just construct a single `module.exports` based on env vars, so i just add plugins, presets (or modify some options) conditionally

that way I better see which plugins are common to all possible values and which ones differ

but whatever works for you :)

.babelrc.js
```diff
@@ -0,0 +1,81 @@
...
+      ['transform-react-remove-prop-types', {mode: 'wrap'}],
+    ],
+    presets: [['env', {modules: false}], 'react', 'flow', 'stage-1'],
+  };
+}
+
+if (env === 'rollup') {
+  module.exports = {
+    comments: false,
+    plugins: ['external-helpers'],
```
Andarist: technically this could be passed to rollup-plugin-babel and it would just merge it with what this `.babelrc.js` returns, that way you wouldn't have introduce separate env var just for rollup, but u'd somewhat loose 'centralization' of all babel configuration 

i wouldnt concern much about the latter as its kinda temporary, because with babel@7 and its corresponding version of the babel plugin you won't have to deal with external-helpers at all, this optimization will be applies automatically by the plugin

TrySound: Yeah, I'm gonna open PR with babel7 later

PR #1007, opened 2018-02-09
Fix #1004 Update MultiGrid width when cell width is changed by getter function

### LINKED ISSUE #1004: MultiGrid does not update right grid when fixed column size changes
I have a MultiGrid where the column widths can be changed by the user. When that happens on a fixed column the column itself is updated but the right grid is not updated to use the free space.

Here is an example:
https://codesandbox.io/s/jllvqo5n0w

MarcelSchmitt:
The calculations for the new leftGridWidth where already handled correctly but the MultiGrid was not updated with it.
I included a short if to only forceUpdate when the leftGridWidth really changed.

### DIFF
--- a/source/MultiGrid/MultiGrid.js
+++ b/source/MultiGrid/MultiGrid.js
@@ -139,9 +139,14 @@ export default class MultiGrid extends PureComponent {
         rowIndex,
       });
 
+    let oldLeftGridWidth = this._leftGridWidth;
     this._leftGridWidth = null;
     this._topGridHeight = null;
     this._maybeCalculateCachedStyles(null, this.props, null, this.state);
+    if (oldLeftGridWidth != this._leftGridWidth)
+    {
+      this.forceUpdate();
+    }
   }
 
   componentDidMount() {
--- a/source/MultiGrid/MultiGrid.jest.js
+++ b/source/MultiGrid/MultiGrid.jest.js
@@ -248,6 +248,44 @@ describe('MultiGrid', () => {
       expect(bottomLeftAfter.style.getPropertyValue('width')).toEqual('75px');
       expect(bottomRightAfter.style.getPropertyValue('width')).toEqual('325px');
     });
+    
+    it('should update the MultiGrid width when cell width changes by getter function', () => {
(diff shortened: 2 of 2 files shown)


source/MultiGrid/MultiGrid.js
```diff
@@ -139,9 +139,14 @@ export default class MultiGrid extends PureComponent {
         rowIndex,
       });
 
+    let oldLeftGridWidth = this._leftGridWidth;
     this._leftGridWidth = null;
     this._topGridHeight = null;
     this._maybeCalculateCachedStyles(null, this.props, null, this.state);
+    if (oldLeftGridWidth != this._leftGridWidth)
```
petersendidit: The same problem exists when there is fixed rows and their height changes. If this code checked both the `leftGridWidth` and the `topGridHeight` it would handle both correctly. 

```diff
diff --git a/source/MultiGrid/MultiGrid.js b/source/MultiGrid/MultiGrid.js
index 74675e6..c98fe15 100644
--- a/source/MultiGrid/MultiGrid.js
+++ b/source/MultiGrid/MultiGrid.js
@@ -139,12 +139,12 @@ export default class MultiGrid extends PureComponent {
         rowIndex,
       });
 
-    let oldLeftGridWidth = this._leftGridWidth;
+    const oldLeftGridWidth = this._leftGridWidth;
+    const oldTopGridHeight = this._topGridHeight;
     this._leftGridWidth = null;
     this._topGridHeight = null;
     this._maybeCalculateCachedStyles(null, this.props, null, this.state);
-    if (oldLeftGridWidth != this._leftGridWidth)
-    {
+    if (oldLeftGridWidth != this._leftGridWidth || oldTopGridHeight != this._topGridHeight) {
       this.forceUpdate();
     }
   }
```

Here is a test case that would cover the fixed row height changing: https://gist.github.com/petersendidit/a80d38510a2d752a231715061ab3eab1

wuweiweiwu: nit: use `!==` for comparing primitives

wuweiweiwu: use `!==` for comparing primitives

PR #1009, opened 2018-02-10
Replace deprecated / async-unsafe lifecycles

### LINKED ISSUE #1008: Replace deprecated / async-unsafe lifecycles
Replace `componentWillMount`, `componentWillReceiveProps`, and `componentWillUpdate` with async-safe lifecycles in advance of React 16.4 deprecation warnings. Use [react-lifecycles-compat](https://github.com/reactjs/react-lifecycles-compat) to ensure backwards compatibility.

Relates to reactjs/rfcs/pull/6

In-progress documentation available at reactjs/reactjs.org/pull/587 and reactjs/reactjs.org/pull/596

RFC at https://github.com/reactjs/rfcs/blob/master/text/0006-static-lifecycle-methods.md

bvaughn:
**Note** this PR updates all components (including examples) _except_ for `Grid`. `Grid` was turning out to be a large update, so I reverted the changes to it in 72d7fd1 and we can migrate it in a follow up commit.

For all other (non-`Grid` components) this PR includes:
* Replacing `componentWillMount`, `componentWillReceiveProps`, and `componentWillUpdate` with async-safe lifecycles in advance of React 16.4 deprecation warnings.
* Using [react-lifecycles-compat](https://github.com/reactjs/react-lifecycles-compat) where necessary to ensure backwards and forwards compatibility with the new static `getDerivedStateFromProps` lifecycle.
* Tidied up and replaced a few constructor-binded methods with arrow functions.

Relates to [this RFC](https://github.com/reactjs/rfcs/blob/master/text/0006-static-lifecycle-methods.md). In-progress reference documentation available [here](https://deploy-preview-596--reactjs.netlify.com/blog/2018/03/15/update-on-async-rendering.html).

Resolves #1008


source/WindowScroller/WindowScroller.js
```diff
@@ -138,25 +136,24 @@ export default class WindowScroller extends React.PureComponent<Props, State> {
...
+      prevProps.scrollElement != null &&
+      scrollElement != null
     ) {
-      this.updatePosition(nextScrollElement, nextProps);
+      this.updatePosition(scrollElement);
 
-      unregisterScrollListener(this, scrollElement);
-      registerScrollListener(this, nextScrollElement);
+      // $FlowFixMe Why is this necessary? There's a check right above.
+      unregisterScrollListener(this, prevProps.scrollElement);
```
TrySound: because updatePosition may change this value value.

bvaughn: Ah! Nice observation.

TrySound: before is worked ok, because it was const instead of field

source/WindowScroller/WindowScroller.js
```diff
@@ -138,25 +136,24 @@ export default class WindowScroller extends React.PureComponent<Props, State> {
...
+      registerScrollListener(this, scrollElement);
 
-      this._unregisterResizeListener(scrollElement);
-      this._registerResizeListener(nextScrollElement);
+      this._unregisterResizeListener(prevProps.scrollElement);
+      this._registerResizeListener(scrollElement);
     }
   }
-
```
TrySound: why remove newline?

bvaughn: Probably an accident. This is a WIP PR. Not quite ready for nit-review yet 😉

bvaughn: (I'll do a final cleanup pass myself once I'm done with the major refactor)

bvaughn:
Just a quick update- the bulk of the migration is done, but `Grid` triggers a warning due to a `setState` call during `render`. I'm in the process of fixing this locally, but it has a bigger footprint than I wish because it involves a lot of instance properties/methods.

bvaughn:
At this point, I find myself wanting to make everything in `Grid` static, except for the lifecycles that are required to be instance methods- but even those, I think should just defer to static methods.

My initial attempt to only update things that need access to cached values on the instance feels like it's getting out of control. So many of the helper methods call each other.

It's unfortunate that there's so much complexity within the timing and sequence of things `Grid` does. I hope the current test coverage is sufficient to cover my ass with these changes.

bvaughn:
Made some progress today but got distracted by another alpha release and some other work things. Will resume tomorrow!

wuweiweiwu:
@bvaughn I can help migrate to `getDerivedStateFromProps` :)

bvaughn:
I think the only component left to migrate is `Grid` and it's kind of in a halfway done state.

bvaughn:
Since updates to `Grid` are a lot more involved than the other components, I've decided to revert the changes in this branch and just focus on landing the others. `Grid` can be updated (by me, or someone else) in a follow up PR. This way I'm not holding up so many things.

bvaughn:
This PR is now ready for review and testing!

source/Masonry/Masonry.js
```diff
@@ -328,11 +332,11 @@ export default class Masonry extends React.PureComponent<Props> {
     );
   }
 
-  _debounceResetIsScrollingCallback() {
+  _debounceResetIsScrollingCallback = () => {
```
wuweiweiwu: nit: some functions are arrows and some are not.

source/Collection/CollectionView.js
```diff
@@ -177,6 +178,41 @@ export default class CollectionView extends React.PureComponent {
...
+        scrollLeft: 0,
+        scrollTop: 0,
+      };
+    } else if (
+      nextProps.scrollLeft !== prevState.scrollLeft ||
+      nextProps.scrollTop !== prevState.scrollTop
+    ) {
+      return {
+        scrollLeft:
+          nextProps.scrollLeft != null
```
wuweiweiwu: nit: `!==` unless `nextProps` only contains updated props? since `undefined == null`

source/CellMeasurer/CellMeasurer.DynamicHeightTableColumn.example.js
```diff
@@ -88,11 +82,11 @@ export default class DynamicHeightTableColumn extends React.PureComponent {
         </div>
       </CellMeasurer>
     );
-  }
+  };
```
wuweiweiwu: nit: semi colon inconsistencies

source/WindowScroller/WindowScroller.js
```diff
@@ -138,22 +135,22 @@ export default class WindowScroller extends React.PureComponent<Props, State> {
     this._isMounted = true;
   }
 
-  componentWillReceiveProps(nextProps: Props) {
-    const scrollElement = this.props.scrollElement;
-    const nextScrollElement = nextProps.scrollElement;
+  componentDidUpdate(prevProps: Props, prevState: State) {
```
wuweiweiwu: do we need to include `prevState` if we're not using it?

source/CellMeasurer/CellMeasurer.DynamicHeightTableColumn.example.js
bvaughn: Prettier formats everything 🙂

source/Collection/CollectionView.js
bvaughn: This check handles null and undefined. It's probably not necessary but I'd prefer to be safe

source/Masonry/Masonry.js
bvaughn: It's actually intentional. I think I might have a lint rule to protect against unnecessary arrow functions. Not sure though. But I added arrow functions where I wanted to ensure the correct "this" even after passing references around

source/WindowScroller/WindowScroller.js
bvaughn: I guess we don't _need_ to but it's part of the lifecycles signature so it's just habit to add it. 😁

bvaughn:
Thanks for the review @wuweiweiwu! I added a couple of responses inline. They're a bit terse because I'm on my mobile

source/Masonry/Masonry.js
wuweiweiwu: Oh I see!

source/Collection/CollectionView.js
wuweiweiwu: Ok that makes sense.

source/MultiGrid/MultiGrid.js
```diff
@@ -644,7 +632,6 @@ export default class MultiGrid extends React.PureComponent {
         ref={this._bottomLeftGridRef}
         rowCount={Math.max(0, rowCount - fixedRowCount) + additionalRowCount}
         rowHeight={this._rowHeightBottomGrid}
-        scrollTop={scrollTop}
```
TrySound: Why this?

TrySound:
I like how much cleaner became code in a few places.

source/ArrowKeyStepper/ArrowKeyStepper.js
```diff
@@ -40,45 +40,35 @@ export default class ArrowKeyStepper extends React.PureComponent<
...
-        scrollToRow,
-      });
-    } else if (prevScrollToColumn !== scrollToColumn) {
-      this.setState({scrollToColumn});
-    } else if (prevScrollToRow !== scrollToRow) {
-      this.setState({scrollToRow});
+      return {
+        scrollToColumn: nextProps.scrollToColumn,
+        scrollToRow: nextProps.scrollToRow,
+      };
```
timhwang21: This is more a comment on the new React API than anything, but I have to admit I found this hard to read, especially because `_setScrollPosition()` ends up calling `this.setState()`, `_updateScrollPositionForScrollToCell()` ends up maybe calling `_setScrollPosition()`, and this lifecycle ends up returning an object that is "magically" set to state. All 3 methods handle somewhat overlapping responsibilities yet they all do so in a different way.

(The added ending `return null` in `_setScrollPosition()` also threw me for a second, because I associated returning with `calculateDerivedStateFromProps`. So that method either ends with `setState` (which returns nothing) or returning nothing.)

I'm not sure what the best way to make this clearer would be (or if more experience would be enough). Maybe type signatures? I don't really use flow -- does it have a way to signify "this method will have a side effect via setting state" or something?

source/MultiGrid/MultiGrid.js
```diff
@@ -549,22 +503,57 @@ export default class MultiGrid extends React.PureComponent {
...
+    this._lastRenderedStyle = style;
+    this._lastRenderedStyleBottomLeftGrid = styleBottomLeftGrid;
+    this._lastRenderedStyleBottomRightGrid = styleBottomRightGrid;
+    this._lastRenderedStyleTopLeftGrid = styleTopLeftGrid;
+    this._lastRenderedStyleTopRightGrid = styleTopRightGrid;
+    this._lastRenderedWidth = width;
   }
 
-  _onScroll(scrollInfo) {
+  _prepareForRender() {
```
timhwang21: FWIW I'm a fan of this pattern & think it's much more readable than the large render functions that something like `CollectionView` has.

bvaughn:
Oh, shoot. I'm sorry @timhwang21 and @TrySound for merging before addressing your comments. They didn't show up for me until _after_ the merge. ☹️

source/MultiGrid/MultiGrid.js
bvaughn: Looks like it may have been an accident. Good catch.

bvaughn: Oh, no never mind. I just removed the explicit `scrollTop={scrollTop}` because it wasn't necessary. I'm already spreading `{...props}` above. 😄

bvaughn: The other props are explicitly passed because they have different names.

source/ArrowKeyStepper/ArrowKeyStepper.js
```diff
@@ -28,10 +29,9 @@ type Props = {
   scrollToRow: number,
 };
 
-export default class ArrowKeyStepper extends React.PureComponent<
-  Props,
-  ScrollIndices,
-> {
+type State = ScrollIndices;
+
+class ArrowKeyStepper extends React.PureComponent<Props, State> {
```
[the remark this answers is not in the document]
bvaughn: Yup.

source/Collection/CollectionView.js
```diff
@@ -26,7 +27,7 @@ const SCROLL_POSITION_CHANGE_REASONS = {
  * Monitors changes in properties (eg. cellCount) and state (eg. scroll offsets) to determine when rendering needs to occur.
  * This component does not render any visible content itself; it defers to the specified :cellLayoutManager.
  */
-export default class CollectionView extends React.PureComponent {
+class CollectionView extends React.PureComponent {
```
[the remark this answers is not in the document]
bvaughn: Some components haven't been migrated to Flow types yet. Ideally, all would have types.

source/Grid/types.js
```diff
@@ -17,8 +17,11 @@ export type CellRendererParams = {
 
 export type CellRenderer = (props: CellRendererParams) => React.Element<*>;
 
+export type CellCache = {[key: string]: React.Element<*>};
+export type StyleCache = {[key: string]: Object};
+
 export type CellRangeRendererParams = {
-  cellCache: Object,
+  cellCache: CellCache,
```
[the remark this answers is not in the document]
bvaughn: No objection really 😄 Focused PRs are generally good. I just did some tidying up as I went along.

source/ArrowKeyStepper/ArrowKeyStepper.js
bvaughn: I think this comment is more pertaining to `CollectionView` than `ArrowKeyStepper`?

I think the null return from `_setScrollPosition` was just an oversight on my part. 😄

timhwang21: Ah, yeah -- comment should be there instead :)

dobryanskyy:
Guys, when is this going to be merged into some npm version?

TrySound:
@dobryanskyy Probably in a week.

TrySound:
@dobryanskyy However it's also depends on https://github.com/bvaughn/react-virtualized/pull/1053, so not sure.

/cc @wuweiweiwu

dobryanskyy:
I see, thanks. You're doing a great product

wuweiweiwu:
@TrySound @dobryanskyy ASAP as soon as the tests pass :)

PR #1013, opened 2018-02-14
Fix #986: call requestAnimationFrame on `window`

### LINKED ISSUE #986: Error: "Uncaught TypeError: Illegal invocation"
When scrolling in a `List` component, do I get the exception `Uncaught TypeError: Illegal invocation` in Chrome. The error is thrown at https://github.com/bvaughn/react-virtualized/blob/9.18.0/source/utils/requestAnimationTimeout.js#L33.

My setup:
```
  "dependencies": {
    "babel-core": "^6.25.0",
    "babel-loader": "7.x",
    "babel-plugin-add-module-exports": "^0.2.1",
    "babel-plugin-syntax-dynamic-import": "^6.18.0",
    "babel-plugin-transform-class-properties": "^6.24.1",
    "babel-polyfill": "^6.23.0",
    "babel-preset-env": "^1.5.2",
    "babel-preset-react": "^6.24.1",
    "babel-preset-stage-0": "^6.22.0",
    "react": "^15.6.1",
    "react-dom": "^15.6.1",
    "webpack": "^3.0.0",
    "webpack-dev-server": "^2.5.0"
    ...
  }
```

This issue seems to be related: https://github.com/fritz-c/react-sortable-tree/issues/230

bmakuh:
Currently `requestAnimationFrame`/`cancelAnimationFrame` are getting called in another context besides `window` (or `self`). This throws an error as an illegal operation, so this just calls the function with the explicit context of whatever `win` is.

I'm happy to write tests for this, but I just don't really know _how_ to test this sort of thing. If anyone has ideas, let me know and I'll add them to this PR.

### DIFF
--- a/source/utils/animationFrame.js
+++ b/source/utils/animationFrame.js
@@ -5,7 +5,7 @@ type CancelAnimationFrame = (requestId: number) => void;
 type RequestAnimationFrame = (callback: Callback) => number;
 
 // Properly handle server-side rendering.
-export let win;
+let win;
 if (typeof window !== 'undefined') {
   win = window;
 } else if (typeof self !== 'undefined') {
@@ -36,5 +36,5 @@ const cancel =
     (win: any).clearTimeout(id);
   };
 
-export const raf: RequestAnimationFrame = (request: any);
-export const caf: CancelAnimationFrame = (cancel: any);
+export const raf: RequestAnimationFrame = (request: any).bind(win);
+export const caf: CancelAnimationFrame = (cancel: any).bind(win);
--- a/source/utils/animationFrame.js
+++ b/source/utils/animationFrame.js
@@ -5,7 +5,7 @@ type CancelAnimationFrame = (requestId: number) => void;
 type RequestAnimationFrame = (callback: Callback) => number;
 
 // Properly handle server-side rendering.
-let win;
+export let win;
 if (typeof window !== 'undefined') {
   win = window;
 } else if (typeof self !== 'undefined') {
--- a/source/utils/requestAnimationTimeout.js
+++ b/source/utils/requestAnimationTimeout.js
@@ -1,13 +1,13 @@
 /** @flow */
 
-import {caf, raf} from './animationFrame';
+import {caf, raf, win} from './animationFrame';
 
 export type AnimationTimeoutId = {
   id: number,
 };
 
 export const cancelAnimationTimeout = (frame: AnimationTimeoutId) =>
-  caf(frame.id);
+  caf.call(win, frame.id);
 
 /**
  * Recursively calls requestAnimationFrame until a specified delay has been met or exceeded.
@@ -25,12 +25,12 @@ export const requestAnimationTimeout = (
     if (Date.now() - start >= delay) {
       callback.call();
     } else {
-      frame.id = raf(timeout);
+      frame.id = raf.call(win, timeout);
     }
   };
 
   const frame: AnimationTimeoutId = {
-    id: raf(timeout),
+    id: raf.call(win, timeout),
   };
 
   return frame;
--- a/source/utils/requestAnimationTimeout.js
+++ b/source/utils/requestAnimationTimeout.js
@@ -1,13 +1,13 @@
 /** @flow */
 
-import {caf, raf, win} from './animationFrame';
+import {caf, raf} from './animationFrame';
 
 export type AnimationTimeoutId = {
   id: number,
 };
 
 export const cancelAnimationTimeout = (frame: AnimationTimeoutId) =>
-  caf.call(win, frame.id);
+  caf(frame.id);
 
 /**
  * Recursively calls requestAnimationFrame until a specified delay has been met or exceeded.
@@ -25,12 +25,12 @@ export const requestAnimationTimeout = (
     if (Date.now() - start >= delay) {
       callback.call();
     } else {
-      frame.id = raf.call(win, timeout);
+      frame.id = raf(timeout);
     }
   };
 
   const frame: AnimationTimeoutId = {
-    id: raf.call(win, timeout),
+    id: raf(timeout),
   };
 
   return frame;


TrySound:
@bmakuh What context are you talking about? We have two browser and

bmakuh:
@TrySound in `animationFrame.js`, `win` can be one of `window`, `self`, or `{}`. If we only do `raf.call(window, timeout)`, then that would break in Node.

TrySound:
Still not clear. What is the problem are you trying to achieve? raf is called only on scroll. Scroll does not exist in node.

bmakuh:
As referenced in the title, it fixes Issue #986, where browsers throw hundreds of exceptions while you scroll because `requestAnimationFrame` has been copied _off_ of `window` and so the value of `this` is now illegal. You get this error in Firefox:
```
TypeError: 'cancelAnimationFrame' called on an object that does not implement interface Window.
```

and this in Chrome:
```
Uncaught TypeError: Illegal invocation
```

If you want more context on the bug, there's more discussion on the issue:
https://github.com/bvaughn/react-virtualized/issues/986

vlebedeff:
@TrySound 
> What context are you talking about? We have two browser and node. Node shouldn't call raf.
`requestAnimationFrame` and `cancelAnimationFrame` are assigned to `raf` and `caf` in a point-free style. Thus when you call `raf` or `caf` they're going to have a different `this` compared to the case when you explicitly call `window.requestAnimationFrame(...)` or `window.cancelAnimationFrame(...)`. That said, IMHO `raf` and `caf` should either reference `window.requestAnimationFrame.bind(window)` / `window.cancelAnimationFrame.bind(window)` or explicitly told that `window` should be their `this` with `call`.

TrySound:
Can you reproduce the problem in [codesandbox](https://codesandbox.io/)?

source/utils/animationFrame.js
```diff
@@ -5,7 +5,7 @@ type CancelAnimationFrame = (requestId: number) => void;
 type RequestAnimationFrame = (callback: Callback) => number;
 
 // Properly handle server-side rendering.
-let win;
+export let win;
```
steppefox: Instead of exporting win object and calling it with win, why you just don't want to export raf/caf with bound context `win`?

```
export const raf = request.bind(win);
export const caf = cancel.bind(win);
```

steppefox:
Btw, maybe I'm wrong but looks like the problem appears only in ES build, commonjs build works fine

source/utils/animationFrame.js
bmakuh: Good call @steppefox, I can change that. I do think that it happens in the commonjs build too though; we're using webpack 1 currently and support for ES modules wasn't added until webpack 2.

steppefox: @bmakuh Btw, please try to provide some example in demo repository if you have a time of course, because it really looks strange and guys won't approve it then. Or i'll do than on  weekends if you want, cuz I already have some demo app, need to change import of react-virtualized to ES here: https://github.com/steppefox/sandbox-react-contactlist

steppefox:
@TrySound https://github.com/steppefox/sandbox-react-contactlist here is an example repo with existing bug, please check it out) Just clone and `npm i && npm start`

bmakuh:
@steppefox I just updated it to use `.bind()` instead, FYI.

bvaughn:
The repro project seemingly has unrelated errors for me when run locally:
> Uncaught TypeError: classnames_1.default is not a function
>  at List.render (app.js:25947)

steppefox:
@bvaughn yep, fusebox sometimes blowing up something... let me publish it to GH Pages, i think that's will be better

Nauzer:
I'm having the same issue when using ES build. @bmakuh 's fix seems to work for me.

jokklan:
This fix, solved the problem for me in #986! Thank you :)!

bmakuh:
@jokklan if you need to use this before it gets merged in, I've published a fork with the fix in it: `"@ben.makuh/react-virtualized": "9.18.6-beta.3"`

jokklan:
Thanks @bmakuh ! Awesome :)!

uqmessias:
@bmakuh I'm facing a similar issue, but even your patch didn't solve my issue.

For some weird reason, the `dist/es/utils/requestAnimationTimeout.js` file is unable to be exported by Webpack on the project that I'm working on.

I renamed it to `dist/es/utils/bla.js` and it worked well.

Do you know what could cause this weird issue?

bmakuh:
@uqmessias hmm, I'm not sure, but that maybe sounds like a separate issue?

uqmessias:
@bmakuh yesterday, right after send you a message here, I submitted a PR (#1134 ) with some changes that worked for me.
Now I'm trying to extract some files from the project I'm working on in order to post it somewhere and show the issue happening on a real project.

Anyway, thanks for answering!

domenkozar:
I can reproduce this on Chromium Version 94.0.4606.61 (Official Build) (64-bit)

This PR fixes the issue.

To reproduce follow https://github.com/domenkozar/parcel-web-components-css-modules-issue

felipernb:
I'm observing the same issue when using `react-virtualized` via `react-lazylog` and building with `parcel`. This fix seems to solve it

felipernb:
I know this might be seen as a bundling issue, as @steppefox mentioned years ago
> Btw, maybe I'm wrong but looks like the problem appears only in ES build, commonjs build works fine

but what would be the downside of just merging this? It seems to me that it just makes the module more robust.

felipernb (approved):
I know this might be seen as a bundling issue, as @steppefox mentioned years ago

 >   Btw, maybe I'm wrong but looks like the problem appears only in ES build, commonjs build works fine

but what would be the downside of just merging this? It seems to me that it just makes the module more robust.

yuxiang-projz:
reproduce when building with `parcel`

silphir:
I had same issue when I build with `parcel v2`

PR #1019, opened 2018-02-25
Refactor babel config

TrySound:
Before #994 I want to tweak babel config a bit

- remove `babel-plugin-react-transform` which seems useless right now. redbox doesn't work and catching errors is not necessary with error boundaries.
- replaced `stage-1` with `stage-2` preset
- refactored .babelrc with js preset. It's now two times smaller.

@bvaughn how `comments: false` in babel config affect code? I saw it reduces somehow umd size about 50kb


bvaughn:
> @bvaughn how comments: false in babel config affect code? I saw it reduces somehow umd size about 50kb

I don't understand what you're asking.

TrySound:
Never mind, just confused a bit.

bvaughn:
Okay 😄 

I don't have any problems with this change, although I haven't had the time to do a before and after comparison between this and what's in `master` to make sure it doesn't cause any problems in the build.

If you've done that, then feel free to merge.

PR #1020, opened 2018-02-25
Prefer react namespace over default export

TrySound:

### DIFF
--- a/source/ArrowKeyStepper/ArrowKeyStepper.example.js
+++ b/source/ArrowKeyStepper/ArrowKeyStepper.example.js
@@ -1,6 +1,6 @@
 /** @flow */
 
-import React from 'react';
+import * as React from 'react';
 import {
   ContentBox,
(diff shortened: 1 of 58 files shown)


bvaughn:
Why?

TrySound:
https://github.com/bvaughn/react-virtualized/pull/1021#issuecomment-369129661

bvaughn:
Would it be sufficient to change this:
```js

class AutoSizerExample extends PureComponent<{}, State> {
```
To this:
```js
import React from 'react';

class AutoSizerExample extends React.PureComponent<{}, State> {
```
Just curious.

TrySound:
Okay. I just think `import * as React` is more future friendly.

bvaughn:
I'm not really opposed to the "import * as" syntax. I was just curious.

If you think it's the right move in our source code, I don't strongly object.

PR #1021, opened 2018-02-25
Force to prefer react namespace via docs

TrySound:


bvaughn:
Why?

TrySound:
@bvaughn Because rollup can interop cjs only as default exports or only as named exports. This is very bad example, because two interops are used in the same time. And I hope react will land esm for v17 with named exports.
```js
import React, { Component } from 'react';
```

bvaughn:
I don't think our docs need to make Rollup-specific considerations. The [official React docs](https://reactjs.org/tutorial/tutorial.html) show:
```js
import React from 'react';
import ReactDOM from 'react-dom';
```
So I think it's reasonable for RV docs to do the same.

PR #1031, opened 2018-03-02
Fix #1015 Set scrollLeft and scrollTop in state object to match the actual values

### LINKED ISSUE #1015: When browser is zoomed or on high-DPI screens, onRowsRendered fires twice with an incorrect startIndex when scrolling via scrollToIndex
This [Code Sandbox project](https://codesandbox.io/s/2wlnn21ypp?expanddevtools=1&module=%2FExample.js) demonstrates the issue.

I've been working on a Chrome extension that uses a `<List>` component to render a few hundred items.  The user can press <kbd>Pg Up</kbd> or <kbd>Pg Down</kbd> to move the selection a screen at a time, which scrolls the next selected row into view (if needed) by setting the List's `scrollToIndex` prop.

After developing it on a standard DPI laptop and monitor, I recently tested it on a Windows 10 laptop with a 4K screen, where the UI is scaled to 250% in the display settings.  I noticed that when paging up the list after paging down, it scrolls one row too far, which never happened on the standard DPI screen.

In the code example above, the list is exactly the height of 5 rows.  Open the console to see the indices returned by `onRowsRendered`.

If you scroll the list so that an odd-numbered item is partially visible at the bottom of the list, such as *Item 7*, then `onRowsRendered` will report 2 as the `startIndex` and 7 as the `stopIndex`, since 4 whole rows and 2 partial rows are visible at this point.  If you then click the *Down* button, *Item 7* will scroll fully into view and become selected, and *Item 3* will be visible at the top of the list.

If you do this on a normal resolution screen, `onRowsRendered` will correctly report the indices as 3 and 7 after you click *Down*.  But if you do the same thing on a screen scaled at something other than a whole multiple of 100%, then `onRowsRendered` fires twice, first with 3 and 7 and then again with 2 and 7.  The latter event is wrong, since after scrolling *Item 7* fully into view, *Item 2* is no longer visible.  If you then click the *Up* button, it scrolls too far and shows *Item 2*, since the component is relying on `onRowsRendered` to track which rows are currently visible.

I've also seen the same behavior on a 4K desktop monitor with Windows 10 scaled to 175%.  Changing that to 200% seems to make the extra, incorrect `onRowsRendered` event go away.  I haven't had a chance to try it on a Mac with a Retina screen.

This issue seems like it may be due to a rounding error somewhere, though I don't know why it would cause `onRowsRendered` to fire twice.  For instance, if you change the `RowHeight` to 44 instead of 45, it doesn't happen, nor if you change the number of visible rows to 4 instead of 5 (4 * 45 = 180 vs. 5 * 45 = 225).

Test environment:

- Windows 10
- Chrome 64
- React 15.6.2 in the extension, 16.0.0 in the Code Sandbox
- React Virtualized 9.9.0 and 9.18.5

prachp:
I'm trying to fix https://github.com/bvaughn/react-virtualized/issues/1015.
Since it requires browser zoom or high-DPI screen to reproduce, I'm not sure how to create a test.

More detail about the issue [here](https://github.com/bvaughn/react-virtualized/issues/1015#issuecomment-369495037)

This PR will make the scrollLeft and scrollTop in the state have the same values as the actual scrollLeft and scrollTop.

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -618,6 +618,7 @@ export default class Grid extends React.PureComponent<Props, State> {
           columnOrRowCountJustIncreasedFromZero)
       ) {
         this._scrollingContainer.scrollLeft = scrollLeft;
+        this.state.scrollLeft = this._scrollingContainer.scrollLeft;
       }
       if (
         !autoHeight &&
@@ -627,6 +628,7 @@ export default class Grid extends React.PureComponent<Props, State> {
           columnOrRowCountJustIncreasedFromZero)
       ) {
         this._scrollingContainer.scrollTop = scrollTop;
+        this.state.scrollTop = this._scrollingContainer.scrollTop;
       }
     }
 


TrySound:
I wouldn't recommend you to mutate state. This may lead to more problems than we have right now.

prachp:
@TrySound Thank you for the comment. What approach do you think is the best way to fix this issue?

fwextensions:
@prachp FWIW, I tried monkeypatching your changes into the umd dist of React Virtualized that I'm using and they did not fix it.  I'm still seeing `onRowsRendered` fire twice, the second time with the wrong `startIndex` value, when running a Chrome extension on a high-DPI screen.

I haven't tested the zoomed browser case with those changes yet, however.

prachp:
Thank you @fwextensions for testing on high-DPI screen. I don't have high-DPI screen so I tested with browser zoom only. Close this PR since it doesn't fix the issue.

PR #1034, opened 2018-03-03
docs/multiColumnSortTable.md: fix function references.

paulkoegel-wessels:
+ fix function references so they match the initial imported function name (`createMultiSort` -> `createTableMultiSort`).

For exported function name from source/index.js, see:
https://github.com/bvaughn/react-virtualized/blob/de107e81f6d412d3561b169f7fdbfdbe25dd7b00/source/index.js#L21

### DIFF
--- a/docs/Column.md
+++ b/docs/Column.md
@@ -38,7 +38,7 @@ function ({
 }): any
 ```
 
-A [default `cellDataGetter`](https://github.com/bvaughn/react-virtualized/blob/master/source/Table/defaultCellDataGetter.js) is provided that simply returns the attribute as a String.
+A [default `cellDataGetter`](https://github.com/bvaughn/react-virtualized/blob/master/source/Table/defaultTableCellDataGetter.js) is provided that simply returns the attribute as a String.
 This function expects to operate on either a vanilla Object or a Map-like object with a get method.
 You should override this default method if your data is calculated or requires any custom processing.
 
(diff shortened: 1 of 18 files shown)


TrySound:
@bvaughn Why did you introduced this inconsistency? `source/Table` and `source/index` should have the same exports since `source/index` currently is not recommended as a good way to import components.

bvaughn:
On my mobile so can't really look.

Assume what you're asking about is why I sometimes rename and export with "table" in the name?

This is an old pattern. Doesn't really work well with deep imports and we should probably change it in version 10.

bvaughn:
(change it so both versions have the "table" bit at part of the name)

TrySound:
@paulkoegel-wessels Could you tweak also source code? Also we still need current version as an alias in both `source/index` and `source/Table` to not introduce breaking changes. You can add `TODO` to remove them later.

bvaughn:
Yeah that seems reasonable.

paulkoegel-wessels:
@TrySound Sure, I'll take a look. Should I add the `TODO` as a code comment or is there some other place where they are kept in this project?

TrySound:
I think it's only in `source/index` and `source/Table/index`. Tweak both please.

bvaughn:
An inline TODO comment is good. You can also add a comment on the version
10 bucket issue in GitHub.

paulkoegel-wessels:
Just updated with the following:
+ in `source/Table/index.js`:
  - renamed the exported function names (from, e.g., `createMultiSort` to `createTableMultiSort`)
  - added aliases for the old function names (`createMultiSort` etc.) for compatibility, added a TODO so they can be removed in version 10
  - renamed the **imported** function names as well - for consistency - their file names (hope this is okay)
+ in `source/index.js`:
  - we no longer need aliases here, only export the new names (`createTableMultiSort` etc.). Didn't add any fallbacks with old names (`createMultiSort` etc.) as there were no such exports before.

Let me know if this works for you.

paulkoegel-wessels:
One thing I wasn't sure about:
+ perhaps we should also change type names from `CellDataGetterParams` to `TableCellDataGetterParams`, e.g. here:
https://github.com/paulkoegel-wessels/react-virtualized/blob/9f34739186c679bdeb961ce63cfa601ac2a646dc/source/Table/defaultTableCellDataGetter.js#L2

PR #1038, opened 2018-03-05
Add ScrollTo props to fixed grid elements in MultiGrid

MarkBarbieri:

---

Hi,  I'm using MultiGrid wrapped in an ArrowKeyStepper element (in 'cells' mode).  This is working well but currently only the bottom right grid receives scrollToColumn & scrollToRow props from the parent MultiGrid.  Therefore, if navigating with ArrowKeyStepper in the TopLeft/BottomLeft/TopRight grids they do not receive render events and it is not possible to manage focus.

I've run npm test against this and it's all good.  Thanks

### DIFF
--- a/source/MultiGrid/MultiGrid.js
+++ b/source/MultiGrid/MultiGrid.js
@@ -279,17 +279,25 @@ export default class MultiGrid extends React.PureComponent {
     return (
       <div style={this._containerOuterStyle}>
         <div style={this._containerTopStyle}>
-          {this._renderTopLeftGrid(rest)}
+          {this._renderTopLeftGrid({
+            ...rest,
+            scrollToColumn,
+            scrollToRow,
+          })}
           {this._renderTopRightGrid({
             ...rest,
             onScroll,
             scrollLeft,
+            scrollToColumn,
+            scrollToRow,
           })}
         </div>
         <div style={this._containerBottomStyle}>
           {this._renderBottomLeftGrid({
             ...rest,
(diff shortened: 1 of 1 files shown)


bvaughn:
Can you add a test (that fails without this change and passes with it)?

TrySound:
@bvaughn probably we shouldn't add features which will be removed or deprecated in major release?

MarkBarbieri:
@bvaughn, I've never written tests before but I can try.  As the grids in question don't have the scrollTo props at the moment I assume that creating a failing test should be easy enough.  

@TrySound, based on your comments, should I be taking a different path here before I consider utilising this functionality in our project?  Is there a change pending that I may not be aware of?

TrySound:
@MarkBarbieri We gonna use props as default values and manipulate scroll with component methods.

MarkBarbieri:
OK.  if you'd prefer not to merge this then in the short term I can probably try and manage this issue via cellRangeRenderer or through forceUpdate() if scrollToColumn/scrollToRow indexes fall into the fixed columns/rows and force rendering and focus.

Sounds like this might be a good use case to test scrolling with component methods, so I'd be more than happy to help test it in this project when you have a working model.

Just to make sure we're on the same page, this is for tracking the row/column of the focused cell when MultiGrid is wrapped in an ArrowKeyStepper component.

okonet:
I'm not sure why it was closed. Is this something that will be fixed in the upstream?

PR #1039, opened 2018-03-05
docs/List.md: add info about default value of scrollToIndex

taneba:
Thanks for this great project!

I added the defalut value of scrollToIndex in List doc (now it is only in the source code).
I hope this would be helpfull.


bvaughn:
Hi @taneba! I appreciate docs PRs, so thank you for taking the time to make one!

I'm not sure if this particular one adds much value for `List` though, since the default value is kind of an internal detail. `List` checks for < 0 to determine that no `scrollToIndex` value has been set- but it could check for `null` or something else too without impacting the external API.

Maybe I'm overlooking something though?

taneba:
thank you for feedback!

> the default value is kind of an internal detail. 
it makes sense. Certainly, it's not a basic usage to be concerned about default value.
If there comes similar voice again, it'd better to consider again.

If nothing else, plz close this!

wuweiweiwu:
Closing this. Thanks for the PR @taneba hope there are more to come :)

PR #1040, opened 2018-03-07
Add options to hide MultiGrid top-right and bottom-left grid scrollbars but still make the scroll events work on those grids

RaviDasari:
The main issue here is that we want scroll events to work on top-right and bottom-left grids but dont want those scrollbars. I fixed this issue with simple css and an extra container. This [stackoverflow](https://stackoverflow.com/questions/16670931/hide-scroll-bar-but-while-still-being-able-to-scroll) is the inspiration for my solution. here is how it looks - observe that in the below demo, I could use mouse scroll to scroll through the MultiGrid though the top-right and bottom-left grids doesn't show any scrollbars.

image
I tested this in latest chrome, firefox, edge and IE11 and it works. in short, I am simply hiding the scrollbar by adding an extra container. So I am sure this works in all browsers.


RaviDasari:
I am running tests and will add tests soon. @bvaughn what do you think of this fix?

source/MultiGrid/MultiGrid.js
```diff
@@ -732,16 +773,31 @@ export default class MultiGrid extends React.PureComponent {
...
     );
+
+    if (hideTopRightGridScrollbar) {
+      return (
+        <div
+          style={{
+            ...this._topRightGridStyle,
+            height,
+            width,
+            overflowX: 'hidden',
```
wuweiweiwu: I don't think you need to wrap it in another div

If you need to. you can just modify `this._topRightGridStyle`

```js
this._topRightGridStyle = {
   ...this._topRightGridStyle,
   height,
   width, 
   overflowX: 'hidden'
}
```

RaviDasari: Yes. We need this extra wrapper so we can hide the scrollbar with a hack.

source/MultiGrid/MultiGrid.js
```diff
@@ -713,16 +738,32 @@ export default class MultiGrid extends React.PureComponent {
...
 
     if (!fixedRowCount) {
       return null;
     }
 
-    const additionalColumnCount = showHorizontalScrollbar ? 1 : 0;
+    const additionalColumnCount = showHorizontalScrollbar ? 1 : 0,
+      height = this._getTopGridHeight(props),
+      width = this._getRightGridWidth(props),
+      scrollbarSize = this.state.showHorizontalScrollbar ? this.state.scrollbarSize : 0;
```
wuweiweiwu: nit: `showHorizontalScrollbar` is already deconstructed above. `scrollbarSize` can also be put there.

source/MultiGrid/MultiGrid.js
wuweiweiwu: Ah I see. Thanks

source/MultiGrid/MultiGrid.js
```diff
@@ -47,6 +49,8 @@ export default class MultiGrid extends React.PureComponent {
     styleBottomRightGrid: {},
     styleTopLeftGrid: {},
     styleTopRightGrid: {},
+    hideTopRightGridScrollbar: false,
```
wuweiweiwu: nit: I don't think there is a need to specify defaultProps for these. Since it would be `undefined` if not passed in. Which is a falsey value

mattpage:
What's the status on this PR?

source/MultiGrid/MultiGrid.js
RaviDasari: I added because there was an existing pattern for `enableFixedColumnScroll` and `enableFixedRowScroll`. So followed it.

RaviDasari:
@wuweiweiwu , I think I addressed all the comments and added tests.

Also added docs.

DGulshan:
Any updates on this PR?

source/MultiGrid/MultiGrid.js
RaviDasari: @wuweiweiwu , are you waiting for some change here?

wuweiweiwu:
@RaviDasari @DGulshan looks good to me. I want to wait for #1053 to get in first

wuweiweiwu:
I'll to make sure that the test pass with the new changes in Grid. but it looks good so far!

source/MultiGrid/MultiGrid.js
```diff
@@ -611,32 +615,54 @@ class MultiGrid extends React.PureComponent {
...
 
     if (!fixedColumnCount) {
       return null;
     }
 
-    const additionalRowCount = showVerticalScrollbar ? 1 : 0;
+    const additionalRowCount = showVerticalScrollbar ? 1 : 0,
+      height = this._getBottomGridHeight(props),
+      width = this._getLeftGridWidth(props),
+      scrollbarSize = this.state.showVerticalScrollbar ? this.state.scrollbarSize : 0,
```
wuweiweiwu: `scrollbarSize` is now in `state.instanceProps`.

RaviDasari: I think state structure change in Grid but not in MutliGrid so it is working fine. 
Also, I ran `yarn lint` and it seem to be all good too.

So we are good here?

RaviDasari:
@wuweiweiwu looks like my code had couple of lint errors, fixed them along with few other issues I found in other parts of the code. Though it is irrelevant as part of this PR but I hope it is fine.

wuweiweiwu:
Looks good! tests pass. I think this can land in the 9.19.0 release

Issue #1012, opened 2018-02-13
InfiniteLoader and Table


cristianfraser:
I'm hitting a wall over here, I think it might be a bug. (Using version `9.13.0`, I'll try this updating a bit later, but the changelog doesn't make me think it's fixed?)

So I have a component that's like this (omitting boilerplate):

```
rowCount = this.rowCount()

return (
  <InfiniteLoader
    isRowLoaded={this._isRowLoaded}
    loadMoreRows={this._loadMoreRows}
    rowCount={rowCount}
  >
    <Table
      rowGetter={rowGetter}
      rowRenderer={rowRenderer}
      rowCount={rowCount}
    >
      {columns.map(column => <Column {...column} />)}
    </Table>
  </InfiniteLoader>
)
```

`columns` is an array of objects with `Column`'s properties (`label`, `dataKey`, `width`)

The problem arises with `InfiniteLoader`:

* The component mounts, data isn't fetched yet, so `data = [], data.length = 0`
* `rowCount` is going to be `1`(!!!), because there are more available items to be fetched.
* This row is out of the available data: it is not loaded yet, so I would like to render a `LoadingRow` component for this one row.

Using `List` instead of `Table` this is rather simple, in the `rowRenderer` method check if that row is loaded using `isRowLoaded` and render accordingly. But I haven't been able to do the same using `Table`, because before `rowRenderer` gets called, I get an `Uncaught TypeError` in `defaultCellDataGetter` here:

```
export default function defaultCellDataGetter(_ref) {
  var dataKey = _ref.dataKey,
      rowData = _ref.rowData;

  if (typeof rowData.get === 'function') {   <<<=== HERE
    return rowData.get(dataKey);
  } else {
    return rowData[dataKey];
  }
}
```

Because `rowData` is `undefined`, and that's because the `rowGetter` function is like this `rowGetter = (list, index) => list[index]`. So before being able to check in `rowRenderer` whether the row is loaded or not, it crashes in `defaultCellDataGetter`, because `rowData` is `undefined`.

I could get around this just adding `cellDataGetter` to each column, but it doesn't look right.

That's all I think. I don't know if I'm getting this right (and it is a bug), or if there is another way this should be done.

bvaughn:
> That's all I think. I don't know if I'm getting this right (and it is a bug), or if there is another way this should be done.

It's not a bug.

[...]

PR #1046, opened 2018-03-12
doc changes (small)

ryanbas21:
grammar

Just browsing through docs, made small grammar fixes when I saw them. Looking at it as if I was a complete beginner, I was curious if the term higher order component should be assumed already, so I added that small definition in parentheses. 

### DIFF
--- a/docs/ArrowKeyStepper.md
+++ b/docs/ArrowKeyStepper.md
@@ -1,7 +1,7 @@
 ArrowKeyStepper
 ---------------
 
-High-order component that decorates another virtualized component and responds to arrow-key events by scrolling one row or column at a time.
+A High-order component (A component that takes a component and returns a component) that decorates another virtualized component and responds to arrow-key events by scrolling one row or column at a time.
 This provides a snap-to behavior rather than the default browser scrolling behavior.
 
 Note that unlike the other HOCs in react-virtualized, the `ArrowKeyStepper` adds a `<div>` element around its children in order to attach a key-down event handler.
--- a/docs/ArrowKeyStepper.md
+++ b/docs/ArrowKeyStepper.md
@@ -1,25 +1,25 @@
 ArrowKeyStepper
 ---------------
 
-A High-order component (A component that takes a component and returns a component) that decorates another virtualized component and responds to arrow-key events by scrolling one row or column at a time.
+A High-order component (Decorates multiple virtualized components such as `Grid` or `List` in order to synchronize their scroll positions. Passes synchronized scroll offsets to a render prop.) that decorates another virtualized component and responds to arrow-key events by scrolling one row or column at a time.
 This provides a snap-to behavior rather than the default browser scrolling behavior.
 
 Note that unlike the other HOCs in react-virtualized, the `ArrowKeyStepper` adds a `<div>` element around its children in order to attach a key-down event handler.
 The appearance of this wrapper element can be customized using the `className` property.
 
 ### Prop Types
-| Property | Type | Required? | Description |
-|:---|:---|:---:|:---|
-| children | Function | ✓ | Function responsible for rendering children. This function should implement the following signature: `({ onSectionRendered: Function, scrollToColumn: number, scrollToRow: number }) => PropTypes.element` |
-| className | String |  | CSS class name to attach to the wrapper `<div>`. |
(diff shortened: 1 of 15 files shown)


docs/ArrowKeyStepper.md
```diff
@@ -1,7 +1,7 @@
 ArrowKeyStepper
 ---------------
 
-High-order component that decorates another virtualized component and responds to arrow-key events by scrolling one row or column at a time.
+A High-order component (A component that takes a component and returns a component) that decorates another virtualized component and responds to arrow-key events by scrolling one row or column at a time.
```
TrySound: I'd say it's not higher order component. Can you call it `A component with children as function that wraps...`?

docs/ScrollSync.md
```diff
@@ -1,7 +1,7 @@
 ScrollSync
 ---------------
 
-High order component that simplifies the process of synchronizing scrolling between two or more virtualized components.
+High-order component that simplifies the process of synchronizing scrolling between two or more virtualized components.
```
TrySound: `A component with children as function that...`

ryanbas21:
@TrySound So I've actually never heard the term `high-order` component. What would the definition of it be? I even asked at work and no one knew.

TrySound:
Higher order component is function which wraps react component and returns new one
```js
import { pure } from 'recompose'

const enhanced = pure(
  ({ width, height }) => <div style={{ width, height }}></div>
)
```

it's mostly not quite useful since introduces props collision. Render prop or children as function is more clean way to do things this time. Also children as functions is used for react stuff (context api).

ryanbas21:
I'm confused now, is `high order === higher order`? Or are they different terms. I understand a `higher order component` as a component that takes a `Component` and returns a `Component`, usually applying some new props to said component. 

So when we say a `high-order` component are we saying more like a render prop?

TrySound:
Probably spelling

docs/ArrowKeyStepper.md
bvaughn: Yeah, I guess these docs are a little out of date. `ArrowKeyStepper` used to be implemented more as a HOC but changed to use the function children / render props a long time ago. Maybe the docs could be updated to say something more like...

> Decorates another react-virtualized component such as `Grid` or `List`. Passes `scrollToColumn` and `scrollToRow` props to a render function to enable user to navigate content using arrow keys.

docs/ScrollSync.md
bvaughn: Maybe something like

> Decorates multiple virtualized components such as `Grid` or `List` in order to synchronize their scroll positions. Passes synchronized scroll offsets to a render prop.

chris-fa:
@ryanbas21 since you're doing a PR on docs, what do you think on documenting that the `<Table>` `rowGetter` property should not receive a null nor undefined, but a blank object, too?

Here's the issue for reference: https://github.com/bvaughn/react-virtualized/issues/1012, check my last comment

ryanbas21:
@chris-fa sure! I'll check it out later tonight.

docs/ScrollSync.md
ryanbas21: @bvaughn Cool, this makes sense. is this something that I should include in each instance of `high-order` or as a separate glossary type page?

aem:
As a project it makes sense to settle on a terminology, and "higher-order" is the accepted term afaik. While "high-order" is an absolute term, "higher-order" references the fact that it's simply a function of a higher order than the function it wraps, which makes sense to me.

> since you're doing a PR on docs, what do you think on documenting that the <Table> rowGetter property should not receive a null nor undefined, but a blank object, too?

@chris-fa we should keep changes focused, it doesn't make sense to lump more docs changes into this PR just because it's open. Unless those changes are related to these, let's open a separate PR for those

ryanbas21:
@aem I generally think higher order in terms of components being it will take and return although I guess you could say it has to take and or return a component. It definitely would be a more understood term, but when I first read high-order I and others I asked just assumed it was another way of saying higher order, but with how this discussion has gone I haven't felt that its the same definition.

ryanbas21:
FWIW - it seems the markdown plugin I installed also reformatted the markdown to make stuff line up - if that is unwanted I can revert this.

ryanbas21:
closing cause i blew up my git.

PR #1047, opened 2018-03-12
Change `import` to `import type` to fix circular dependency

ahutchings:
This fixes a circular dependency warning when bundling react-virtualized with rollup.

```
(!) Circular dependency: node_modules/react-virtualized/dist/es/WindowScroller/WindowScroller.js -> node_modules/react-virtualized/dist/es/WindowScroller/utils/onScroll.js -> node_modules/react-virtualized/dist/es/WindowScroller/WindowScroller.js
```


bvaughn:
Thanks~

PR #1051, opened 2018-03-15
Fix create multi sort bug

lideeyaa:
This PR fixes a bug found in the `createMultiSort` function. When selecting a new column to sort, the original code clears the `sortBy` array of all non-selected keys, however, fails to clear the `sortDirection` object. By failing to do so, multi-select will not work on any future shift+clicks. Let me explain: in the `event.shiftClick` if-statement block starting on line 59, since the old keys remain in the `sortDirection` object, it will never push the new column value to the `sortBy` array, and thus only ever returning one key in `sortBy`. In my PR, I make sure the `sortDirection` object is cleared of all the non-selected key-values.

### DIFF
--- a/source/Table/createMultiSort.jest.js
+++ b/source/Table/createMultiSort.jest.js
@@ -86,6 +86,29 @@ describe('createMultiSort', () => {
       simulate(multiSort.sort, 'a');
       expect(multiSort.sortBy).toEqual(['a']);
     });
+
+    it('resets sort-direction fields', () => {
+      const multiSort = createMultiSort(jest.fn(), {
+        defaultSortBy: ['a', 'b'],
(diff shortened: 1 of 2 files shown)


aem:
Hi there! Can you please add some positive and negative test cases to verify your PR's changes? That will help track if this breaks in the future

lideeyaa:
Okay. I added some tests. Are those sufficient?

mikeldking:
I can confirm this bug. @aem it would be awesome if this can get merged!  Also thanks @lideeyaa for the fix.
multisortbug

wuweiweiwu:
@lideeyaa Thanks for contributing! Planning to release this soon!

Issue #1026, opened 2018-02-27
CellMeasurer propTypes shouldn't expect instance of CellMeasurerCache


kof:
Currently CellMeasurer explicitly checks the type of cache prop to be CellMeasurerCache. If it is expected that user is able to implement his own Cache class, then one shouldn't validate that way, instead it should be checking for specific methods to be implemented.

kof:
I worked this around temporarily by extending CellMeasurerCache but seems unnecessarily.

diogofcunha:
We should create an interface with the public api and type check based on that, nice catch. 
Will do that tomorrow

PR #1058, opened 2018-03-17
Extracted CellMeasurer public interface

diogofcunha:
Fixes https://github.com/bvaughn/react-virtualized/issues/1026.

@bvaughn not totally confident about flow typings as I'm honestly not aware of best practices, shall we copy the types for the CellMeasurer as it's done for WindowScroller so that they can be pulled with the module and users can implement the interface or it's there anything I'm not aware of.

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,4 +1,4 @@
-  const env = process.env.NODE_ENV;
+const env = process.env.NODE_ENV;
 
 if (env === 'commonjs' || env === 'es') {
   module.exports = {
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,4 +1,4 @@
-const env = process.env.NODE_ENV;
+  const env = process.env.NODE_ENV;
 
 if (env === 'commonjs' || env === 'es') {
   module.exports = {
@@ -13,7 +13,7 @@ if (env === 'commonjs' || env === 'es') {
     ],
     plugins: [
       'transform-runtime',
-      ['flow-react-proptypes', {deadCode: true}],
+      ['flow-react-proptypes', {deadCode: true, useESModules: true}],
       ['transform-react-remove-prop-types', {mode: 'wrap'}],
     ],
     presets: [['env', {modules: false}], 'react', 'flow', 'stage-2'],
--- a/source/CellMeasurer/CellMeasurer.js
+++ b/source/CellMeasurer/CellMeasurer.js
@@ -1,7 +1,7 @@
 /** @flow */
 import * as React from 'react';
 import {findDOMNode} from 'react-dom';
-import CellMeasurerCache from './CellMeasurerCache.js';
+import {CellMeasureCache} from './CellMeasurerCache.js';
 
 type Children = (params: {measure: () => void}) => React.Element<*>;
 
@@ -11,7 +11,7 @@ type Cell = {
 };
 
 type Props = {
-  cache: CellMeasurerCache,
+  cache: CellMeasureCache,
   children: Children | React.Element<*>,
   columnIndex?: number,
   index?: number,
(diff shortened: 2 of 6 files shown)


source/CellMeasurer/CellMeasurer.js
```diff
@@ -1,7 +1,7 @@
 /** @flow */
 import * as React from 'react';
 import {findDOMNode} from 'react-dom';
-import CellMeasurerCache from './CellMeasurerCache.js';
+import {CellMeasureCache} from './CellMeasurerCache.js';
```
bvaughn: This won't actually change the generated prop-types:
```js
  cache:
    typeof _CellMeasurerCache.CellMeasureCache === "function"
      ? require("prop-types").instanceOf(_CellMeasurerCache.CellMeasureCache)
          .isRequired
      : require("prop-types").any.isRequired
```

diogofcunha: Made a few changes and the generated prop types now look like:
Will test it against react-virtualized-tree soon.

```
var babelPluginFlowReactPropTypes_proptype_CellMeasureCache = require('./types').babelPluginFlowReactPropTypes_proptype_CellMeasureCache || require('prop-types').any;

// ...
cache:
  typeof babelPluginFlowReactPropTypes_proptype_CellMeasureCache === 'function' 
    ? babelPluginFlowReactPropTypes_proptype_CellMeasureCache.isRequired 
      ? babelPluginFlowReactPropTypes_proptype_CellMeasureCache.isRequired 
      : babelPluginFlowReactPropTypes_proptype_CellMeasureCache 
    : require('proptypes').shape(babelPluginFlowReactPropTypes_proptype_CellMeasureCache).isRequired

```

diogofcunha: @bvaughn our current version of **babel-plugin-flow-react-proptypes** is far behind (^13.0.0) when the package is on 21.0.0.

Interface support was added in 17.0.0 (https://github.com/brigand/babel-plugin-flow-react-proptypes/pull/175), with the current version the propType for the interface is not being generated, causing the generated **types.js** to be empty. 

Is there any know blocker in updating the package or can I go for that?

TrySound: I upgraded it some time ago and published 5 releases trying to fix its issues. We can try again.

diogofcunha: Thanks @TrySound, was checking the release history and can see we tried to update to 15.0.0 at the time, were our issues caused by this bug (https://github.com/brigand/babel-plugin-flow-react-proptypes/commit/890c166ebc0f4db6df05fe6e58bebb322fccaae3) fixed in 16.0.0 release?

diogofcunha: will probably just cherry pick your original commit and change the version to the latest

TrySound: There was a few issues. I guess they are fixed now. I'm gonna land it after rollup integration.

diogofcunha: I have good news, tried to upgrade and all seems good on rv side, the build output seems to be converting flow types to propTypes correctly.

Will use a local npm link to test it against https://github.com/diogofcunha/react-virtualized-tree. If it works there I guess it's good to go.

diogofcunha: Tested this and after adding useESModules flag everything seems to be working fine for babel-plugin-flow-react-proptypes update, no tests were broken and I can't see any runtime errors.

Tomorrow will try to add a custom cache just to make sure the original issue gets resolved.

.babelrc.js
```diff
@@ -1,4 +1,4 @@
-const env = process.env.NODE_ENV;
+  const env = process.env.NODE_ENV;
```
TrySound: remove spaces

PR #1064, opened 2018-03-26
Optimize webpack by setting sideEffects to false

0xR:
This allows webpack 4 to optimize the bundle by only importing the used modules instead of everything.

(Admittedly I didn't read all of the source code so I don't know whether there is some module which does have side effects)

[Webpack documentation](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free)
[Webpack example](https://github.com/webpack/webpack/tree/master/examples/side-effects)


TrySound:
Thanks

Issue #587, opened 2017-02-21
scrollToIndex will not SSR the specified index.


viqh:
Hi All,

**First of all thank you for such a great component!**

Recently tried to SSR the `{ List }` component and could not server side the specific index. 
This is the code (a small POC) that I'm running.

**App.js**
```
import React, { Component } from 'react';
import { List } from 'react-virtualized';

const generateImgs = (n) => {
  let imgs = [];

  for (let i = 0; i < n; i++) {
    const imgHeight = i > 9 ? `8${i}` : `80${i}`;
    const imgUrl = `http://placehold.it/800x${imgHeight}`;
    const styles = { height: `${imgHeight}px`, border: '1px solid black' };

    imgs.push({
      img: <img key={i} src={imgUrl} style={styles} alt={`${imgHeight}`} />,
      height: imgHeight,
    });
  }
  return imgs;
};

export default class App extends Component {
  constructor(props) {
    super(props);

    this.list = [...generateImgs(100)];
  }

  static defaultProps = { goToIndex: 0 };

  rowRenderer = ({ key, index, isScrolling, isVisible, style }) => {
    return (
      <div
        key={key}
        style={style}
      >
        {this.list[index].img}
      </div>
    );
  };

  getRowHeight = ({ index }) => {
    return parseInt(this.list[index].height, 10);
  };

  render() {
    const { goToIndex } = this.props;

    return (
      <div style={{ width: '800px', margin: '0 auto' }}>
        <List
          height={700}
          overscanRowCount={3}
          rowCount={this.list.length}
          rowHeight={this.getRowHeight}
          rowRenderer={this.rowRenderer}
          scrollToIndex={goToIndex}
          width={800}
        />
      </div>
    );
  }
}
```

**server.js**

```
import express from 'express';
import React from 'react';
import { renderToString } from 'react-dom/server';
import App from './app/App';
import template from './template';
const PORT = 3333;

const server = express();

server.use('/assets', express.static('assets'));

server.get('/', (req, res) => {
  const initialState = {
    goToIndex: 50
  };

  const appString = renderToString(<App {...initialState} />);
  console.log('appString = ', appString);

  res.send(template({
    body: appString,
    initialState: JSON.stringify(initialState)
  }));
});

server.listen(PORT);
console.log(`Listening on http://localhost:${PORT}`);
```

**SSR output "appString ="**
```
[...]

PR #1053, opened 2018-03-15
feat: Migrating Grid.js to getDerivedStateFromProps

wuweiweiwu:
follow up to #1009 

especially need eyes on these parts:
https://github.com/bvaughn/react-virtualized/pull/1053/files#diff-43c6f2c20a98e4816031a2f99092ecb4R791

https://github.com/bvaughn/react-virtualized/pull/1053/files#diff-868d04d08bdf53ce86a26e3f8a99bccdR2316

cc @bvaughn

PR #1072, opened 2018-04-09
Added scrollToRow and scrollToColumn support for ssr

microcood:
Render cells defined with scrollToRow and scrollToColumn with server dom. I consider this as an intuitive behaviour, also was not confronted in https://github.com/bvaughn/react-virtualized/issues/587


wuweiweiwu:
@microcood This is currently blocked by #1053 We'll take a look at this as soon as that is in.

wuweiweiwu:
@microcood #1053 is in! There are some changes to how some values (ex: `scrollBarMeasured`) are stored, in `instanceProps` So if you can update your PR to reflect the new API it'll be awesome and we can merge this! 

:)

microcood:
Updated :point_up:

PR #1073, opened 2018-04-10
Add table rowkey

j3k0:
This PR adds a new property to the `Table` element, namely `rowKey`.

As its name implied, it allows to customize React `key` for each row.

The obvious reason is to leverage existing identifiers in the users data, instead of relying on indexes. More practically, this also solves the bug described below (which might be of interest on its own).

---

The bug this solves.

I have have dynamic data in my table, with rows frequently inserted and deleted. Using event delegation on the whole table element, I determine which cell was clicked. Cells have a custom rendered which sets `data-something="whatever"`, that's how I extract my data the event's target element.

After removing a few rows, I discover the target elements had `data-something` pointing to the wrong data: that of the cell that was previously at the same location. Suspecting some optimization-related-code reusing the same DOM element because it has the same key and type, I tried this quick fix that got it worked.

I'm beginner-level with React, so unsure if that might point out to a deeper issuer.

Anyways, as I read about it, it seems like the bug I have is identical to what is described here: [using indexes as keys is an anti-pattern](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318).

> Let me explain, a key is the only thing React uses to identify DOM elements. What happens if you push an item to the list or remove something in the middle? If the key is same as before React assumes that the DOM element represents the same component as before. But that is no longer true.

So, using smarter row keys seems to be the answer to such bugs.

### DIFF
--- a/docs/Table.md
+++ b/docs/Table.md
@@ -37,6 +37,7 @@ This component expects explicit `width` and `height` parameters.
 | rowHeight | Number or Function | ✓ | Either a fixed row height (number) or a function that returns the height of a row given its index: `({ index: number }): number` |
 | rowRenderer | Function |  | Responsible for rendering a table row given an array of columns. [Learn more](#rowrenderer) |
 | rowStyle | Object or Function |  | Optional custom inline style to attach to table rows. This value may be either a style object or a function with the signature `({ index: number }): Object`. Note that for the header row an index of `-1` is provided. |
+| rowKey | String or Function |  | Either the `dataKey` to use as a unique identifier or a function with the signature `({ rowData: any }): String. When no `rowKey` is provided, a row identifier is generated from its index. |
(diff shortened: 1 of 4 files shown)


j3k0:
Any chance to have this bug fix enter master?

j3k0:
Anyone?

wuweiweiwu:
I don't think specifying a custom `rowKey` is necessary. You can use [`event.target`](https://developer.mozilla.org/en-US/docs/Web/API/Event/target) for your use case. For such use cases, it usually doesn't require an API change but rather how you're using the component

j3k0:
@wuweiweiwu Ok, regardless of the bug. React specifies in the doc that using indexes as key is a bad idea.

From https://reactjs.org/docs/lists-and-keys.html#keys

> We don’t recommend using indexes for keys if the order of items may change. This can negatively impact performance and may cause issues with component state.

I (and maybe others) fell into a bug due to that, react is allowed to consider the inner content as identical when the key doesn't change.

PR #1076, opened 2018-04-13
make maxscrollmax size configurable

scarlettlite:

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -31,7 +31,7 @@ import {
   requestAnimationTimeout,
   cancelAnimationTimeout,
 } from '../utils/requestAnimationTimeout';
-
+import {getMaxElementSize} from './utils/maxElementSize.js'; 
 /**
  * Specifies the number of milliseconds during which to disable pointer events while a scroll is in progress.
  * This improves performance and makes scrolling smoother.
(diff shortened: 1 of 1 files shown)


TrySound:
It's not something which should be configurable. It's constant value which depends on browser abilities.

scarlettlite:
if that is the case how would react virtualised work for react components which have scroll height more than 1.5 million? This breaks 'scrollToRow' use case . I need this to work for my app. Are are any other suggestions ? @TrySound @bvaughn

paulkenney:
@TrySound @bvaughn Is there a specific reason these values should not be changed? Are there specific problems in any specific browsers with higher values?

TrySound:
@paulkenney Yeah, you should find safe values or you will have even bigger problems. For example when chrome reaches limit it just stop render after it. When firefox reaches limit it starts proportionally decrease width. So different browsers even behaves differently. Currently we hardcoded size for chrome which is `1.67771e7 `, others get IE size.

@scarlettlite `scrollToRow` should work. Virtualized scales scroll when the size is more then calculated in this function. Probably you found a bug. Can you create [codesandbox](http://codesandbox.io/) example which reproduces the problem?

scarlettlite:
https://codesandbox.io/s/rl5jv04k3q I created this example the reproduces the issue. Here I take as input the div and where In div ( the vertical offset in div ) I want to scroll to and then compute the scroll top that I pass to the grid . On Safari/Firefox if the scroll Height of the virtualised grid exceeds 1.5 million this calculation fails.

mraiguo:
May I ask, which version of Internet explorer does version 8.x support?

PR #1080, opened 2018-04-17
add pre-commit to run prettier and lint

bardiarastin:
this will add running prettier and lint before committing, @wuweiweiwu  what do you think about adding test too ?


bvaughn:
I don't really object to this.

PR #1084, opened 2018-04-22
Fix #1083, in InfiniteLoader prefer multiples of minimumBatchSize

### LINKED ISSUE #1083: InfiniteLoader spams loadMoreRows on scroll with slow API / small minimumBatchSize
Similar to #345, ran into it because of an unrelated bug in my code leading to slow API responses. See the impact of this issue in [this CodePen](https://codepen.io/luke-gumbley/pen/mxopjR?editors=1010).

Looking at the memoization code, I'd like to suggest adding a preference in [scanForUnloadedRanges](https://github.com/bvaughn/react-virtualized/blob/5fa323680c17023f06c35dfd1f928a42f3894031/source/InfiniteLoader/InfiniteLoader.js#L173) to load multiples of minimumBatchSize where possible. This would reduce the volume of duplicated calls.

I understand that I should be storing loaded rows in the state of my component (per the [documentation](https://github.com/bvaughn/react-virtualized/blob/master/docs/InfiniteLoader.md#tracking-loaded-and-loading-rows)), and I'll absolutely do that. I also know this tweaks the meaning of minimumBatchSize... and if you're storing loaded rows in the consuming component, as you should, this issue has no impact. However it was a very small change so I haven't lost much time if you don't like it :)

luke-gumbley:
For the first and last unloaded range, instead of accepting any size
over minimumBatchSize, scan forward/backward until reaching a multiple
of minimumBatchSize (if possible). Modify tests to reflect the new
behaviour.

### DIFF
--- a/source/InfiniteLoader/InfiniteLoader.jest.js
+++ b/source/InfiniteLoader/InfiniteLoader.jest.js
@@ -175,7 +175,7 @@ describe('InfiniteLoader', () => {
     it('should not interfere with :threshold', () => {
       render(
         getMarkup({
-          minimumBatchSize: 10,
+          minimumBatchSize: 15,
           threshold: 10,
         }),
       );
@@ -263,7 +263,7 @@ describe('InfiniteLoader', () => {
     expect(loadMoreRowsCalls).toEqual([{startIndex: 0, stopIndex: 19}]);
     loadMoreRowsCalls.splice(0);
     innerOnRowsRendered({startIndex: 0, stopIndex: 20});
-    expect(loadMoreRowsCalls).toEqual([{startIndex: 0, stopIndex: 20}]);
+    expect(loadMoreRowsCalls).toEqual([{startIndex: 0, stopIndex: 39}]);
   });
 
   it('resetLoadMoreRowsCache should reset memoized state', () => {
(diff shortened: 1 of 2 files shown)


bvaughn:
Why is this better?

luke-gumbley:
Issue #1083 has an example. The way it stands, if you can scroll through (minimumBatchSize - threshold) rows faster than the API can load them, InfiniteLoader will then call loadMoreRows for every row you continue to scroll past. This is because scanForUnloadedRanges only specifies a _minimum_ size for the first and last ranges, which defeats the memoization (because the upper bound changes by one row each time).

For a potentially unbounded range (such as the first and last range), specifying the size be a multiple of minimumBatchSize reduces the number of calls in this scenario by a factor of minimumBatchSize.

If there are 5 visible rows, threshold is 10, and minimumBatchSize is 20, before the change you get:
scroll at 0: loadMoreRows(0,19)
scroll to 1-4: <no call>
scroll to 5: loadMoreRows(0,20)
scroll to 6: loadMoreRows(0,21)
scroll to 7: loadMoreRows(0,22)
etc.

After the change, you get:
scroll at 0: loadMoreRows(0,19)
scroll to 1-4: <no call>
scroll to 5: loadMoreRows(0,39)
scroll to 6-24 <no call>

None of this changes the underlying requirement that consumers maintain their own record of which rows have been requested. It's a minor improvement to the memoization in cases where that record is not being kept, and results in no change at all if that record is being kept.

PR #1111, opened 2018-05-17
bind requestAnimationFrame to window for IE11

jorgejar:
* currently InfiniteLoader does not trigger LoadMoreRows in IE11 because it throws an error when requestAnimationFrame is not called with the window context. 

 * this PR simply binds the window to requestAnimationFrame so that when it is called, it works correctly

### DIFF
--- a/source/utils/animationFrame.js
+++ b/source/utils/animationFrame.js
@@ -16,7 +16,7 @@ if (typeof window !== 'undefined') {
 
 // requestAnimationFrame() shim by Paul Irish
 // http://paulirish.com/2011/requestanimationframe-for-smart-animating/
-const request =
+const request = (
   win.requestAnimationFrame ||
   win.webkitRequestAnimationFrame ||
   win.mozRequestAnimationFrame ||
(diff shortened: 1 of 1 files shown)


jorgejar:
fixed formatting and squashed commits.

I just found this problem only arises with dev mode with webpack, probably how its bundled with `eval`, production builds work fine.
Ill keep the PR up in case it there is interest in fixing the dev mode experience.

wuweiweiwu:
hm. interesting.

@jorgejar do you mind creating an issue? so we'll make a note of it and prioritize accordingly. Thanks :)

wuweiweiwu:
Also, can you try running `yarn start` to start the development server for the demo website and see if the error persists? It currently works for Chrome on MacOS

jorgejar:
> Also, can you try running yarn start to start the development server for the demo website and see if the error persists? It currently works for Chrome on MacOS

this only happened on development bundles for IE11, the result from  `yarn start` will not render in IE11, but it happened in the project I was building. 
I checked the issues, it looks like this would solve #986. This PR has the same solution as #1013, so Ill let the discussion happen on those threads.

PR #1114, opened 2018-05-17
upgrade react-lifecycles-compat

pigcan:
```js
(0, _reactLifecyclesCompat2.default)(Grid);
```

-->

```
_reactLifecyclesCompat2.default is not a function
```

----

`edit:`

It was confirmed that this bug was caused by my local multiple react-lifecycles-compat versions.

So I change this PR's title.

But I'm looking forward this PR could be merged.

I fixed eslint errors and upgraded react-lifecycles-compat to newest version.

### DIFF
--- a/source/ScrollSync/ScrollSync.example.js
+++ b/source/ScrollSync/ScrollSync.example.js
@@ -236,12 +236,8 @@ export default class GridExample extends React.PureComponent {
   _renderLeftSideCell({columnIndex, key, rowIndex, style}) {
     const rowClass =
       rowIndex % 2 === 0
-        ? columnIndex % 2 === 0
-          ? styles.evenRow
-          : styles.oddRow
-        : columnIndex % 2 !== 0
-          ? styles.evenRow
-          : styles.oddRow;
+        ? columnIndex % 2 === 0 ? styles.evenRow : styles.oddRow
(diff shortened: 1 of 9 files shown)


source/ScrollSync/ScrollSync.example.js
```diff
@@ -236,8 +236,12 @@ export default class GridExample extends React.PureComponent {
   _renderLeftSideCell({columnIndex, key, rowIndex, style}) {
     const rowClass =
       rowIndex % 2 === 0
-        ? columnIndex % 2 === 0 ? styles.evenRow : styles.oddRow
-        : columnIndex % 2 !== 0 ? styles.evenRow : styles.oddRow;
+        ? columnIndex % 2 === 0
```
wuweiweiwu: Why are these changed?

pigcan:
```
239:32  error  Replace `·?·styles.evenRow` with `⏎··········?·styles.evenRow⏎·········`  prettier/prettier
  240:32  error  Replace `·?·styles.evenRow` with `⏎··········?·styles.evenRow⏎·········`  prettier/prettier
```

pigcan: Because these codes would cause eslint errors so I fixed it 0.0

Issue #1119, opened 2018-05-24
9.19 throws setState(...): You passed an undefined or null state object; instead, use forceUpdate()


hyperknot:
Upgrading to 9.19 started throwing a React warning with React 15.6:
```
Warning: setState(...): You passed an undefined or null state object; instead, use forceUpdate()
```

I'm using it like this:

```
import AutoSizer from 'react-virtualized/dist/commonjs/AutoSizer'
import List from 'react-virtualized/dist/commonjs/List'

@observer
export class FeatureList extends React.Component {
  rowRenderer({ key, index, style }) {
    return <FeatureListItem key={key} style={style} feature={featureStore.featureArray[index]} />
  }

  render() {
    if (!mapUIStore.ready) return null

    return (
      <ul id="panel-items-list">
        <AutoSizer>
          {({ height, width }) => (
            <Observer>
              {() => (
                <List
                  width={width}
                  height={height}
                  rowCount={featureStore.count}
                  rowHeight={31}
                  rowRenderer={this.rowRenderer}
                  scrollToIndex={mapUIStore.scrollToIndex}
                />
              )}
            </Observer>
          )}
        </AutoSizer>
      </ul>
    )
  }
}
```

The warning happens when scrollToIndex is updated.

wuweiweiwu:
react 16 allows return `null` from `setState` to avoid a re-render. 

https://teamtreehouse.com/library/returning-null-from-setstate.

I would suggest upgrading react. but it's also just a warning in dev.

I'll look into refactoring some `setState` calls to avoid setting `null`

Hope that helps!

wuweiweiwu:
Its because there was no changes that weren't backwards compatible following semvar and the rewrite to 16.x lifecycle functions had no api changes.

wuweiweiwu:
@TrySound Should this be a breaking changes?

TrySound:
Yep this is a breaking change. Try to fix it. We will return null in v10.

wuweiweiwu:
Gotcha

Issue #1117, opened 2018-05-22
Grid: rendering using react-test-renderer breaks in 9.19.0


wuweiweiwu:
```js
import renderer from 'react-test-renderer';

const tree = renderer
      .create(<Grid ... />)
      .toJSON();
```

Error message: 
```js
      Error: Uncaught [TypeError: Cannot read property 'scrollLeft' of null]
```

Possibly not working because react-test-renderer doesn't support the polyfilled lifecycle functions?

wuweiweiwu:
https://github.com/facebook/react/issues/7740 react-test-renderer. Doesn't support refs. Need to pass in `createNodeMock` option

nihgwu:
@wuweiweiwu but how can I pass `createNodeMock` to `Grid`? I created a `Table` component on `Grid`, and testing the `Table` component

nihgwu:
can we check before accessing the ref here? https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/Grid.js#L627

wuweiweiwu:
`createNodeMock` is an option in `react-test-renderer` 
https://reactjs.org/docs/test-renderer.html#ideas

wuweiweiwu:
If you want to contribute a PR that'll be awesome!!

nihgwu:
@wuweiweiwu seems `createNodeMock` is applied to all the nested components when creating, but that's still odd to me that I have to add that option manually in every test case, I'm testing my own `Table` component, and don't use refs in `cDM`, `Grid` should be transparent for me

PR #1121, opened 2018-05-24
make test in react-test-renderer painless

nihgwu:
ref #1117


nihgwu:
Done, without this change, the tests will fail

nihgwu:
BTW, can we release a new version? that breaks all my test cases, now I have to use a wrapper to apply the mock option, also I don't think that's the desired solution, as that would make people thought I was using refs in `cDM` of my `Table` but that's not true

wuweiweiwu:
@nihgwu yep We'll release a minor soon. Probably after I fix #1119 . So for now, you'll have to use `createNodeMock` locally

wuweiweiwu:
Thanks for contributing @nihgwu !

wuweiweiwu:
this is released in 9.20.0!

wuweiweiwu:
a work around is to use `createNodeMock`

ivosh:
@wuweiweiwu Excellent! This is really appreciated.

PR #1129, opened 2018-06-01
No `setState(null)`

wuweiweiwu:
fixes: #1119 

react 15 throws warning when `setState` is called with `null` or `{}`

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -1440,7 +1440,6 @@ class Grid extends React.PureComponent<Props, State> {
       stateUpdate.needToResetStyleCache = false;
       this.setState(stateUpdate);
     }
-    return null;
   }
 
   static _wrapSizeGetter(value: CellSize): CellSizeGetter {
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -459,7 +459,7 @@ class Grid extends React.PureComponent<Props, State> {
             : SCROLL_DIRECTION_BACKWARD
           : this.state.scrollDirectionVertical;
 
-      const newState: Object = {
+      const newState: $Shape<State> = {
         isScrolling: true,
         scrollDirectionHorizontal,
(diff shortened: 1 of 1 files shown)


source/Grid/Grid.js
```diff
@@ -597,29 +597,26 @@ class Grid extends React.PureComponent<Props, State> {
...
     if (!instanceProps.scrollbarSizeMeasured) {
       this.setState(prevState => {
-        prevState.instanceProps.scrollbarSize = getScrollbarSize();
-        prevState.instanceProps.scrollbarSizeMeasured = true;
-        prevState.needToResetStyleCache = false;
-        return prevState;
+        const stateUpdate = {...prevState, needToResetStyleCache: false};
+        stateUpdate.instanceProps.scrollbarSize = getScrollbarSize();
+        stateUpdate.instanceProps.scrollbarSizeMeasured = true;
+        return stateUpdate;
```
TrySound: Is it ok here to mutate state?

source/Grid/Grid.js
```diff
@@ -1433,19 +1430,17 @@ class Grid extends React.PureComponent<Props, State> {
...
+      prevState: this.state,
+      scrollLeft,
+      scrollTop,
     });
+
+    if (stateUpdate) {
+      stateUpdate.needToResetStyleCache = false;
+      this.setState(stateUpdate);
+    }
+    return null;
```
wuweiweiwu: Forgot about this. Will remove

source/Grid/Grid.js
wuweiweiwu: I think so. previously it was [this](https://github.com/bvaughn/react-virtualized/blob/2a7a82813f34bcce3f6d52036c5a96d538041687/source/Grid/Grid.js#L536:L539)

And since I need to set stuff on `instanceProps` I use prevState as a base

Issue #1028, opened 2018-02-28
Add option to not re-render list rows when scrolling


gpascale:
Hi,

I've been digging into performance of my react-virtualized components lately and am curious about the rendering behavior of List.

It seems that List re-renders every visible row (as well as overscan, presumably) every time it scrolls. This is really hurting my performance because my rows are fairly heavyweight and re-rendering them is non-trivial.

I understand that this is intentional because it allows for the customization of rows via the isScrolling parameter, as in the example code for List: https://github.com/bvaughn/react-virtualized/blob/master/docs/List.md. The idea being that you render a simplified row while scrolling and the heavyweight row once stopped. This is ok, but I really don't like that visual effect for my use case. Instead, **I propose that there should be a way to tell List not to continually re-render visible rows when scrolling.**

I made this quick pen to demonstrate how often the rows are being rendered. Of course the performance is fine in this simple example, but with more complex rows, scrolling performance starts to suffer.
https://codepen.io/gpascale/pen/rJoJgO

bvaughn:
That CodePen is a great example.

I've thought a little about maybe changing v10 to let you opt-in to the `isScrolling` param. Maybe we could do this in a backwards compatible way though (by letting you opt-out instead). If there was a prop that told `List` that `rowRenderer` didn't care about that param, then it could skip the extra render when scrolling stopped.

Would you be interested in putting together a PR for this (including docs+tests)?

wuweiweiwu:
@bvaughn I'd love to take this issue :)

bvaughn:
It's all yours

bfang711:
I met exactly the same issue in Grid as in List. Can you please add this enhancement into Grid as well?  React-infinite, another light weight virtualized library doing that, however they don't support other good functionalities that react-virtualized does. 

thank you and please let us know when it is implemented.

th3fallen:
was just tracking down some perf issues myself which led me here, thank you for working on this @wuweiweiwu

chitezh:
@wuweiweiwu any progress on this?

wuweiweiwu:
@bfang711 @chitezh @th3fallen Been busy with #1053 but I'll get a fix out for this as soon as I can :)

emflores:
@wuweiweiwu any update on this?

wuweiweiwu:
@emflores havent had time to work on it. I'll get it in as a patch to 9.19.x release

chitezh:
[...]

PR #1131, opened 2018-06-05
Grid: Allow users to opt out of rerendering on scroll stop

wuweiweiwu:
closes: #1028 

Added `isScrollingOptOut` that prevents re-rendering on scroll end. 

Cells always rerender on scroll end because the cell cache is cleared. Thus if this prop is specified, the visible cells in the cache are not cleared (to keep the cell cache to a manageable size) and thus can be used in `defaultCellRangeRender`.

- [x] add docs for `Grid`
- [x] add tests for `isScrollingOptOut`

### DIFF
--- a/source/Grid/Grid.jest.js
+++ b/source/Grid/Grid.jest.js
@@ -1905,6 +1905,52 @@ describe('Grid', () => {
       expect(cellRendererCalls.length).not.toEqual(0);
     });
 
+    it('should not clear cache if :isScrollingOptOut is true', () => {
+      const cellRendererCalls = [];
+      function cellRenderer({columnIndex, key, rowIndex, style}) {
+        cellRendererCalls.push({columnIndex, rowIndex});
+        return defaultCellRenderer({columnIndex, key, rowIndex, style});
+      }
+      const props = {
+        cellRenderer,
+        columnWidth: 100,
+        height: 40,
+        rowHeight: 20,
+        scrollTop: 0,
+        width: 100,
+        isScrollingOptOut: true,
+      };
+
+      render(getMarkup(props));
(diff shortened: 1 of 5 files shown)


source/Grid/Grid.jest.js
```diff
@@ -1905,6 +1905,52 @@ describe('Grid', () => {
       expect(cellRendererCalls.length).not.toEqual(0);
     });
 
+    it('should not clear cache if :isScrollingOptOut is true', () => {
+      const cellRendererCalls = [];
+      function cellRenderer({columnIndex, key, rowIndex, style}) {
```
aem: as more people start to push perf it might make sense to make some test utils to do things like track render calls rather than a bespoke function in each test. looks great for this issue, but maybe open an issue for that? it's something i may have some time to tackle next week

wuweiweiwu: Sounds good. I'll open an issue.

nihgwu:
the current approach is tricky, if we user custom cellRangeRenderer, we probably omit that opt out, I'd suggest we can simply pass `isScrolling = undefined` when calling cellRangeRenderer if `isScrollingOptOut` is `true`, BTW I prefer the term `usingIsScrolling` from `react-window`

wuweiweiwu:
@nihgwu Do you mind opening an issue for that? I like the idea of passing isScrolling = false when calling cellRangeRenderer 

and that propName makes more sense as well.

It'll be better tracking for me. Thanks!

nihgwu:
`isScrolling` to be `undefined` would be much better as `false` would lead wrong usage on user land

wuweiweiwu:
I see. I'd prefer something other than undefined though.

PR #1143, opened 2018-06-19
Remove pinnng scrollTop to 0

ofersadgat:
This is done so that if there are elements which appear above the current element, scrollOffset.top - this._positionFromTop will return a negative number thereby correctly telling the child that it is not being rendered. This allows the child to render only when it becomes necessary to do so as opposed to rendering always.

Can add a test later if you feel necessary. I first want to see if there is a reason that this shouldnt be done.

### DIFF
--- a/source/WindowScroller/WindowScroller.js
+++ b/source/WindowScroller/WindowScroller.js
@@ -91,9 +91,9 @@ export default class WindowScroller extends React.PureComponent<Props, State> {
 
   state = {
     ...getDimensions(this.props.scrollElement, this.props),
-    isScrolling: false,
     scrollLeft: 0,
     scrollTop: 0,
+    isScrolling: false,
   };
 
   updatePosition(scrollElement: ?Element = this.props.scrollElement) {
@@ -118,9 +118,38 @@ export default class WindowScroller extends React.PureComponent<Props, State> {
         width: dimensions.width,
       });
     }
+
+    const scrollProperties = this._getScrollProperties(scrollElement) || {};
+    if (
+      scrollProperties.scrollTop !== this.state.scrollTop ||
+      scrollProperties.scrollLeft !== this.state.scrollLeft
+    ) {
+      this.setState(scrollProperties);
+    }
+  }
+
+  _getScrollProperties(scrollElement) {
+    if (!this._isMounted) {
+      return;
+    }
+
+    const {onScroll} = this.props;
+
+    if (scrollElement) {
+      const scrollOffset = getScrollOffset(scrollElement);
+      const scrollLeft = scrollOffset.left - this._positionFromLeft;
+      const scrollTop = scrollOffset.top - this._positionFromTop;
+
+      return {
+        scrollLeft,
+        scrollTop,
+      };
+    }
   }
 
   componentDidMount() {
+    this._isMounted = true;
+
     const scrollElement = this.props.scrollElement;
 
     this._detectElementResize = createDetectElementResize();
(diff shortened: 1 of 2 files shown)


aem:
just got around to investigating this a bit. internally at my company we take advantage of the scroll pinning (do some subtraction, pass in the result, and if it's <= 0 then scroll to top) and i imagine we're not alone. i think a better approach would be to do this client-side, conditionally rendering the components based on screen position. seems like rendering/not rendering the element is a bit out of scope

aem:
@wuweiweiwu would be interested to hear your thoughts

ofersadgat:
@aem I confused, you say if the result is less than or equal to zero, so allowing negative numbers would still be handled correctly in your case, right?

wuweiweiwu:
I think its reasonable for the render prop to pass in a negative value so that children can know that its above the virtualization window and perhaps return some sort of placeholder. 

I'll have too test this locally . @ofersadgat do you mind adding some tests? Theres a good chance this would be merged

ofersadgat:
Yup, can do. I'm on vacation right now, so I'll submit it first thing when I get back (about 2 weeks).

ofersadgat:
@wuweiweiwu to give you a heads up, I started taking a look at this again and found a few bugs: scrollTop is not updated if the position of the element updates (but because the position of the elements updates the scrollTop should update as well) so this will be a slightly larger change than anticipated. Please let me know if you see any problems with updating scrollTop in updatePosition. Thanks.

wuweiweiwu:
@ofersadgat No problem! I will have to look at the specific changes in `updatePosition` but I think it should be fine

source/WindowScroller/WindowScroller.js
```diff
@@ -91,9 +91,9 @@ export default class WindowScroller extends React.PureComponent<Props, State> {
 
   state = {
     ...getDimensions(this.props.scrollElement, this.props),
-    isScrolling: false,
     scrollLeft: 0,
     scrollTop: 0,
+    isScrolling: false,
```
ofersadgat: will move this back

ofersadgat:
I want to write a test along the lines of:

```js
  it('should update scrollTop when a resize event occurs', () => {
    const renderFn = jest.fn();

    // Simulate scrolled documentElement
    const component = render(
      getMarkup({
        headerElements: <div id="a" style={{height: 501}} />,
        renderFn,
      }),
    );
    render._mountNode.querySelector('#a').size.height = '200px';
    expect(renderFn).lastCalledWith(
      expect.objectContaining({
        scrollTop: -200,
      }),
    );
  });
```
I.e. if a node above the current one changes size, that will cause a resize on the scollelement which should cause new props to get passed down. ut I'm not quite sure how to get this test to work with JSDom. Any help here would be appreciated.

Thanks!

wuweiweiwu:
@ofersadgat hmm. if an element changes size, it will not trigger a rerender unless `forceUpdate` is called since these are all `PureComponents` 

This will take a bit more work then I initially thought. I'm not sure if this will introduce new bugs. 

Right now the changes look/sound fine, but we'll have to do some extensive smoke tests to make sure it doesnt introduce new bugs.

ofersadgat:
@wuweiweiwu I'm confused. If the parent element changes size, then the size of the scrollElement will change which means that `_registerResizeListener` will call `_onResize` which will call `updatePosition` which will call setState with the new parameters. If the parameters have changed, we should get a rerender which will pass the new props down. Or have I missed something?

PR #1166, opened 2018-07-10
Expose container ref in Grid

jtmcgrath:
Hey!

I need to be able to imperatively call `ref.scroll({ left: n, behavior: 'smooth' })` on the container of a virtualized grid based on users clicking a button - at present the lib doesn't have any way to expose that ref, think the simple approach here should work? Is this something you'd support?

Not sure if this really counts as a feature which needs testing or documentation, so I've left those for now.

- [ ] For any new features or bug fixes, both positive and negative test cases have been added

### DIFF
--- a/source/Grid/Grid.js
+++ b/source/Grid/Grid.js
@@ -97,6 +97,9 @@ type Props = {
   /** Unfiltered props for the Grid container. */
   containerProps?: Object,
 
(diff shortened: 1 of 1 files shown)


jtmcgrath:
Any chance someone could take a look at this? Happy to make any changes which might be requested!

PR #1175, opened 2018-07-17
Allow users to override default table row styles

### LINKED ISSUE #1171: Allow rowStyle on tables to override default styles?
I have a table where I would like to have some content overflow from a row. However, passing a `rowStyle` object with `overflow: 'visible'` can't do it, because the `Table` component merges the `rowStyle` object with some other styles (including `overflow: 'hidden'`) [here](https://github.com/bvaughn/react-virtualized/blob/de107e81f6d412d3561b169f7fdbfdbe25dd7b00/source/Table/Table.js#L395).

We use css in js everywhere in the project, so using a className to override this would be a little inconsistent and confusing.

So I was wondering what the reasoning behind forcing default some styles is?

I would suggest changing this
```
style: {
    ...rowStyleObject,
    height: headerHeight,
    overflow: 'hidden',
    paddingRight: scrollbarWidth,
    width: width,
}
```

To this
```
style: {
    height: headerHeight,
    overflow: 'hidden',
    paddingRight: scrollbarWidth,
    width: width,
    ...rowStyleObject,
}
```

To give users the option to override the styles if they want, while still keeping the defaults if they don't override them.

If you guys agree to this change, I'm more than happy to submit a PR for it!

emroussel:
This allows users to override default row styles on tables with a style object.

Since the object was spread before the defaults, there was no way to override them, and users had to use a class.

Fixes  #1171 

I looked if this change should be made in other places, and didn't see anything, but please let me know if I missed something!

### DIFF
--- a/source/Table/Table.js
+++ b/source/Table/Table.js
@@ -392,11 +392,11 @@ export default class Table extends React.PureComponent {
             className: cn('ReactVirtualized__Table__headerRow', rowClass),
             columns: this._getHeaderColumns(),
             style: {
-              ...rowStyleObject,
               height: headerHeight,
               overflow: 'hidden',
               paddingRight: scrollbarWidth,
               width: width,
+              ...rowStyleObject,
             },
           })}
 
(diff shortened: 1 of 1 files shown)


emroussel:
@wuweiweiwu any chance you get this merged and released soon?

wuweiweiwu:
I'll push out a minor release soon! @emroussel

emroussel:
Thanks!

Issue #1173, opened 2018-07-16
Using the MultiGrid isScrolling is only true on the grid that scrolls


mauricedb:
### What is the current behavior?

Using the MultiGrid cellRenderer the isScrolling argument is only true on the grid that scrolls, the other grids will always report isScrolling as false.

The downside here is that optimizations to code based on the isScrolling argument can't be done. In my specific case, I get quite a bit of lag with the row headers dragging slower due to drag&drop, context menu and swipe logic in there. As none of that can actually happen when scrolling I want to use the isScrolling argument to render a simpler row header.

See https://codesandbox.io/s/z6921736wx for a working example. Scroll using the mouse wheel or slider bar and observe that only the grid that is scrolled updates the status in the cell.

### What is the expected behavior?

I expect the isScrolling argument to be correct in each of the four connected Grid instances.

### Which versions of React and react-virtualized, and which browser / OS are affected by this issue? Did this work in previous versions of react-virtualized?
|                   |          |
|-------------------|----------|
| Browser           | Chrome/FireFox/Edge |
| OS                | Windóws 10.1803 |
| React             | 16.4.1   |
| React DOM         | 16.4.1   |
| react-virtualized | 9.20.1   |

zag2art:
it's still there. "version": "9.21.0"

chnakamura:
I am still experiencing this behavior on version `9.22.3`, but I was able to ameliorate it pretty easily by storing the value `dataGridIsScrolling` in state and forcing one last update after we are done scrolling.

```
const [dataGridIsScrolling, setDataGridIsScrolling] = useState<boolean>(false);

useEffect(() => {
// This allows us to trigger rendering the tooltips once we have stopped scrolling. 
// We do this because rendering the tooltips while scrolling causes the component to lag,
// and the tooltips are not useful while the grid is scrolling.
if (multiGridRef !== null && dataGridIsScrolling == false) {
	multiGridRef.forceUpdateGrids();
}
}, [dataGridIsScrolling])
```

PR #1201, opened 2018-08-27
Set the isScrolling prop in cellRenderer for the other child grids

mauricedb:
Fixes: #1173 

- [x] ~For any new features, documentation has been added~
- [x] ~For any documentation changes, the text has been proofread and is clear to both experienced users and beginners.~

### DIFF
--- a/source/MultiGrid/MultiGrid.js
+++ b/source/MultiGrid/MultiGrid.js
@@ -228,7 +228,7 @@ class MultiGrid extends React.PureComponent {
 
     // scrollTop and scrollLeft props are explicitly filtered out and ignored
 
-    const {scrollLeft, scrollTop} = this.state;
+    const {scrollLeft, scrollTop, isScrolling} = this.state;
 
     return (
       <div style={this._containerOuterStyle}>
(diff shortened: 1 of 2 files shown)


source/MultiGrid/MultiGrid.js
```diff
@@ -238,13 +238,15 @@ class MultiGrid extends React.PureComponent {
             ...rest,
             onScroll,
             scrollLeft,
+            isScrolling: isScrolling || undefined,
```
wuweiweiwu: Can this be just `isScrolling`?

mauricedb: @wuweiweiwu,
Good question.

It can but that causes the test **should call forceUpdate() on inner Grids** to fail because there is a 5th `cellRenderer` call. I could change the test but doing an extra render would cause a bit of overhead so I choose for this solution.

wuweiweiwu: hm interesting. I would think that setting isScrolling to false should have the same effect as setting it to undefined

mauricedb: Same for me but my experience was different. No harm in double checking, maybe I did something wrong, as it takes only a minimal effort.

zag2art:
What are we waiting for? The PR looks fine, isn't it?

zag2art:
Just took a look at the PR, it doesn't turn off isScrolling for leftBottom and rightTop tables.

PR #1206, opened 2018-09-04
Fix #1004 Update MultiGrid width and height when cell width or height is changed by getter function

### LINKED ISSUE #1004: MultiGrid does not update right grid when fixed column size changes

roncrush:
Based on an existing pull request for this same issue (https://github.com/bvaughn/react-virtualized/pull/1007), I made the requested changes since the original pull request has not been updated in a long time.

Essentially this fixes the issue in MultiGrid where if you had a fixed column/row and the width/height of a cell within the fixed column/row changed, it would not update the grid to reflect the new width/height.

There are 2 tests to cover the height and width scenarios (the tests are from the first pull request).

### DIFF
--- a/source/MultiGrid/MultiGrid.jest.js
+++ b/source/MultiGrid/MultiGrid.jest.js
@@ -309,8 +309,8 @@ describe('MultiGrid', () => {
       expect(bottomLeftAfter.style.getPropertyValue('width')).toEqual('75px');
       expect(bottomRightAfter.style.getPropertyValue('width')).toEqual('325px');
     });
-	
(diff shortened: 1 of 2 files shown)


PythEch:
Sad to see that this PR wasn't merged, I had the exact same issue.

My workaround is the following:
```javascript
this.setState({ ... }, () => this.multiGridRef && this.multiGridRef.recomputeGridSize())
```

PR #1207, opened 2018-09-04
Add onColumnClick to Table

grahamlyus:
This change adds `onColumnClick` to `Table`, works the same way as `onHeaderClick`.

I was attaching `onClick` in my `Column` `cellRenderer`, but it doesn't work when the cell is empty. Attaching to the `div` in `_createColumn` works regardless.

### DIFF
--- a/source/Table/Table.js
+++ b/source/Table/Table.js
@@ -85,6 +85,12 @@ export default class Table extends React.PureComponent {
     noRowsRenderer: PropTypes.func,
 
     /**
+     * Optional callback when a column is clicked.
+     * ({ columnData: any, dataKey: string }): void
+     */
+    onColumnClick: PropTypes.func,
+
+    /**
      * Optional callback when a column's header is clicked.
      * ({ columnData: any, dataKey: string }): void
      */
@@ -426,6 +432,7 @@ export default class Table extends React.PureComponent {
   }
 
   _createColumn({column, columnIndex, isScrolling, parent, rowData, rowIndex}) {
+    const {onColumnClick} = this.props;
     const {
       cellDataGetter,
       cellRenderer,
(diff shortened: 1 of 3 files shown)


scy0846:
I definitely could use this. There are some columns I need the ability to click and others I need to disable as they contain buttons in them

source/Table/Table.js
```diff
@@ -447,6 +454,10 @@ export default class Table extends React.PureComponent {
       rowIndex,
     });
 
+    const onClick = event => {
+      onColumnClick && onColumnClick({columnData, dataKey, event});
```
wuweiweiwu: I don't think its necessary to pass in the `event`. We should just invoke the callback with columnData and datakey

wuweiweiwu:
Can you also add unit tests?  To make sure that column clicks invoke the callback with the right parameters?

source/Table/Table.js
grahamlyus: Indeed, I am not using `event` in my handlers, but I added it to make it the same callback parameters as `onHeaderClick`

grahamlyus:
Tests are in `source/Table/Table.jest.js`, also based on the `onHeaderClick` tests

eubrunomiguel:
columnData seems to always be undefined

christianbalderrama:
Is there any implementations of this `onColumnClick` function as well as `onHeaderClick`? I can't find any implementations/example of these features in the documentation.

christianbalderrama:
Wanted to follow up as well, `columnData` is undefined.

wuweiweiwu:
@christianbalderrama  @eubrunomiguel  can you open an issue? Thanks!

B3Kay:
`columnData` is undefined...

Issue #1203, opened 2018-08-28
e.target.className.indexOf is not a function


dmydry:
Unfortunately I can't make a working repro. This is a main reason of this issue.

In our production code we often get client errors with the next traceback: 
`TypeError: e.target.className.indexOf is not a function`.

Apparently this is the next code: https://github.com/bvaughn/react-virtualized/blob/master/source/vendor/detectElementResize.js#L73

Browser occurrences are quite different (last chrome is in top). Hence it's not a browser issue.

We use WindowScroller, AutoSizer and List components with quite standart props and behaviour.
```
        <WindowScroller>
          {({ height, isScrolling, scrollTop }) => (
            <AutoSizer disableHeight>
              {({ width }) => (
                <List
                  isScrolling={isScrolling}
                  overscanRowCount={3}
                  scrollTop={scrollTop}
                  autoHeight
                  height={height}
                  rowCount={listCount}
                  rowHeight={rowHeight}
                  rowRenderer={this.rowRenderer}
                  width={width}
                />
              )}
            </AutoSizer>
          )}
        </WindowScroller>
```

Could anyone give me a tip, what is a possible reason of this issue?

|                   |          |
|-------------------|----------|
| React             |     16.4.1     |
| React DOM         |    16.4.1      |
| react-virtualized |     9.20.1     |

wuweiweiwu:
I'm guessing that className is returning a `null` or `undefined`. 

A fix would be to check that `e.target` and `e.target.className` exists

dmydry:
@wuweiweiwu thats right and I guess I will do a fork after a while. I just prefer to use the original package source if possible.
Just asked if anyone got the same issue. And if so, maybe we just should add an extra checks for a `e.target.className` and `typeof e.target.className.indexOf === 'function'`. This should broke nothing but fix the issue.

wuweiweiwu:
@dmydry would you be interested in submitting a pull request for this?

dmydry:
@wuweiweiwu done ;)

I didn't find any tests for the `detectElementResize` listeners. 
Will be happy to dive deeper into the project and cover if possible.

wuweiweiwu:
@dmydry Awesome! thanks for contributing!

PR #1210, opened 2018-09-06
Add an extra check for an e.target.className.indexOf function

dmydry:
These 2 lines of code can fix a possible reason of the issue [1203](https://github.com/bvaughn/react-virtualized/issues/1203)


wuweiweiwu:
closes #1203

PR #1260, opened 2018-11-06
Fix #1159: CollectionView scrollTop/scrollLeft

### LINKED ISSUE #1159: Setting scrollTop/scrollLeft on a Collection does not scroll in 9.19.0 or later.
Example:  https://codesandbox.io/s/9j5v53ryq4?module=%2FExample.js

React-virtualized versions 9.19.0 and later (9.19.1, 9.20.0) do not apply the scrollTop/scrollLeft values from ScrollSync to a Collection component properly. The end result is that the items to be drawn _are_ updated to match what should be in the view window, but the _view window_ is not scrolled to show those items. So what is visible is an unmoving Collection that slowly has the upper rows go invisible until there are no more rows being drawn in the view window instead of the Collection's view window scrolling to match.

Reverting to react-virtualized 9.18.5, the above example works as expected. The lower Collection does scroll in response to changes to the scrollTop value. You can test this by changing the dependency version for react-virtualized in the above sandbox without making any other changes.

dawnmist:
Props scrollTop/scrollLeft are only applied when this.state.scrollPositionChangeReason is REQUESTED, but getDerivedStateFromProps was not setting scrollPositionChangeReason. This meant that these props were not being applied to the view.

- [x] For any new features, documentation has been added - N/A
- [x] For any documentation changes, the text has been proofread and is clear to both experienced users and beginners - N/A

closes #1159

### DIFF
--- a/source/Collection/Collection.jest.js
+++ b/source/Collection/Collection.jest.js
@@ -457,7 +457,7 @@ describe('Collection', () => {
   describe(':scrollLeft and :scrollTop properties', () => {
     it('should render correctly when an initial :scrollLeft and :scrollTop properties are specified', () => {
       let indices;
-      render(
+      const collection = render(
         getMarkup({
           onSectionRendered: params => {
             indices = params.indices;
@@ -467,6 +467,9 @@ describe('Collection', () => {
         }),
       );
       compareArrays(indices, [3, 4, 5, 7, 8, 9]);
+      expect(
+        collection._collectionView.state.scrollPositionChangeReason,
+      ).toEqual('requested');
     });
 
     it('should render correctly when :scrollLeft and :scrollTop properties are updated', () => {
@@ -479,7 +482,7 @@ describe('Collection', () => {
         }),
       );
       compareArrays(indices, [0, 1, 2, 3]);
-      render(
+      const collection = render(
         getMarkup({
           onSectionRendered: params => {
             indices = params.indices;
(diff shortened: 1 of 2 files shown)


rstoneIDBS:
Has this PR slipped through the cracks? We would love to be able to switch to the latest version, but without this fix we either have to patch or stick on 9.18.5

dawnmist:
As far as I know, no-one has ever reviewed it - I've certainly never received any form of feedback ever since the initial labelling of the bug 2 years ago.

The only reason I can think of for the issue was that at the time of creating the pull request I hadn't been able to get puppeteer working, hence couldn't run the tests properly locally, and the count of checkboxes on the pull requests list page therefore wasn't "correct".

I've updated the pull request to current master, and updated the tests that were _incorrectly_ passing on master (because they were not checking the `scrollPositionChangeReason` from state that is _required_ for the scrollTop and scrollLeft values to be actually _used_). These tests now fail on master, but are passing with the pull request.

It is my hope that after 18 months waiting for a review (or almost 2 years waiting for _any_ action on the bug report) that the test changes will finally trigger this into being looked at. It feels really awful to try to contribute and simply be ignored entirely despite a critical bug having been introduced that even with a fully worked example doesn't get any action and a 2 line pull request fixing it doesn't get any feedback as to why it is being left untouched/unreviewed.

We're still stuck on react-virtualized 9.18.5 because the implementation of `getDerivedStateFromProps` for the CollectionView component has been fundementally _broken_ ever since it was written.

sid-ag5:
We have also been stuck on 9.18.5 for a while now because of this issue, and we want to be prepared for concurrent mode when it comes along.

@wuweiweiwu Can this be looked into? It's a very blocking issue and seems to be going unnoticed.

wuweiweiwu:
Will release and update!

wuweiweiwu:
released! https://www.npmjs.com/package/react-virtualized/v/9.22.2 hope the update process is smooth

PR #1312, opened 2019-01-27
Use sparse array for cell position caches

trxcllnt:
This PR swaps out the hash based cell position caches and the underlying binary search with a constant-time [sparse array](https://github.com/trxcllnt/linear-layout-vector/blob/master/index.js) linear layout vector implementation.

The vector stores element sizes in a single dimension. Sizes are stored in blocks with power-of-two length. The vector supports constant-time `(index) => position` lookup, update, insertion, and removal by shifting the index to find the block index, then a mask (mod) to find the cell index.

Reverse lookup `(position) => index` is currently linear with respect to the number of blocks/size of each block. Flamegraphs indicate that on a table with 1MM rows and with the default block_size of 128, this amounts to about 0.15-1ms (150-1000 microseconds) per call when scrolling at the end of the list. If avoiding linear scans is an issue on principle, it would be possible to cache and invalidate the block sizes' prefix sum, and implement this lookup as a binary search again. The block size can also be intelligently tuned based on total element count for a classic space/time trade-off.

This change means the following methods are now `O(1)`:
```
getSizeAndPositionOfCell
getSizeAndPositionOfLastMeasuredCell
getUpdatedOffsetForIndex
```

`getVisibleCellRange` is `O(n)`, where `n=num_blocks`

The biggest practical benefit is that cell dimensions are stable across invalidation. After initial measurement the scrollbars won't jump around, as you can see in this screen capture I uploaded to youtube. The first half is rendering a MultiGrid with `react-virtualized@9.21.0`, and the second half is rendering the same MultiGrid with the current PR. You can find the source for this demo [here](https://github.com/trxcllnt/arrow-viewer/blob/484acdbf56f084e9a1fc24f86743cd7a3d919ffc/src/App.tsx#L12).

[react virtualized scrolling](http://www.youtube.com/watch?v=8hzP8AZMqp4 "react virtualized scrolling")

- [ ] For any new features or bug fixes, both positive and negative test cases have been added

### DIFF
--- a/package.json
+++ b/package.json
@@ -171,6 +171,7 @@
     "babel-runtime": "^6.26.0",
     "clsx": "^1.0.1",
     "dom-helpers": "^2.4.0 || ^3.0.0",
+    "linear-layout-vector": "0.0.1",
     "loose-envify": "^1.3.0",
     "prop-types": "^15.6.0",
     "react-lifecycles-compat": "^3.0.4"
--- a/source/Grid/utils/CellSizeAndPositionManager.js
+++ b/source/Grid/utils/CellSizeAndPositionManager.js
@@ -1,5 +1,6 @@
 /** @flow */
 
+import LinearLayoutVector from 'linear-layout-vector';
 import type {Alignment, CellSizeGetter, VisibleCellRange} from '../types';
 
 type CellSizeAndPositionManagerParams = {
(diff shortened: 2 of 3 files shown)


wuweiweiwu:
@trxcllnt What is the size impact of this dependency?

trxcllnt:
2.2kb minified + gzipped, though that could be even smaller if cleaned up and converted from ES5 to an ES6 class. The version up on npm is a few years old, and that was converted to ES5 from ActionScript a few years before that.

omerts:
@trxcllnt This PR really caught my attention, and is an interesting concept.. How is `getSizeAndPositionOfCell` `O(1)` if there is still a for loop iterating over the cells and calculating their size?

Issue #999, opened 2018-02-03
Make grid uncontrollable in favor of ref methods


TrySound:
According this thread

> rename eg `scrollToColumn` and `scrollToRow` to `defaultScrollToColumn` and `defaultScrollToRow` and then make people use the `scrollToCell` method to update those values after mount.

https://twitter.com/brian_d_vaughn/status/959836416103493632

bvaughn:
Let's add this to the version 10 umbrella (#658) 😄 I think it's worth doing

somi92:
Hi guys,
	
First of all thank you for great work on RV.
	
I have a few questions regarding this issue.
	
We are using RV on the project I'm working on and it has been a great experience so far. However, we are having issues since React 16.4 update, probably due to changes in the way getDerivedStateFromProps gets called coupled with the mirroring of some scroll props in state (as mentioned in the above referenced Twitter thread). From what I can see, this is the root cause of some really annoying bugs like these https://github.com/bvaughn/react-virtualized/issues/1170, https://github.com/bvaughn/react-virtualized/issues/1179 and making the grid uncontrollable should fix it.
	
@bvaughn I see that you use and promote this approach in your new library https://github.com/bvaughn/react-window/issues/86, so I've made a small demo to try it out on Grid and MultiGrid in RV (it fixes the issue https://github.com/bvaughn/react-virtualized/issues/1170 and enables me to use keyboard navigation on latest React version). 
Here is the change on my fork https://github.com/somi92/react-virtualized/commit/0d838bdc8aa37ddf6908e42386f5a38699247994. In the example, the Grid is made uncontrollable in regards to scrollTop/scrollLeft and MultiGrid relies on ref methods for syncing (for the sake of demo only Grid and MultiGrid are changed, so things like ScrollSync won't work).
	
Is this the right way to go and would something like this be considered for the next release? 
I am aware this is a breaking change and it wouldn't be merged soon, so is there any info on the release date of next major version?
	
We would appreciate your feedback and would be happy to help in fixing this.

bvaughn:
Not sure if @wuweiweiwu has anything actively planned for a 10.x release, or if this would fit into that at all. (I'm totally uninvolved with this library at this point.)

wuweiweiwu:
@somi92 Nothing is currently being actively planned. But I will take a look at your fork. And perhaps will start work on v10 when I have some time. :)

PR #1320, opened 2019-02-05
Make scroll offset uncontrollable in favor of ref methods

somi92:
Changes based on discussion from https://github.com/bvaughn/react-virtualized/issues/999.

Summary of changes:

* Grid component
  - Make Grid uncontrollable, rename scrollTop/scrollLeft props defaultScrollTop/defaultScrollLeft (only to be used on mount)
  - Limit scrollTop and scrollLeft to maximum possible values state update in scrollToPosition method(prevents issues when calling scrollToPosition multiple times with larger than max)
  - Update tests and docs

* MultiGrid component
  	- Remove scrollLeft and scrollTop from state
  	- Remove getDerivedStateFromProps and state handling code componentDidMount
	- Update all _onScroll methods to use scrollToPosition for syncing
	- Add scrollToPosition method (pass-thru that calls of the main scrolling grid which is _bottomRightGrid)
	- Update tests and docs

* List component
  - Make List uncontrollable, rename scrollTop to (only to be used on mount)
  - Update tests and docs

* Table component
  - Make Table uncontrollable, rename scrollTop to defaultScrollTop (only to be used on mount)
  - Update tests and docs

* ScrollSync component
  	- Add flow types
  	- Add registerChild to children render function parameters
  	- Implement _registerChild to store refs on scrollable children
  	- Call scrollToPosition on children refs in _onScroll handler sync them
  	- Update tests, docs and example

* WindowScroller component
  	- Update flow types
  	- Rename existing registerChild method to registerChildContainer
  	- Add registerChild to children render function parameters
  	- Implement _registerChild to store ref on scrollable child
  	- Call scrollToPosition on child ref in _handleWindowScrollEvent
  	  handler
  	- Update tests, docs and example


wuweiweiwu:
@somi92 Thanks for doing this! I'll take some time soon to thoroughly review this and test.

source/ScrollSync/ScrollSync.js
```diff
@@ -1,20 +1,50 @@
...
-     */
-    children: PropTypes.func.isRequired,
-  };
+export default class ScrollSync extends React.PureComponent<Props, State> {
+  _elements: (ScrollableElement | null)[] = [];
+  _onScroll: (params: OnScrollParams) => void;
+  _registerChild: (element: ScrollableElement | null) => void;
 
-  constructor(props, context) {
+  constructor(props: Props, context: any) {
```
wuweiweiwu: nit: can we get the contextTypes

source/ScrollSync/ScrollSync.js
```diff
@@ -67,5 +101,18 @@ export default class ScrollSync extends React.PureComponent {
...
       scrollWidth,
     });
+    this._elements.forEach(element => {
+      if (element != null && element.scrollToPosition)
+        element.scrollToPosition({scrollLeft, scrollTop});
+    });
+  }
+
+  _registerChild(element: ScrollableElement | null) {
+    if (element == null || !element.scrollToPosition) {
```
wuweiweiwu: nit: i'd even just check `!element`

source/ScrollSync/ScrollSync.js
```diff
@@ -67,5 +101,18 @@ export default class ScrollSync extends React.PureComponent {
...
+      console.warn(
+        "ScrollSync registerChild expects to be passed a component with 'scrollToPosition' function",
+      );
+    }
+    this._elements.push(element);
```
wuweiweiwu: Should we register if its falsey?

somi92:
@wuweiweiwu Thanks! I'm a bit short on time for a couple of days but will respond to your remarks by Tuesday.

wuweiweiwu:
@somi92 No problem! Take your time :) 

Your effort on this has been very much appreciated

source/ScrollSync/ScrollSync.js
somi92: You're right, probably not, there is no much sense in keeping the element in that case. I'll fix this!

source/ScrollSync/ScrollSync.js
somi92: Will fix! Should we consider keeping the part with scrollToPosition check since that's the thing we actually need from the element:

```suggestion
    if (!element || !element.scrollToPosition) {
```

source/ScrollSync/ScrollSync.js
somi92: Yes, I would like to, but I'm struggling with this one a bit since I don't have much experience with flow and legacy context 😅 I'm not sure what should I exactly do, this context here is just passed to the parent.

PR #1343, opened 2019-03-06
Mitigate memory and performance issues for very large lists. 

omerts:
Current implementation of `CellSizeAndPositionManager` is prune to out of memory exceptions and laggy behavior when dealing with very large lists. Specifically this is due to the fact that many cells sizes are calculated, and stored in cache, causing a memory bloat, and slower performance. This pull request mitigates the issue in cases where the CellSize is a numeric value, so at least there is a way of working with very large lists. It does so, by arithmetically calculating offsets, size, and position. As the size is static for every cell there is no need to iterate over other cells to figure out these values. 

It can be easily reproduced in the demo just by raising the List's number of rows to 40 million items, and pulling the scroll down.

Before the fix (pull scroller down was also laggish):
image

After fix (scroller was very responsive):
image

### DIFF
--- a/source/Grid/utils/CellSizeAndPositionManager.js
+++ b/source/Grid/utils/CellSizeAndPositionManager.js
@@ -1,17 +1,17 @@
 /** @flow */
 
-import type {Alignment, CellSizeGetter, VisibleCellRange} from '../types';
+import type {Alignment, CellSize, VisibleCellRange} from '../types';
 
 type CellSizeAndPositionManagerParams = {
   cellCount: number,
-  cellSizeGetter: CellSizeGetter,
+  cellSize: CellSize,
   estimatedCellSize: number,
 };
 
 type ConfigureParams = {
   cellCount: number,
   estimatedCellSize: number,
-  cellSizeGetter: CellSizeGetter,
+  cellSize: CellSize,
 };
 
 type GetUpdatedOffsetForIndex = {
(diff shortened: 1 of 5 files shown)


eladhayun:
👍 need this ASAP, hope will be merged soon

source/Grid/utils/CellSizeAndPositionManager.js
```diff
@@ -92,10 +92,6 @@ export default class CellSizeAndPositionManager {
     return 0;
   }
 
-  _isCellSizeNumeric(): number {
-    return typeof this._cellSize === 'number';
-  }
```
tomasro27: I don't feel like you would need to remove this. Is there any specific reason?

source/Grid/utils/CellSizeAndPositionManager.js
```diff
@@ -139,6 +164,20 @@ export default class CellSizeAndPositionManager {
     return this._cellSizeAndPositionData[index];
   }
 
+  /**
+   * This method returns the size and position for the cell at the specified index.
+   * It decides on the correct calculation method according to the type of CelSize.
```
tomasro27: typo: CellSize

tomasro27:
Does this PR fixes the issue when there are columns with different width, or would it only work in the case that cells have a fixed height/width?

Also, this is optimizing the calculation of the cell sizes, but it doens't change the way caching is done, correct?

For very large grids, dragging the scrollbar makes the grid crash out of memory. However, doing different scrolling events with the mouse doesn't cause this, because the cache is cleared after every scroll event. I was wondering if it is worth adding more logic to the cell and style caches to avoid this, when scrolling through millions of rows.

source/Grid/utils/CellSizeAndPositionManager.js
omerts: I think it was due to flow type error

omerts:
@tomasro27 Sorry for the late response. 

> Does this PR fixes the issue when there are columns with different width, or would it only work in the case that cells have a fixed height/width?
It only fixes the issue for fixed width/height. That's why I only called it "mitigate" the issue.

> Also, this is optimizing the calculation of the cell sizes, but it doens't change the way caching is done, correct?
Correct, since it does all calculation with arithmetic operation there is no need for a cache. 

> For very large grids, dragging the scrollbar makes the grid crash out of memory. However, doing different scrolling events with the mouse doesn't cause this, because the cache is cleared after every scroll event. I was wondering if it is worth adding more logic to the cell and style caches to avoid this, when scrolling through millions of rows.
IMHO if cache is cleared on every scroll event this is even a greater reason to add this. If you keep on scrolling down, and cache is invalidated, the component will have to recalculate all the cells up to the current cell every time. This will fix this issue for fixed size rows/columns.

hieuhlc:
Hi @omerts @tomasro27 can I ask the progress of this issue? It seems very promising...

kaniseo:
Can we get an update on this? I'm Dealing with 200 million rows and it would be awesome to move this forward.

arsnyder16:
@omerts @wuweiweiwu  @tomasro27  @bvaughn  Where does this stand? What needs to be done to move this forward?

 We experiencing some performance issues that this is going to address

Issue #807, opened 2017-09-08
Rows rendered outside of view


nifgraup:
**Steps to reproduce**
- visit https://bvaughn.github.io/react-virtualized/#/components/List
- set "Num rows" to 100000
- scroll top bottom
- click "Use dynamic row heights?"

**What happens**
The list is just white. Everything is fine when "Num rows" is set to 10000. The DOM nodes are positioned outside of the list. It's possible to bypass this issue by making DEFAULT_MAX_SCROLL_SIZE very large.

bvaughn:
At a glance, I think this maybe indicates a bug with cached styles and scaling. RV handles resized rows properly in non-scaling case (eg 1000 rows) but needs a nudge to properly reset itself when things change from non-scaling to scaling (or maybe when scaling changes from one scale to another, note sure).

FWIW when I say "scaling" I'm referring to [this](https://bvaughn.github.io/forward-js-2017/#/24/5) which is done by [this](https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/utils/ScalingCellSizeAndPositionManager.js).

avranju:
I am seeing a similar behavior with a table component inside an `AutoSizer`. A new row gets dynamically added about twice a second and after about 16000 rows I see the screen going blank with every update. If I scroll up or down a bit it comes back and then blanks out when the next row is added.

nifgraup:
I can not reproduce anymore.

nifgraup:
Still an issue, I didn't test properly when closing.

nifgraup:
This used to work, but detection of Chromium based browsers broke when chrome.webstore was removed around v71. Submitted pull request: https://github.com/bvaughn/react-virtualized/pull/1355

PR #1355, opened 2019-04-08
Fix detection of Chromium based browsers in getMaxElementSize

nifgraup:
Fixes reintroduction of #807.


wuweiweiwu:
@nifgraup  do you mind explaining why this fixes #807 I dont think i fully understand

nifgraup:
@wuweiweiwu #807 was fixed in an unknown commit, and verified fixed on Mar 26 2018, https://github.com/bvaughn/react-virtualized/issues/807#issuecomment-376132305. The issue is Chromium specific and the fix depends on detection with the isChrome() function. That function stopped working in Chromium 71 and this pull request fixes that.

wuweiweiwu:
ok @nifgraup thanks for clarifying

PR #1380, opened 2019-05-28
fix(Table): aria-readonly is not allowed on rowgroup

### LINKED ISSUE #1313: Table has invalid aria-readonly attribute
You can see this in the [current demo for Table](https://bvaughn.github.io/react-virtualized/#/components/Table) and navigating to the `ReactVirtualized__Table__Grid` in devtools.

### What is the current behavior?

The `ReactVirtualized__Table__Grid` has attribute `aria-readonly` set to `true` by default.

### What is the expected behavior?

There should not be an `aria-readonly` attribute set on an element with role: `rowgroup`. See [ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/#rowgroup). It doesn't break app behavior but in violating a standard it triggers errors in some accessibility testing tools.

eps1lon:
Closes #1313

Didn't find any existing test for aria-readonly so I'm not sure if you even want tests for this.

### DIFF
--- a/source/Table/Table.js
+++ b/source/Table/Table.js
@@ -416,6 +416,7 @@ export default class Table extends React.PureComponent {
 
         <Grid
           {...this.props}
+          aria-readonly={null}
           autoContainerWidth
           className={clsx('ReactVirtualized__Table__Grid', gridClassName)}
           cellRenderer={this._createRow}


source/Table/Table.js
```diff
@@ -416,6 +416,7 @@ export default class Table extends React.PureComponent {
 
         <Grid
           {...this.props}
+          aria-readonly={null}
```
eps1lon: Not allowed on [rowgroup](https://www.w3.org/TR/wai-aria-1.1/#rowgroup): https://www.w3.org/TR/html-aria/#allowed-aria-roles-states-and-properties

PR #1399, opened 2019-07-18
Table: Add `key` to RowRendererParams type definition

kgregory:
`defaultRowRenderer` accepts a type of `RowRendererParams` and destructures `key`, which isn't present in the type definition.  This pull request updates the type definition and the `Table` documentation for `rowRenderer`.

### DIFF
--- a/docs/Table.md
+++ b/docs/Table.md
@@ -149,6 +149,7 @@ This function accepts the following named parameters:
 |:---|:---|
 | className | Row-level class name |
 | columns | Array of React nodes |
+| key | Unique key within array of rendered rows |
 | index | Row index |
 | isScrolling | Boolean flag indicating if `Table` is currently being scrolled |
 | onRowClick | Optional row `onClick` handler |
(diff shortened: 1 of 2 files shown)


docs/Table.md
```diff
@@ -149,6 +149,7 @@ This function accepts the following named parameters:
 |:---|:---|
 | className | Row-level class name |
 | columns | Array of React nodes |
+| key | Unique key within array of rendered rows |
```
wuweiweiwu:
```suggestion
| key | Unique value within array of rendered rows |
```

kgregory: This is fine, but I chose the original wording to be consistent with the [documentation for List](https://github.com/bvaughn/react-virtualized/blob/master/docs/List.md#rowrenderer)

source/Table/types.js
```diff
@@ -33,6 +33,7 @@ export type RowRendererParams = {
   columns: Array<any>,
   index: number,
   isScrolling: boolean,
+  key: string,
```
wuweiweiwu:
```suggestion
  key: string | number,
```

wuweiweiwu:
closed in 147fa55181fd186c8312c07507fe6cad38a7d190

I was unable to push changes onto your branch. 

thanks for doing this!

kgregory:
Not a problem.  Thanks for seeing it through.

Issue #1400, opened 2019-07-18
Cell Measurer's Dynamic Height List component example has broken images


nickstew:
## Issue

The example site for react-virtualized has broken images when attempting to display how the dynamic height list functionality works.

## Steps to reproduce

1. Navigate to https://bvaughn.github.io/react-virtualized/#/components/CellMeasurer
2. Click 'List: dynamic height image' button
3. Open dev tools console and see that https://fillmurray.com is not resolving

## Solution

Hopefully, https://www.fillmurray.com site adds a redirect or alias for https://fillmurray.com.  Until then, the line below can be simply changed to have 'www' in the hostname.

https://github.com/bvaughn/react-virtualized/blob/8738097a6ead422f74590d6a631bd5fd83ebc831/source/CellMeasurer/CellMeasurer.DynamicHeightList.example.js#L53

Fix:

```
const source = `https://www.fillmurray.com/${imageWidth}/${imageHeight}`;
```

I will immediately be submitting a PR to fix this example issue.

Have a great day!

PR #1401, opened 2019-07-18
#1400 Fixed hostname issue with images on example page

nickstew:
#1400 Here is my fix for the issue I just submitted on the example page not loading images.


nickstew:
Thanks, when I see a problem with any library I try to fix it myself.  :)  Thanks for reviewing and merging.  EZ PZ.

PR #1411, opened 2019-08-09
Split Table._setScrollbarWidth by adding public function Table.getScrollbarWidth

RoosDaniel:
For uses where a custom (external) header-row has to be used, there's little support for matching its width to the `Table`'s width. Most of it can be circumvented by summing `Column` widths, but there's no way to get the `scrollbarWidth` used since the method `_setScrollbarWidth` does both the calculation and `setState` parts.

With this change, implementations that rely on matching this repo's internal `_setScrollbarWidth` function with an external (almost identical) function could instead use `getScrollbarWidth`.

### DIFF
--- a/docs/Table.md
+++ b/docs/Table.md
@@ -61,6 +61,10 @@ This may be appropriate if the underlying row data has changed but the row sizes
 
 Gets offset for a given row and alignment.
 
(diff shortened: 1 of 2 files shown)


wuweiweiwu:
patch release

PR #1412, opened 2019-08-10
Adding CRA/Jest transformIgnorePatterns info

michal-ciechan:
I just spent couple of hours trying to figure out why one of my pages wasn't working with Jest, and blowing up on `jsdom` render via `react-testing-library`.

Turns out that all `node_modules` by default get ignored for transpiling, and therefore the `import` statement in `react-virtualized` blows up.

An option would be to pre-compile the library before publishing to npm (which is why I am assuming I haven't come across this yet)

But for now added a note to README in case someone else gets stuck on this for a while, as there wasn't anything obvious anywhere for someone new like me!

I am fairly new to JavaScript / React eco-system.

Apologies if this is the wrong section/area to add it to, let me know if there is a better place to put this, or better/more descriptive info as to what the underlying problem is.

- [ ] Format your code with [prettier](https://github.com/prettier/prettier) (`npm run prettier`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,15 +219,15 @@ With CRA ([Create-React-App](https://github.com/facebook/create-react-app)) and 
  Example `package.json`
  
  ```json
- {
- ...
-  "jest": {
+  {
     ...
-    "transformIgnorePatterns": [
-      "node_modules/(?!react-virtualized)/"
-    ]
+    "jest": {
+      ...
+      "transformIgnorePatterns": [
+        "node_modules/(?!react-virtualized)/"
+      ]
+    }
   }
-}
   ```
   **OR**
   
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -200,6 +200,51 @@ There are also a couple of how-to guides:
 * [Natural sort Table](docs/tableWithNaturalSort.md)
 * [Sorting a Table by multiple columns](docs/multiColumnSortTable.md)
 
+Testing
+---------------
+As this library does not come pre-compiled, it requires to be transformed by your loader, otherwise you may get errors such as:
```
wuweiweiwu: This depends on which version you're using. for example, react-virtualized also ships cjs builds which in this case would not break

wuweiweiwu: *common js

michal-ciechan: My issue was due to Create React App being setup by default to not transpile anything in node modules.

I've seen some people copy the code of default row renderer into their code to get around this. But for someone new to the ecosystem it wasn't obvious what and how to fix. I am happy to change the wording to anything, but definitely think it is worth mention to help someone else.

PR #1419, opened 2019-09-09
chore: allow dom-helpers@5.x

dominykas:
The breaking change was the dropped support of IE8 (not supported here) and a change in export layout: https://github.com/react-bootstrap/dom-helpers/blob/master/CHANGELOG.md#500-2019-08-30.


dominykas:
Just realized this would be a breaking change and would not work with older versions - will revisit shortly.

PR #1428, opened 2019-10-07
AutoSizer support for multi window added

mbman:
If `AutoSizer` is used in multi window enviroment, where the React app is running in a main window/process and renders content to a different child window, `AutoSizer` will not work properly because it's using the main window object and not the child one, who's dimensions we need.

This PR fixes the issue by detecting the current window in AutoSizer from `this._autoSizer.parentNode.ownerDocument.defaultView;`, which has the reference to proper window object and passing it to `detectElementResize`. In case that `defaultView` is `undefined` we fallback to default window object detection.


mbman:
Hi @bvaughn, could you please review this PR? Is this project still actively maintained? Thanks

mbman:
@wuweiweiwu could you please review this PR?

PR #1556, opened 2020-07-15
Add engines in package.json

MichaelDeBoey:


package.json
```diff
@@ -153,5 +153,8 @@
     "transform": [
       "loose-envify"
     ]
+  },
+  "engines": {
+    "node": ">=10.16.3"
```
wuweiweiwu, line 158: are there any dependencies that we have that need this node version? or else id rather not lock engine break other peoples development environments

PR #1579, opened 2020-08-14
Set sideEffects field in package.json to setup Webpack tree shaking

ramblingenzyme:
This marks any CSS files in the project as having side effects. This means Webpack won't strip out the CSS files if they've been imported as compared to marking the whole project as side effect free. 

This also lets Webpack tree shake, i.e. remove, exports that haven't been imported by consumers.
I've also removed the information in the README about direct imports and Webpack aliases for bundle size, since now bundlers will do the hard work for users.

Using this "app" with this fork linked using yarn, I compared the Webpack bundle when setting sideEffects to false, and to marking any CSS files as having side effects.
```js
import 'react-virtualized/styles.css';
import { AutoSizer } from "react-virtualized";
import React from "react";
import { render } from "react-dom";

const app = React.createElement(AutoSizer, null);

const rootElement = document.getElementById("yes");
if (rootElement) {
    render(app, rootElement);
}
```

### sideEffects: false
Screen Shot 2020-08-14 at 3 23 52 pm

### sideEffects: ["*.css"]
Screen Shot 2020-08-14 at 3 24 19 pm

As you can see, the second version still includes styles.css in the bottom right as you'd expect. If you want to test this locally this is the webpack config I used.
```js
const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;

module.exports = {
    module: {
        rules: [
            {
                test: /\.css$/i,
                use: ['style-loader', 'css-loader'],
            },
        ],
    },
    plugins: [
        new BundleAnalyzerPlugin()
    ]
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,34 +107,6 @@ import 'react-virtualized/styles.css';
 
 // You can import any component you want as a named export from 'react-virtualized', eg
 import {Column, Table} from 'react-virtualized';
-
-// But if you only use a few react-virtualized components,
-// And you're concerned about increasing your application's bundle size,
-// You can directly import only the components you need, like so:
-import AutoSizer from 'react-virtualized/dist/commonjs/AutoSizer';
-import List from 'react-virtualized/dist/commonjs/List';
-```
-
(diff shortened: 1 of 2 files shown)


ramblingenzyme:
@wuweiweiwu could you give this a look please?

PR #1610, opened 2020-11-06
Consider floating point widths and heights of parent node

delijah:
In some cases (like having a dynamic grid layout on the page) parent nodes have floating point widths and heights. `offsetWidth` and `offsetHeight` does not consider this, it just delivers integers. This results in wrong layouting. `getBoundingClientRect` considers floating point values, and therefore fixes the wrong layouting.

Sidenote: According to some research i've done, `getBoundingClientRect` returns the same values for `width` and `height` as `offsetWidth` and `offsetHeight`, except for the case when a parent node uses `transform: scale(...)`.

### DIFF
--- a/source/AutoSizer/AutoSizer.jest.js
+++ b/source/AutoSizer/AutoSizer.jest.js
@@ -19,14 +19,14 @@ describe('AutoSizer', () => {
     disableHeight = false,
     disableWidth = false,
     foo = 456,
-    height = 100,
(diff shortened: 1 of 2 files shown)


delijah:
Thinking about adding a fallback to `offsetWidth` and `offsetHeight` for older browsers ... 🤔

What do you think?

https://caniuse.com/getboundingclientrect

Issue #1575, opened 2020-08-05
Support Trusted Types


shhnjk:
## Bug Report
When [Trusted Types](https://web.dev/trusted-types/) is enabled on a website, following lines causes Trusted Types violation due to string assignment to a dangerous sink (i.e. _innerHTML_).
https://github.com/bvaughn/react-virtualized/blob/aef3918563f5e4a93cccdca2a02e92be665c1d90/source/vendor/detectElementResize.js#L185-L187

This can be prevented by changing those lines to something like:
```
var expandTriggerDiv = document.createElement('div');
expandTriggerDiv.class = 'expand-trigger';
expandTriggerDiv.appendChild(document.createElement('div'));
var contractTriggerDiv = document.createElement('div');
contractTriggerDiv.class = 'contract-trigger';
element.__resizeTriggers__.appendChild(expandTriggerDiv);
element.__resizeTriggers__.appendChild(contractTriggerDiv);
```
OR
```
var htmlString = 
  '<div class="expand-trigger"><div></div></div>' + 
  '<div class="contract-trigger"></div>';
if (window.trustedTypes) {
  var staticPolicy = trustedTypes.createPolicy(
    'react-virtualized-auto-sizer', {createHTML: () => htmlString});
  element.__resizeTriggers__.innerHTML = staticPolicy.createHTML('');
} else {
  element.__resizeTriggers__.innerHTML = htmlString;
}
```
Note that React already [supports Trusted Types](https://github.com/facebook/react/pull/16157).

shhnjk:
@wuweiweiwu, could you take a look at #1614 or #1608 to fix this issue? We are having deployment issue due to this bug. Thanks!

shhnjk:
@wuweiweiwu could you create a new v9.22.3 release with https://github.com/bvaughn/react-virtualized/commit/dfc11907cb720db58e974579993f8da5c6265c47?

PR #1614, opened 2020-11-21
Create Trusted Type policy for assigning an HTML in detectElementResize.js

shhnjk:
When react virtualized is used in a page where it enforces [Trusted Types](https://web.dev/trusted-types/), Trusted Types violation will trigger because there is an `innerHTML` usage in `detectElementResize.js`.

I have made a fix to create a Trusted Type policy that returns the same HTML to eliminate this Trusted Types violation.

This PR fixes https://github.com/bvaughn/react-virtualized/issues/1575.


shhnjk:
Either this or #1608 should fix the Trusted Types issue.

Issue #1199, opened 2018-08-21
Grid is not accessible using VoiceOver


backwardok:
## Reporting a Bug

Steps to reproduce:
1. Turn on VoiceOver (VO) and open Safari
2. Open a Grid example (I used the [example in the documentation](https://bvaughn.github.io/react-virtualized/#/components/Grid))
3. Use standard VO navigation (ctrl + option + arrows) to get to the grid
4. Try to navigate into the grid either by using standard VO navigation

### What is the current behavior?

Grid is simply named as "grid" and VO announces that there are 0 rows and 0 columns. Trying to navigate into the grid using the default "step into" key combination (ctrl + option + shift + down arrow) doesn't do anything. 

### What is the expected behavior?

The grid should have a meaningful name (not just "grid"). There should be an accurate number of rows and columns. There should also be a way to get into the grid and navigate around to hear the different values in the grid.

This is achievable by wrapping each row in a container and using `role="row"` on that container. Each column header cell should have `role="columnheader"`. Each row header cell should have `role-"rowheader"`. Each data cell should have `role="gridcell"`.

If the data in the table is dynamically loaded, then `aria-colcount` and `aria-rowcount` should be set on the the container that has `role="grid"`.

Additionally, `aria-readonly="true"` is currently getting set for all cases, even if the elements aren't ever editable. This attribute should only be applied if there is a state when they will be editable. Otherwise, that attribute should just be omitted entirely.

For more detailed information on this, check out http://w3c.github.io/aria-practices/#grid.

### Which versions of React and react-virtualized, and which browser / OS are affected by this issue? Did this work in previous versions of react-virtualized?
|                   |          |
|-------------------|----------|
| Browser           |  Safari (using 11.1.2, but repros on 10.x as well)        |
| OS                | MacOS (tested with 10.13.6)         |
| React             | ^16.2.0 (whatever your documentation server is running, but verified also in 16.2.0)         |
| React DOM         | ^16.2.0 (whatever your documentation server is running, but verified also in 16.2.1)         |
| react-virtualized | Tested on 9.20.1 (and verified it exists in 9.18.3)         |

This also occurs on iOS using Safari running VoiceOver there as well.

jsomsanith:
BTW I need that for my components too, and I'm interested in contributing, I started to fork and add the changes : https://github.com/jsomsanith/react-virtualized/tree/jsomsanith/fix/table_a11y
Tests missing.

Can any maintainer assign this issue to me ?

[...]

PR #1624, opened 2021-01-07
Fix grid roles for accessbility

asnewman:
Related to #1199 which was said to be fixed by #1208 but the PR only addressed the issue with Tables and not Grids.
This PR fixes a default role and adds another role to each cell which allows screen readers to correctly read Grids. Without this change, screen readers don't recognize any cells within the Grid and fails to announce them (tested with Voice Over on Firefox and Chrome).

### DIFF
--- a/source/Grid/Grid.jest.js
+++ b/source/Grid/Grid.jest.js
@@ -1102,7 +1102,7 @@ describe('Grid', () => {
     });
   });
 
-  describe('styles, classNames, and ids', () => {
+  describe('styles, classNames, ids, and roles', () => {
     it('should use the expected global CSS classNames', () => {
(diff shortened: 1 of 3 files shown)


wuweiweiwu (approved):
thanks for this fix and additional test cases. much appreciated

novwhisky:
Last publish was Dec 2022... can we get a release?

mallio:
This change actually broke my site a11y. Now I can't get VoiceOver to read cells because the innerScrollContainer default changed from the accurate 'rowgroup' to the inaccurate 'row' (since it contains the rows).

Working around it by setting containerRole='rowgroup' to restore the original functionality.

PR #1627, opened 2021-01-14
Use DOM API instead of creating Trusted Types policy to append a markup

shhnjk:
Trusted Types policy creation fails when there is a Trusted Type policy enforcement on the document and `addResizeListener` is called twice. 
Use DOM API to create the same tree so that we don't need to make Trusted Types policy creation little more complicated.


matiastucci:
@bvaughn any ETA for next release so I get this change? Thanks!

Blacknikomo:
Hi! We faced an issue when React-virtualize tries to create the Trusted Policy entries several times. It leads to the error "Policy already exists". Seemingly, this commit fixes the issue. But somehow, it's not appeared in the 9.22.3 version. Can you publish the next version 9.22.4 with this fix?

PR #1642, opened 2021-03-30
Fix bug in WindowScroller::updatePosition

yamadapc:
Hello, I'm opening this PR to fix a bug found in the `WindowScroller::updatePosition` method.

I've pushed a full sandbox reproduction of the bug, with documentation of what is the problem and the fix to https://codesandbox.io/s/7mdt8. I'd be happy to discuss the problem further on a video-call and or send you a video-recording I've of it through e-mail.

In summary, the bug happens when trying to react to "header height" changes with `WindowScroller`. The documentation points to calling `WindowScroller::updatePosition`, however, while that updates the cached position on the `WindowScroller` instance variables, it does not update the `scrollTop` state variable. This causes virtual-lists linked to it to not properly render immediately after a header height change; only after a subsequent scroll event.

I've added E2E tests which reproduce the error and a prop that fixes it by forcing a scroll event handler call.

There're tests for the prop being true/false so we can see that when it's false the problem exists but when true the problem is fixed.

The fix is essentially flagged by this prop and would be opt-in. I'm not sure if that's the best way.

### DIFF
--- a/jest-puppeteer.config.js
+++ b/jest-puppeteer.config.js
@@ -1,0 +1,6 @@
+module.exports = {
+  launch: {
+    headless: process.env.HEADLESS !== 'false',
+    devtools: process.env.DEVTOOLS === 'true',
+  },
+};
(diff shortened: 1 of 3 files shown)


thexpand:
When is this planned for release?

yamadapc:
I'll release it in the next few days.

Issue #1256, opened 2018-10-30
WindowScroller with scrollElement changing from null


OriR:
### Do you want to request a *feature* or report a *bug*?
**Bug**

### What is the current behavior?
When the `scrollElement` switches from `null` to a valid element and then unmounting `WindowScroller` - it crashes.
See here -> https://codesandbox.io/s/7wp9opl9n0

Click on the `unmount window scroller` and you'll see the error of `cannot read property 'splice' of undefined`.

I believe this is because to these lines
https://github.com/bvaughn/react-virtualized/blob/07e916701f7210ef819f19d63f4e96a2c999b34e/source/WindowScroller/WindowScroller.js#L142-L154

If the previous `scrollElement` was `null` we're not unsubscribing the even handlers, which is a memory leak but not the error, but we're also not subscribing to the current `scrolleElement` scroll events.
The error arises when you unmount the component and try to unsubscribe while we haven't even subscribed.

I think it should handle registering of the new `scrollElement` events separately from unregistering to the previous `scrollElement` events, that way you always register and deregister when needed.
Something like this:
```js
if (prevScrollElement !== scrollElement) {
  if (prevScrollElement) {
    unregisterScrollListener(this, prevScrollElement);

    this._unregisterResizeListener(prevScrollElement);
  }

  if (scrollElement) {
    this.updatePosition(scrollElement);

    registerScrollListener(this, scrollElement);

    this._registerResizeListener(scrollElement);
  }
}
```

### What is the expected behavior?
Change the `scrollElement` from `null` to a valid element without crashing.

I have a workaround that's working right now, doing this:
```js
<WindowScroller scrollElement={this.myElementRef} key={this.myElementRef}>
  ...
</WindowScroller>
```
I thought about it after reading what @bvaughn wrote [here](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key).
`key` controls whether to re-mount a certain component, if the key is different react will recycle everything and start again, that way I was able to skip the `componentDidUpdate` for cases when the `scrollElement` changes.

### Which versions of React and react-virtualized, and which browser / OS are affected by this issue? Did this work in previous versions of react-virtualized?
|                   |          |
|-------------------|----------|
| Browser           | all of them |
| OS                | all of them |
| React             | 16.x.x |
| React DOM         | 16.x.x |
| react-virtualized | 9.21.0 |

dannycochran:
[...]

PR #1660, opened 2021-06-08
Bug fix: Changing scrollElement prop on multiple WindowScrollers doesn't mount new scroll handlers

Aegz:
Hey guys,

I noticed there was a bug with the `WindowScroller` which stopped it from properly mounting the scroll and resize handlers to a new scrollElement if it is passed in via props. This is also considerably worse with multiple `WindowScroller`s and I want to fix that too.

Example: Single WindowScroller but we change the scrollElement from window to a div. 
https://codepen.io/willhua/pen/zYZwygO?editors=1111

The current workaround I have seen people using is to have a `key` prop on the `WindowScroller` which forcibly remounts the whole component therefore bypassing the logic in the `componentDidUpdate` lifecycle handler which normally deals with unmounting the old handlers and mounting the new ones.

E.g.
```
<WindowScroller scrollElement={someElement} key={someElement}>
```

I noticed this issue also happens with multiple `WindowScroller`s since when you change the `scrollElement` on multiple `WindowScroller`s this series of events occurs:
1. WindowScroller 1 mounts to window - attaches scroll/resize listeners
2. WindowScroller 2 mounts to window - does nothing
3. WindowScroller 1 prop change to a div - Fails to mount the listener to the new `scrollElement` because the props on the Components have already changed so this [condition fails](https://github.com/bvaughn/react-virtualized/blob/abe0530a512639c042e74009fbf647abdb52d661/source/WindowScroller/utils/onScroll.js#L73). This is most likely because we simply store references to the components themselves in `mountedInstances`
4. WindowScroller 2 prop change to a div - does not unmount old scroll/resize listeners due to existing `mountedInstances` (i.e. `!mountedInstances.length` is never true)

The problem lies here in unregisterScrollListener:
https://github.com/bvaughn/react-virtualized/blob/abe0530a512639c042e74009fbf647abdb52d661/source/WindowScroller/utils/onScroll.js#L84

The problem in this particular example is that we remove WindowScroller 1 then immediately add it back to that array straight away in `componentDidUpdate` on line 158:
https://github.com/bvaughn/react-virtualized/blob/abe0530a512639c042e74009fbf647abdb52d661/source/WindowScroller/WindowScroller.js#L146-L163

Once we get around to updating WindowScroller 2 we fail this condition in `unregisterScrollListener`  and never unmount the old listeners:
https://github.com/bvaughn/react-virtualized/blob/abe0530a512639c042e74009fbf647abdb52d661/source/WindowScroller/utils/onScroll.js#L87

<br />

I also noticed that with multiple `WindowScroller`s we weren't unmounting the handlers provided from `detectElementResize.js` since the `scrollListener` reference we are attempting to unmount is not the same reference as the one used to mount it (i.e. we have a closure which builds a `scrollListener` function then we attempt to unmount from a different closure with a new `scrollListener` function).

See: https://github.com/bvaughn/react-virtualized/blob/abe0530a512639c042e74009fbf647abdb52d661/source/vendor/detectElementResize.js#L194

Also: https://github.com/bvaughn/react-virtualized/blob/abe0530a512639c042e74009fbf647abdb52d661/source/vendor/detectElementResize.js#L224

The fixes:

`onScroll.js` - Instead of just storing a reference to a Component, we store the Component along with the Element we bound the listeners to `{ component, element }`. That way when we change the Component (the props update before we get into the `registerScrollListener` and `unregisterScrollListener` calls which is why they fail at the moment) we instead avoid the prop comparison which is unsafe and failure prone and instead compare against the Element reference we stored against the Component and if there are no known `WindowScroller`s registered against the component we remove the listener.

`detectElementResize.js` - This fails to remove the scroll listener attached to the element if there are multiple instances of the closure and in our case if there is more than 1 `WindowScroller` then we have multiple instances. The problem that occurs here is that if we unmount one of the other `WindowScroller`s the `removeEventListener` call fails to do anything because the `scrollListener` referenced there is generated in each individual closure and will NOT match the one currently bound to the element.

Video for some more evidence:
https://youtu.be/3GOi2yL88Qk

Also, sorry but I had to add `enzyme` to unit test the prop changing.


yamadapc:
Closes #1256

source/WindowScroller/utils/onScroll.js
```diff
@@ -7,7 +7,13 @@ import {
 } from '../../utils/requestAnimationTimeout';
 import type WindowScroller from '../WindowScroller.js';
 
-let mountedInstances = [];
+// Checking a component's props is unreliable. Instead we store the element used at the time
```
yamadapc, line 10: Why is this? Props change before unregistering?

source/vendor/detectElementResize.js
```diff
@@ -221,7 +223,14 @@ export default function createDetectElementResize(nonce, hostWindow) {
         1,
       );
       if (!element.__resizeListeners__.length) {
-        element.removeEventListener('scroll', scrollListener, true);
+        if (element.__scrollListener__) {
```
yamadapc, line 226: I think you don't need to do this on this case because `scrollListener` is declared in this function and captured here.

source/WindowScroller/utils/onScroll.js
Aegz: Yeah, we call this from `componentDidUpdate` and as a result any time we do the checks in `unregisterScrollListener` and `registerScrollListener` we always compare the new props against the given element which makes most of the calls fail

source/vendor/detectElementResize.js
Aegz: The problem in this case is when you have two or more `WindowScroller`s. What happens is the first `WindowScroller` instantiates a `detectElementResize` closure and registers its own `scrollListener` to the element.

Then when we unmount if the first `WindowScroller` isn't the first one to unmount, then the `removeEventListener` call fails because the references are wrong and as a result it doesn't remove the event listener which currently leaves it dangling

Aegz: Just tested the code without it. I had a dangling event listener at some point but it might have just been a defect caused by something else so I'll remove this

yamadapc (approved):

Issue #1657, opened 2021-05-25
Accessibility: Role "grid" is used inappropriately in the simple "list" component


MarcoZehe:
## Bug Report

If an app uses the list component from react-virtualized, the ARIA markup is incorrect. There is a "grid" container, but no "row" children and "gridcell" grandchildren. Screen readers of all platforms get utterly confused, report a grid with 0 rows, and on Mac, VoiceOver even doesn't render any of the child components, making apps largely inaccessible.

### What is the current behavior?

Inaccessible apps due to invalid markup. Grid containers end up with button children or such.

Examples: Signal-Desktop, DeltaChat, Facebook Messenger.

### What is the expected behavior?

List should use a generic container role like "region" instead, since you don't know what children there could be. Normally, "listbox" would be appropriate, but that requires that the children be of role "option". Since unlike with grids, you don't populate list children with default roles, this cannot be guaranteed, so for best generic accessibility, "region" is probably the best role to use here.

### Which versions of React and react-virtualized, and which browser / OS are affected by this issue? Did this work in previous versions of react-virtualized?
|                   |          |
|-------------------|----------|
| Browser           | Several  |
| OS                | Mac, Win |
| React             | Several  |
| React DOM         | Several  |
| react-virtualized | Several  |

nattri:
Facing the same issue. This makes all list items inaccessible to the voiceover when using accessibilities keys like ctrl+option+ right/left arrow.

nattri:
@bvaughn @TrySound Can we please merge this PR - https://github.com/bvaughn/react-virtualized/pull/1673 to fix current thread.

ericbf:
Two years down the line and this is still an issue? What’s the hold up here?

hailthekid:
➕ 1️⃣

ericbf:
This is still an issue

PR #1673, opened 2021-07-19
Adding role list to List component

nattri:
Added role=list for the List component as currently it has role="grid" which is creating accessibility issue in mac voiceover. 
Here is the issue ticket - https://github.com/bvaughn/react-virtualized/issues/1657


jcurtis:
Are there any unit tests that could be added/updated for this?

ericbf:
This is still an issue

PR #1815, opened 2023-04-09
Update peer deps to include React 17 and 18

bvaughn:
Much has changed in the main branch since the last release and I don't know how safe it is to release it. This release was made from the v9.22.3 release tag, and only changes peer dependencies to support React 17 and 18.


bvaughn:
Published.

https://github.com/bvaughn/react-virtualized/releases/tag/v9.22.4

PR #1894, opened 2026-06-06
fix: Fix Grid scrollToCell method not working if a scrollTop prop is defined, which is usual for WindowScroller usage

### LINKED ISSUE #1893: scrollToRow method and scrollToIndex prop in List when using WindowScroller broken since 9.20.0
## Bug Report

https://codesandbox.io/p/sandbox/react-virtualized-list-with-width-100-forked-gr3hcg?file=%2Fpackage.json%3A8%2C31

This example has a WindowScroller, with a AutoSizer and a List.

Clicking scrollToRow will attempt to use the scrollToRow method on the List.

Clicking prop:scrollToIndex will attempt to use the scrollToIndex prop on the List.

Since 9.19.0 when, I think (not checked this is exact cause), the `componentWillReceiveProps` was refactored the scrollToRow function no longer works.
Since 9.20.0 when, I think (not checked this is exact cause), the scrollTop was ignored in preference of scrollToRow etc., the scrollToIndex no longer works.

I investigated specifically the scrollToRow method and the issue is that calling setState since React 16.4 also triggers the getDerivedStateFromProps. Previously it matched componentWillReceiveProps and didn't call on setState. So I believe in the refactor of Grid to getDerivedStateFromProps the behaviour changed - it now gets called after the setState that sets the scrollTop. And unfortunately, the getDerivedStateFromProps ignores the value in prevState.

I think part of the issue is WindowScroller passes its scrollTop in. Which is then 0. And so this overrides the scrollToRow request. Likely this all needs handling in componentDidUpdate so that it can see the previous of both and work out if a change is requested as its at that point I assume it can tell which requested an action.

### What is the current behavior?

https://codesandbox.io/p/sandbox/react-virtualized-list-with-width-100-forked-gr3hcg?file=%2Fpackage.json%3A8%2C31

Change the version accordingly to reproduce.

### What is the expected behavior?

Both options to work

### Which versions of React and react-virtualized, and which browser / OS are affected by this issue? Did this work in previous versions of react-virtualized?
|                   |          |
|-------------------|----------|
| Browser           | Chrome, Safari         |
| OS                | macOS         |
| React             | 16, 17         |
| React DOM         | 16, 17         |
| react-virtualized | 9.20.0+         |

driskell:
- **N/A** For any new features, documentation has been added
- **N/A** For any documentation changes, the text has been proofread and is clear to both experienced users and beginners.

**NOTE:** AI was used to help build the test. The fix is completely my own.

-----

This fixes #1893 

The issue is that WindowScroller requires you to pass scrollTop prop. And in React 16.4+ getDerivedStateFromProps calls on every prop AND state change. So what happens is that once a scrollToCell method is called and everything updated, the final update before componentDidUpdate kicks in to update scroll position causes getDerivedStateFromProps to call but that then detects scrollTop and resets everything.

The fix is for getDerivedStateFromProps to only obey scrollTop/scrollLeft when they change.

Test added that fails without the fix.


driskell:

PR #1895, opened 2026-06-06
fix: Fix Grid scrollToCell not able to scroll to last element when using autoHeight with WindowScroller

driskell:
- **N/A** For any new features, documentation has been added
- **N/A** For any documentation changes, the text has been proofread and is clear to both experienced users and beginners.

**NOTE:** AI was used to help build the tests. The fix is completely my own.

-----

When calling Grid's `scrollToCell` (maybe via List) and autoHeight is true because you might be using WindowScroller, when scrolling to the last item the scroll top is limited as if the Grid was the viewport. Even if you have items after the Grid allowing the window to scroll further it will never scroll past this point. It means with alignment of start, you can't actually ever scroll the last item to be at the start, it'll scroll to a point where that item is further down the page.

This removes the limitation on the scroll if autoHeight is true. Same logic is applied also to the scroll left logic.

Test is added too for both the position manager and the Grid.


driskell:

PR #1896, opened 2026-06-06
feat: Add additional scrollToOffset (prop) / offset (method calls) to allow scrolling to cells but offset by some pixels

driskell:

**NOTE:** AI was used to help build the tests and also to check some of the initial basic work I did for this and fill in some gaps.

-----

This might be something to discuss and it was quick to mock up. It's a feature I found us needing.

Essentially we have WindowScroller, AutoSizer and List. We have a sticky navigation at the top of the page. Using scrollToRow / scrollToCell was problematic as it was hiding the cell underneath the sticky navigation. We needed a way to essentially tell the virtualiser that we need cell X to start 30 pixels down the page. In other words offset by `-30`.

I chose not to mess with alignment (could it be a number for example that becomes offset) as it seems more flexible with a new property. You can use the offset in coordination with any value. For example if you had a footer that floated above content you could still use `end` and an offset to prevent it appearing underneath it.

Open to discussion here and if there's other approaches but this seemed relatively straight forward and a good enhancement - but of course I would when I benefit from it! :)


driskell:
How do I reopen?

driskell:

